PR #20, opened 2017-06-01
fix syntax

xqbumu:


hsluoyz:
Hi @xqbumu ,

the missing ``]`` is intended, this example shows how Casbin handles the parsing error in model file.

xqbumu:
@hsluoyz I finally got it, so I closed it

PR #22, opened 2017-06-12
Fixed columns to properly display csv

drgomesp:


hsluoyz:
Hi @drgomesp ,

I know the CSV display provided by GitHub. But according to [RFC 4180](https://en.wikipedia.org/wiki/Comma-separated_values), a CSV should:

1. is plain text using a character set such as ASCII, various Unicode character sets (e.g. UTF-8), EBCDIC, or Shift JIS,
2. consists of records (typically one record per line),
3. with the records divided into fields separated by delimiters (typically a single reserved character such as comma, semicolon, or tab; sometimes the delimiter may include optional spaces),
4. where every record has the same sequence of fields.

So actually Casbin's CSV doesn't obey 4, which causes this issue.

But Casbin's policy doesn't need to be same number of columns, so I don't know if there's a better file format for Casbin policy.

PR #37, opened 2017-08-29
Feature: Define RoleManager as interface

### LINKED ISSUE #36: RoleManager interface to provide custom implementations
I'm looking to determine if it would be agreeable to switch from RoleManager the struct to RoleManager the interface.

```
type RoleManager interface {
    AddLink(name1 string, name2 string, domain ...string)
    DeleteLink(name1 string, name2 string, domain ...string)
    HasLink(name1 string, name2 string, domain ...string) bool
    GetRoles(name string, domain ...string) []string
    GetUsers(name string) []string
    PrintRoles()
}
```

And the current implementation of RoleManager could become the default implementation.

The only directly impacted part of the code would be
https://github.com/casbin/casbin/blob/master/model/assertion.go

The attribute RM would be interface rbac.RoleManager instead of pointer to an instance of RoleManager. And the implementation of `buildRoleLinks()` https://github.com/casbin/casbin/blob/748b81503e7166d9d43e3d66f3cffae3c8b1ebdd/model/assertion.go#L33 would need a different method for getting a RoleManager to allow a different implementation.

I can start by submitting a PR with the change, and get additional feedback if the above is agreeable.

kenjones-cisco:
Define RoleManager as interface, and rename existing RoleManager struct
to defaultRoleManager to provide a default implementation and backwards
compatability.

Defines RoleManagerConstructor to allow specifying a function for
creating a new RoleManager.

Adds SetRoleManagerConstructor method to Model to allow setting the
function to call for a new instance of RoleManager.
Adds implementation of RoleManagerConstructor to Model that will create
an instance of the defaultRoleManager just like before.

Updates Assertion to reference RoleManager interface instead of an
instance. Updates Assertions.buildRoleLinks() to use newRoleManagerFunc
that holds the RoleManagerConstructor.

All test are passing with the only change being *RoleManager to
RoleManager.

Resolves #36

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,6 +175,24 @@ RESTful | [keymatch_model.conf](https://github.com/casbin/casbin/blob/master/exa
 Deny-override | [rbac_model_with_deny.conf](https://github.com/casbin/casbin/blob/master/examples/rbac_model_with_deny.conf)  | [rbac_policy_with_deny.csv](https://github.com/casbin/casbin/blob/master/examples/rbac_policy_with_deny.csv)
 Priority | [priority_model.conf](https://github.com/casbin/casbin/blob/master/examples/priority_model.conf)  | [priority_policy.csv](https://github.com/casbin/casbin/blob/master/examples/priority_policy.csv)
 
+## RoleManager
+
+To use a custom RoleManager implementation.
+
+```go
+
+    type myCustomRoleManager struct {} // assumes the type satisfies the RoleManager interface
+
+    func newRoleManager() rbac.RoleManagerConstructor {
+        return func() rbac.RoleManager {
+            return &myCustomRoleManager{}
+        }
+    }
+
+    e := casbin.NewEnforcer("path/to/model.conf", "path/to/policy.csv")
+    e.SetRoleManager(newRoleManager())
+```
+
 ## How to use Casbin as a service?
 
 - [Go-Simple-API-Gateway](https://github.com/Soontao/go-simple-api-gateway): A simple API gateway written by golang, supports for authentication and authorization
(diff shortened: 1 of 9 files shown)


hsluoyz:
I have several questions:

1. I saw ``newRoleManagerFunc`` in model.go is a global variable. So two Casbin enforcers will affect each other. I think it should be a member of Enforcer to avoid the conflicts.

2. ``defaultRoleManager`` can be refactored out to ``default_role_manager.go``. ``role_manager.go`` will only contain the interface.

3. The function comments in ``RoleManager`` interface didn't describe the ``domain`` parameter.

kenjones-cisco:
If the `RoleManagerConstructor` is added to the Enforcer, that means that `func (model Model) BuildRoleLinks()` and `func (ast *Assertion) buildRoleLinks()` would need to accept a `RoleManagerConstructor` as input given that `func (ast *Assertion) buildRoleLinks()` creates a new RoleManager instance on each execution.

Any concerns with change the signatures of those methods?

hsluoyz:
The code look good to me. Can you add a working example (e.g. ``TestRBACModelWithCustomRoleManager()``) to ``model_test.go``? In it, you can start with the RBAC example, then set a custom role manager. The role manager can be very simple, such as using hard-coded values. Then call testEnforce to show how the new role manager affects the enforcing results.

```go
func TestRBACModel(t *testing.T) {
	e := NewEnforcer("examples/rbac_model.conf", "examples/rbac_policy.csv")

	testEnforce(t, e, "alice", "data1", "read", true)
	testEnforce(t, e, "alice", "data1", "write", false)
	testEnforce(t, e, "alice", "data2", "read", true)
	testEnforce(t, e, "alice", "data2", "write", true)
	testEnforce(t, e, "bob", "data1", "read", false)
	testEnforce(t, e, "bob", "data1", "write", false)
	testEnforce(t, e, "bob", "data2", "read", false)
	testEnforce(t, e, "bob", "data2", "write", true)
}
```

PR #42, opened 2017-10-13
[feature] Enable multiple policies/roles in API

faceless-saint:
Expand the Management API to include functions allowing for multiple policy definitions to be managed concurrently, for example grouping policies "g" and "g2" in the case of RBAC with resource roles enabled. Preexisting functionality is unchanged.

The primary use case I'm interested in is support for resource roles, as in the example above, but this change will allow for more flexible usage of the API in general, to more closely mirror what is possible for static policies. My goal is to minimize disruption and fit into the existing API as cleanly as possible.

Example:
```golang
enforcer.AddGroupingPolicy("user", "admin")  // unchanged
enforcer.AddNamedGroupingPolicy("g2", "data1", "data_role") // new capability
```

### DIFF
--- a/management_api.go
+++ b/management_api.go
@@ -24,8 +24,8 @@ func (e *Enforcer) GetAllSubjects() []string {
 }
 
 // GetAllNamedSubjects gets the list of subjects that show up in the current named policy.
-func (e *Enforcer) GetAllNamedSubjects(name string) []string {
-	return e.model.GetValuesForFieldInPolicy("p", name, 0)
+func (e *Enforcer) GetAllNamedSubjects(ptype string) []string {
+	return e.model.GetValuesForFieldInPolicy("p", ptype, 0)
 }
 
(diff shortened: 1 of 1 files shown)


faceless-saint:
FYI, there seems to be a CI issue unrelated to my changes. I submitted a dummy PR with no changes and the build failed.

hsluoyz:
The CI was accidentally broken by one of my commits, it has been fixed by: https://github.com/casbin/casbin/commit/944fc29e6ecc6f356f7808b1de3360aca13e7cd0

hsluoyz:
I'm OK with the new function like ``GetAllNamedSubjects()``. But for arguments such as ``g``, ``g2``, we already name them as ``ptype`` in code of adapters and internal functions. So for consistency, can you rename the arguments of these new functions from ``name`` to ``ptype``? The function names don't need this change.

hsluoyz:
lgtm

Issue #43, opened 2017-10-19
RBAC sessions


julisch94:
Hey there,

I want to realize an **RBAC session concept** that supports the following features:
- (A user can have assigned any number of roles.)
- (A role can have assigned any number of permissions.)
- A user needs to activate one or more of his roles to actually being able to use them.
- This should be done with **sessions**.
- A session consists of _one_ user and _one or more_ roles. It should have a creation date and an expiration date.
- The policy matcher should then only grant access if and only if the requesting user has a running session activating the needed role.
- So: It is *not* sufficient to only have the user-role-assignment. You will always need to activate your needed roles with a session. This supports the principle of [Least Privilege](https://en.wikipedia.org/wiki/Principle_of_least_privilege).

This concept is derived from Ravi Sandhu's paper on "role-based access control".

My ideas are the following:
Change the request definition to be

 ```ini
 r = sub, obj, act, time
```

Create a session definition:

 ```ini
s = sub, role, start_time, end_time
```
How am I then able to match an access request. Like this?

 ```ini
m = s(r.sub, p.sub, _, _) && r.obj == p.obj && r.act == p.act && r.time > s.start_time && r.time < s.end_time
```

BTW: All time data will be unix time in nanoseconds, so I hope they'll be comparable.

hsluoyz:
I think you can implement a custom role manager based on [the interface](https://github.com/casbin/casbin/blob/master/rbac/role_manager.go). And you can call ``Enforcer.SetRoleManager()`` to set the role manager to your own. I agree with your session definition. But I think we don't need to use a new name ``s``. We can just use ``g``.

As below, I have allowed to set 4 parameters to role manager's ``AddLink()`` function. So ``Addlink()`` will receive the 4 tuple of ``sub, role, start_time, end_time``.

https://github.com/casbin/casbin/blob/9a33e2dc10f1bcf57f20448254022d91f51cf58c/model/assertion.go#L35-L57

And for role manager's HasLink(), you can still use ``g(r.sub, p.sub, r.time)``. As below, I already allow the 3rd parameter to be passed into ``HasLink(). Although it is the ``domain``, but it can be reused as ``time``. As they are both strings, there's no difference.

https://github.com/casbin/casbin/blob/c1f8fa592a00ea23d117cc88531765fd24f9ca2e/enforcer.go#L246-L272

[...]

PR #44, opened 2017-10-20
Session support for RBAC

julisch94:
- Implemented a custom role manager for RBAC with sessions.
- Also added documentation to the code and included two simple examples: one model example and one policy example.

Not able to write tests because my compiler doesn't find `t.Helper()`.

Targetting https://github.com/casbin/casbin/issues/43


hsluoyz:
Thanks for the quick PR! An issue is the test is missing. ``t.Helper()`` is supported in Go 1.9. Could you upgrade to Go 1.9.1?

julisch94:
I'll do that and add some tests by the end of next week.

julisch94:
Could you please make the function GetUsers(name string) in RoleManager also accept multiple arguments? Because I'd like to use it as "Who is using role r at time t?", which will only be possible if I can pass one more argument to GetUsers.

That would be great!

hsluoyz:
Added here: https://github.com/casbin/casbin/commit/a57b2b7e0f2c28d42ddcef2d98f9b37c83394d47

julisch94:
I have lately added some tests with coverage above 80%.

rbac/session_role_manager.go
```diff
@@ -16,6 +16,7 @@ package rbac
 
 import (
 	"github.com/casbin/casbin/util"
+	"sort"
```
hsluoyz: Please put the built-in imports before github imports.

hsluoyz:
@julisch94  the test failed:

```
	session_role_manager_test.go:48: alpha < bravo at time 1509106078517347100: true, supposed to be false
	session_role_manager_test.go:49: alpha < charlie at time 1509106078517347100: true, supposed to be false
	session_role_manager_test.go:50: bravo < delta at time 1509106078517347100: true, supposed to be false
	session_role_manager_test.go:51: bravo < echo at time 1509106078517347100: true, supposed to be false
	session_role_manager_test.go:52: charlie < echo at time 1509106078517347100: true, supposed to be false
	session_role_manager_test.go:53: charlie < foxtrott at time 1509106078517347100: true, supposed to be false
```

julisch94:
I see you made a few improvements and an import fix. Do these happen to fix the test fails you mentioned? Because having pulled your latest changes, I can't reproduce the failing tests.

hsluoyz:
Yeah, I fixed them, no issues now. 👍

PR #54, opened 2017-12-01
Improve test coverage from 76% to 99%

cdennison:
Hi Yang,

Here's a PR to improve your code coverage from 76% to 99% (I could probably get the last little bit with some refactoring). Not sure how Coveralls is calculating coverage - I'm using go test -coverprofile=coverage.out github.com/casbin/casbin)

I mostly tried to add new test cases but some functionality wasn't testable without minor code changes.
--Added safe methods for AddPolicySafe/RemovePolicySafe/RemoveFilteredPolicySafe
--Simplified the recover blocks for enforcer_safe (I saw that you got your code from Stack Overflow but that answer in my mind is overly complicated and untestable). I used the same approach I've seen in other projects like gin-gonic https://github.com/gin-gonic/gin/blob/master/recovery.go#L37.
--Added a mock file adapter

Also the sync enforcer is hard to test as written. So I replaced all the test cases to test sync instead of the regular one. If you want to test both you can either:
--Duplicate your test cases substantially
--Create a common interface and test that (I'm happy to do that) https://medium.com/agrea-technogies/mocking-dependencies-in-go-bb9739fef008

### DIFF
--- a/enforcer_safe.go
+++ b/enforcer_safe.go
@@ -14,21 +14,16 @@
 
 package casbin
 
-import "errors"
+import (
+	"errors"
+	"fmt"
+)
 
 // NewEnforcerSafe calls NewEnforcer in a safe way, returns error instead of causing panic.
 func NewEnforcerSafe(params ...interface{}) (e *Enforcer, err error) {
 	defer func() {
 		if r := recover(); r != nil {
-			switch x := r.(type) {
-			case string:
-				err = errors.New(x)
-			case error:
-				err = x
-			default:
-				err = errors.New("Unknown panic")
-			}
-
+			err = errors.New(fmt.Sprintf("%v",r))
 			e = nil
 		}
 	}()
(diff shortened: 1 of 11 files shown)


hsluoyz:
The import format and code format (especially the spaces after comma, and between two words are wrongly removed) can be improved. I think you can rebase your commits to fix these formats (instead of fixing them in a new commit). So I can see clearly which part of code is changed in this PR.

cdennison:
I'm happy to format that code in whatever way. I just did 
`go fmt ./...`

Is there another approach I should use? Looks like a I did miss a couple formatting issue so this should be clean now. I also simplified it to only have the first commit. I can submit the second as a separate PR.

hsluoyz:
I don't know whether my format is the same as gofmt, but I just follow a classic format. I have highlighted the places that need to change in the reviews.

cdennison:
Sorry I don't see any comments on the code yet, but will gladly fix the formatting.

hsluoyz:
Can you see the reviews in this PR page?

PR #60, opened 2017-12-19
Tests to demo model/policy getters/setters.

cdennison:
Showing that getters/setters work correctly.


cdennison:
Code coverage went down because this test is not "deterministic" and will sometime fail to finish because it needs a sleep at the end. Fixing in a follow up PR.
https://github.com/casbin/casbin/blob/master/enforcer_test.go#L322

PR #78, opened 2018-03-14
[feature] Add policy filtering

faceless-saint:
A new interface, `FilteredAdapter`, extends the `Adapter` interface
with the ability to load a filtered subset of the back-end policy.
This allows Casbin to more effectively scale when enforcing a very
large number of policies, such as a busy multi-tenant system. There
is also a second built-in file adapter supporting this feature.

To prevent accidental data loss, the `SavePolicy` method is disabled
when a filtered policy is loaded. For maximum compatibility, whether
a given `Adapter` implements the new feature is checked at runtime.


hsluoyz:
1. Please add another adapter for this filtered file adapter here: ``persist/filtered-file-adapter/adapter.go``? I want the default file adapter to be simple and friendly for new users. I know this filtered file adapter is compatible with the original one. But our users use file adapter as an example to implement their own adapter. The filtering code will complicate their understanding. You can implement a filtered adapter, then make it a subclass of the original one to remove the duplicated code, just add your own functions and structs.

2. If possible, please submit the code style change in another commit. The current commit makes me difficult to see where the real change is.

hsluoyz:
1. For the 2nd commit, please add another adapter for this filtered file adapter here: ``persist/filtered-file-adapter/adapter.go``. I want the default file adapter to be simple and friendly for new users. I know this filtered file adapter is compatible with the original one. But our users use file adapter as an example to implement their own adapter. The filtering code will complicate their understanding. You can implement a filtered adapter, then make it a subclass of the original one to remove the duplicated code, just add your own functions and structs.

2. The 2nd commit has some code style re-formatting. If possible, please submit the code style change in another commit. The current commit makes me difficult to see where the real change is.

3. Please fix the ``codeclimate`` and ``coverage/coveralls`` downgrade.

persist/file-adapter/adapter.go
```diff
@@ -85,18 +84,12 @@ func (a *Adapter) loadPolicyFile(model model.Model, handler func(string, model.M
 	}
 	defer f.Close()
 
-	buf := bufio.NewReader(f)
-	for {
-		line, err := buf.ReadString('\n')
-		line = strings.TrimSpace(line)
+	scanner := bufio.NewScanner(f)
```
faceless-saint: I took the opportunity to integrate my streamlined handler into the base file adapter, as it will be simpler for others to use as a reference IMO.

faceless-saint:
1. DONE - the `fileadapter` package has a new struct, `FilteredAdapter`, which implements the new interface. This is implemented cleanly in a separate file, and should provide a useful reference for new filtered adapter implementations in the same way that the existing file adapter is.

2. DONE - all style edits from `go fmt` are now contained in a separate commit without functional changes.

3. DONE - refactored new code to satisfy `codeclimate` constraints and added additional tests to satisfy `coverage/coveralls`. The refactored line scanning, using a `bufio.Scanner` instead of a `bufio.Reader`, was also backported to the appropriate `fileadapter.Adapter` method to further reduce cognitive complexity.

PR #118, opened 2018-08-07
fix CachedEnforcer panic

maplebeats:
1.add locker CachedEnforcer
2.check the interface convert
3.add benchmark cached file


hsluoyz:
Looks good to me. Thanks!

PR #123, opened 2018-08-16
Add test case TestMatcherUsingInOperator & examples/rbac_model_matcher & update README.md

Outman9946:
Add test case TestMatcherUsingInOperator & examples/rbac_model_matcher & update README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,6 +84,12 @@ m = r.sub == p.sub && r.obj == p.obj && r.act == p.act
 # We also support multi-line mode by appending '\' in the end:
 # m = r.sub == p.sub && r.obj == p.obj \ 
 #  && r.act == p.act
+
+# Further more, if you are using ABAC, you may try expression `in` like following:
+# r.obj == p.obj && r.act == p.act || r.obj in ('data2', 'data3')
+# But you SHOULD make sure that the length of the array is MORE than 1, 
+# otherwise there will cause a panic
+
 ```
 
 An example policy for ACL model is like:
--- a/README.md
+++ b/README.md
@@ -112,7 +112,7 @@ Further more, if you are using ABAC,  you can try operator `in` like following i
 m = r.obj == p.obj && r.act == p.act || r.obj in ('data2', 'data3')
 ```
 
-But you **SHOULD ** make sure that the length of the array is **MORE **than **1**,  otherwise there will cause a panic.
+But you **SHOULD** make sure that the length of the array is **MORE** than **1**, otherwise there will cause it to panic.
 
 For more operators, you may take a look at [govaluate](https://github.com/Knetic/govaluate)
 
--- a/README.md
+++ b/README.md
@@ -81,15 +81,6 @@ e = some(where (p.eft == allow))
 [matchers]
 m = r.sub == p.sub && r.obj == p.obj && r.act == p.act
 
-# We also support multi-line mode by appending '\' in the end:
-# m = r.sub == p.sub && r.obj == p.obj \ 
-#  && r.act == p.act
-
-# Further more, if you are using ABAC, you may try expression `in` like following:
-# r.obj == p.obj && r.act == p.act || r.obj in ('data2', 'data3')
-# But you SHOULD make sure that the length of the array is MORE than 1, 
-# otherwise there will cause a panic
-
 ```
 
 An example policy for ACL model is like:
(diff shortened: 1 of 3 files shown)


enforcer_test.go
```diff
@@ -198,6 +198,41 @@ func TestNotUsedRBACModelInMemory(t *testing.T) {
 	testEnforce(t, e, "bob", "data2", "write", true)
 }
 
+func TestMatcherUsingInOperator(t *testing.T) {
+	// In memory
```
hsluoyz: Is it necessary to load from model file and text string at the same time? Both ways are already well tested in other test cases. And this is not this PR's focus. I think just using model file only will be OK.

enforcer_test.go
```diff
@@ -198,6 +198,41 @@ func TestNotUsedRBACModelInMemory(t *testing.T) {
...
+	e.AddPermissionForUser("alice", "data1", "read")
+
+	testEnforce(t, e, "alice", "data1", "read", true)
+	testEnforce(t, e, "anyone", "data2", "read", true)
+
+	// From file config
+	e = NewEnforcer("examples/rbac_model_matcher_using_in_op.conf")
+	e.AddPermissionForUser("alice", "data1", "read")
+
+	testEnforce(t, e, "alice", "data1", "read", true)
```
hsluoyz: The ``testEnforce`` assertions should be thorough. I think only two ``true`` assertions are not persuasive enough. For example, there can be six:

alice, data1
alice, data2
alice, data3
anyone, data1
anyone, data2
anyone, data3

README.md
```diff
@@ -104,6 +95,27 @@ It means:
...
+
+```ini
+# Matchers
+[matchers]
+m = r.obj == p.obj && r.act == p.act || r.obj in ('data2', 'data3')
+```
+
+But you **SHOULD ** make sure that the length of the array is **MORE **than **1**,  otherwise there will cause a panic.
+
+For more operators, you may take a look at [govaluate](https://github.com/Knetic/govaluate/blob/master/README.md)
```
hsluoyz: A little advice, use ``https://github.com/Knetic/govaluate`` instead because it's shorter and does the same thing :)

README.md
```diff
@@ -104,6 +95,27 @@ It means:
...
+We also support multi-line mode by appending '\\'  in the end:  
+
+```ini
+# Matchers
+[matchers]
+m = r.sub == p.sub && r.obj == p.obj \ 
+  && r.act == p.act
+```
+
+Further more, if you are using ABAC,  you can try operator `in` like following:
```
hsluoyz: I like the ``in`` feature. But as Casbin is also ported to Java (jCasbin) and Node.js (Node-Casbin), I think we should remind the users that features like ``in`` are only available in the Golang Casbin.

This is because it's provided by the Golang ``govaluate`` evaluator and we use other evaluators for jCasbin and Node-Casbin. They probably don't have features like ``in``.

enforcer_test.go
Outman9946: That's right, I just want to prove that it will also work well in in-memory mode. 😁

enforcer_test.go
```diff
@@ -199,37 +199,15 @@ func TestNotUsedRBACModelInMemory(t *testing.T) {
...
-	e = NewEnforcer("examples/rbac_model_matcher_using_in_op.conf")
+	e := NewEnforcer("examples/rbac_model_matcher_using_in_op.conf")
 	e.AddPermissionForUser("alice", "data1", "read")
 
 	testEnforce(t, e, "alice", "data1", "read", true)
+	testEnforce(t, e, "alice", "data2", "read", true)
+	testEnforce(t, e, "alice", "data3", "read", true)
+	testEnforce(t, e, "anyone", "data1", "read", false)
+	testEnforce(t, e, "anyone", "data2", "read", true)
 	testEnforce(t, e, "anyone", "data2", "read", true)
```
hsluoyz: Should be ``data3``

README.md
```diff
@@ -104,6 +95,27 @@ It means:
...
+
+```ini
+# Matchers
+[matchers]
+m = r.obj == p.obj && r.act == p.act || r.obj in ('data2', 'data3')
+```
+
+But you **SHOULD ** make sure that the length of the array is **MORE **than **1**,  otherwise there will cause a panic.
```
hsluoyz: Should be **SHOULD** or **SHOULD **? The same about MORE.

Also there's an English grammar issue in ``there will cause a panic.``

Outman9946: Alright, there are something wrong with my MarkDown editor

PR #131, opened 2018-09-13
remove redundant parenthesis from type conversion exprs

Quasilyte:
(T)(v) is identical to T(v) when T is not a pointer type.

Found using https://go-critic.github.io/overview#typeUnparen-ref


hsluoyz:
lgtm

PR #132, opened 2018-09-13
use type switch var binding to reduce the amount of type assertions

Quasilyte:
Found using https://go-critic.github.io/overview#typeSwitchVar-ref

### DIFF
--- a/enforcer.go
+++ b/enforcer.go
@@ -65,28 +65,28 @@ func NewEnforcer(params ...interface{}) *Enforcer {
 	}
 
 	if len(params)-parsedParamLen == 2 {
-		switch params[0].(type) {
+		switch p0 := params[0].(type) {
 		case string:
-			switch params[1].(type) {
+			switch p1 := params[1].(type) {
 			case string:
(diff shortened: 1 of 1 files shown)


enforcer.go
```diff
@@ -339,18 +339,18 @@ func (e *Enforcer) Enforce(rvals ...interface{}) bool {
...
-				if result.(float64) == 0 {
+				if result == 0 {
 					policyEffects[i] = effect.Indeterminate
 					continue
 				} else {
-					matcherResults[i] = result.(float64)
+					matcherResults[i] = result
 				}
 			default:
 				panic(errors.New("matcher result should be bool, int or float"))
```
Quasilyte: This error looks misleading as I don't see `case int` in this switch.
It's a separate issue that should be fixed by someone familiar with this code.

hsluoyz: This is intended, because our library will convert user inputted ``int`` into ``float`` automatically. So our code don't handle ``int`` separately.

hsluoyz:
lgtm

Issue #157, opened 2018-10-29
Structured logging proposal


llonchj:
Thanks for casbin, this is an amazing product I use in several projects.

I am willing to contribute casbin with a logging abstraction interface so we can deliver structured logging to a logger, which by default will be golang `log` package, as implemented today.

The main benefit will be the ability to push logs to other loggers (zap, logrus, etc...) following casbin's development pattern of keeping the core simple and extensible into specific party solutions.

I am willing to start a discussion in order to gather ideas and requirements to proceed with a PR.

vtolstov:
This is good thing to have, can you provide example changes and does it breaks api compatibility.

llonchj:
Hi @vtolstov!

The main idea will be to enrich `util/log_util.go` with a interface to be able to log using golang's log but also logrus or zap or zerolog...

```go

type Logger interface{
  LogPrint(...interface{})
  LogPrintf(...interface{})
}

// Using golang log
type DefaultLogger struct {}
func (l *DefaultLogger) LogPrint(v ...interface{}) {log.Print(v)}
func (l *DefaultLogger) LogPrintf(v ...interface{}) {log.Printf(v)}

var Logger=DefaultLogger{}

// LogPrint prints the log.
func LogPrint(v ...interface{}) {
	if EnableLog {
            Logger.Print(v)
	}
}
```

This is an *incomplete, partial implementation* just for you to evaluate.

Will not break any changes and will provide an application the ability to log using any of the existing libraries for the purpose. 

It is obvious to say that the interface can be extended implementing `Debug`... `Debugf`... this can be done in a second iteration.

I can start working ASAP on this and push a PR as is a requirement on my current project. What do you think?

vtolstov:
i'm +1, but @hsluoyz needs approve this changes..

hsluoyz:
Thanks for the proposal. It's a good feature I think. My concern is that the main Casbin library should not directly import any 3rd-party logging tools. As your code, we can define a ``Logger`` interface, and provide a built-in ``DefaultLogger``. All other ``Logger`` implementations should be done in a separate project, and will be imported into a Casbin instance via the ``SetLogger()`` call in a dependency injection way. Just like how we do it for Adapter, Watcher, RoleManager.

If a Casbin user doesn't call ``SetLogger()``, he should be able to use ``DefaultLogger`` by default.

hsluoyz:
Closed as it has been fixed.

PR #158, opened 2018-10-30
Logger implementation

llonchj:
PR for #157 

@hsluoyz, do you think `utils.EnableLog` might now be redundant but actioning something might break the api? Any thoughts?

Also please consider editing the logging section in the documentation.

### DIFF
--- a/enforcer.go
+++ b/enforcer.go
@@ -19,7 +19,6 @@ import (
 	"fmt"
 
 	"github.com/Knetic/govaluate"
-
 	"github.com/casbin/casbin/effect"
 	"github.com/casbin/casbin/model"
 	"github.com/casbin/casbin/persist"
@@ -272,9 +271,9 @@ func (e *Enforcer) EnableEnforce(enable bool) {
 	e.enabled = enable
 }
 
-// EnableLog changes whether to print Casbin log to the standard output.
+// EnableLog changes whether Casbin will log messages to the Logger.
 func (e *Enforcer) EnableLog(enable bool) {
-	util.EnableLog = enable
+	util.GetLogger().EnableLog(enable)
 }
 
 // EnableAutoSave controls whether to save a policy rule automatically to the adapter when it is added or removed.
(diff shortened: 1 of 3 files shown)


hsluoyz:
Why not add ``EnableLog()`` to the Logger interface? The API will not be broken in this way.

And about documentation, can you propose a PR too? Click the right-top ``EDIT`` in this page: https://casbin.org/docs/en/log-error

llonchj:
@hsluoyz in `utils.go`, EnableLog is a boolean. How will you do this exactly?

hsluoyz:
There's a call:

https://github.com/casbin/casbin/blob/c96868ecc4e0abd15af3b349d58fa663c63aa25f/enforcer.go#L275-L278

llonchj:
As util.EnableLog is already a bool, the following will be a breaking change:

```
// EnableLog changes whether Casbin will log messages to the Logger.
func (e *Enforcer) EnableLog(enable bool) {
	util.EnableLog(enable)
}
```

This will be a more safe code, but if util.Logger is nil, will have no effect

```
// EnableLog changes whether Casbin will log messages to the Logger.
func (e *Enforcer) EnableLog(enable bool) {
   if util.Logger!=nil {
	util.Logger.EnableLog(enable)
   }
}
```

I am quite unhappy with both options, in the second scenario `util.EnableLog` will be ignored and might lead to confusion.

Any thoughts?

hsluoyz:
Remove the global ``EnableLog`` variable. Define it in ``DefaultLogger`` instead. The Casbin enforcer's ``EnableLog()`` function will call into the Logger's ``EnableLog()`` function. Other implementations can choose how to react on the ``EnableLog()`` call.

llonchj:
Thanks for your clarification @hsluoyz. Pushed changes

hsluoyz:
In fact, you don't need to start a new PR.. Rewrite the commits in this PR and this page will be updated automatically.

PR #159, opened 2018-10-30
logger interface

llonchj:


hsluoyz:
After I merged it, I have it's not exactly how I imagine it. The ``Logger`` instance should be a member of ``Enforcer``. ``SetLogger()`` and ``GetLogger()`` should be member functions of ``Enforcer`` instead of global. The advantage is the user can have different logging systems for different Casbin enforcers. And this also conforms to the convention for other Casbin components like Adapter, RoleManager, Watcher, etc. So can you make another PR to fix it?

llonchj:
I agree with you. That was my motivation with the Logger interface.

After seeing you moved the logic to log package, can I suggest a review of the implemented methods of the interface.

Most logging packages allow log level (zap, logrus, zerolog, ...), Golang log package supports Print and Panic. I appreciate the simplicity of golang log model, but we are opening the logger to others. I am inclined to use log level commands as they will be more descriptive.

What are your thoughts?

PR #162, opened 2018-11-14
[enhancement/#96] Skip filtered policy init

### LINKED ISSUE #96: LoadPolicy() is called automatically when initializing with filtered adapter
When I call `enforcer.InitWithAdapter()`, it results in `adapter.LoadPolicy()` being called via `enforcer.InitWithModelAndAdapter()`, even when the adapter is of the type `persist.FilteredAdapter`.

This seems to defeat the purpose of using a filtered adapter for performance considerations. I could check whether the adapter is a filtered adapter inside `adapter.LoadPolicy()` and not query the database if it's filtered. Is this the intended solution?

faceless-saint:
Do not load the full policy when initializing a filtered adapter.

Closes #96

### DIFF
--- a/enforcer.go
+++ b/enforcer.go
@@ -121,7 +121,9 @@ func (e *Enforcer) InitWithModelAndAdapter(m model.Model, adapter persist.Adapte
 
 	e.initialize()
 
-	if e.adapter != nil {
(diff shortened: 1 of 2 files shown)


hsluoyz:
lgtm

PR #164, opened 2018-11-19
add support for glob operator

dafanasiev:
I think that glob operator more simple than regex for end users. Can u add it?

### DIFF
--- a/model/function.go
+++ b/model/function.go
@@ -35,6 +35,7 @@ func LoadFunctionMap() FunctionMap {
 	fm.AddFunction("keyMatch2", util.KeyMatch2Func)
 	fm.AddFunction("regexMatch", util.RegexMatchFunc)
 	fm.AddFunction("ipMatch", util.IPMatchFunc)
+	fm.AddFunction("globMatch", util.GlobMatchFunc)
 
 	return fm
 }
--- a/util/builtin_operators.go
+++ b/util/builtin_operators.go
@@ -20,6 +20,7 @@ import (
 	"strings"
 
 	"github.com/casbin/casbin/rbac"
+	"github.com/gobwas/glob"
 )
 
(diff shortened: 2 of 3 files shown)


slaskis:
Would it work as well with the built in https://golang.org/pkg/path/#Match and thus avoid another dependency?

dafanasiev:
`path.Match` doesn't support "double-start" (`**`) operator, and support only `/` as separator.

hsluoyz:
I agree with @slaskis , is there a way to avoid another dependency? like implementing it with one or several source files? Casbin is a library which will be used by other projects, so I keep the size of dependencies as small as possible. Currently, Casbin only depends on ``govaluate``, which is the expression interpreter.

Pokerkoffer:
Any updates on this?

hsluoyz:
@Pokerkoffer how to avoid another dependency?

hsluoyz:
I'm revisiting this PR. How many people are there that need glob as built-in function?

mousedownmike:
Would it be possible to use the [path.Match()](https://golang.org/pkg/path/#Match) in lieu of `globmatch`?

I come from a Spring Security background where they make heavy use of "Ant Style" [path matching](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/util/AntPathMatcher.html).  That level of complexity can be problematic in a security situation but I think some level of globbing would be useful for broad controls of REST APIs.

hsluoyz:
I have asked the Golang community about built-in glob support: https://github.com/golang/go/issues/37964

hsluoyz:
Finally, a built-in glob matching is integrated into Casbin: https://github.com/casbin/casbin/pull/397

So I will close this one now. Let me know if there are any questions.

Issue #167, opened 2018-11-27
Would you like to implement new function: GetUsersForRoleInDomain


CHCP:
Hi hsluoyz:
  I found some useful functions like: ``GetRolesForUserInDomain, GetRolesForUser, GetUsersForRole``, but there is no ``GetUsersForRoleInDomain``.  Would you like to implement new function ``GetUsersForRoleInDomain``? This is will very useful. Thank you.

![image](https://user-images.githubusercontent.com/5645296/49054887-e26b0900-f249-11e8-8b74-fe71f81e26f2.png)

hsluoyz:
I vote for your idea. Can you make the PR?

CHCP:
I will try this, give me some days.

CHCP:
Hi hsluoyz,
 I implement the function GetUsersForRoleInDomain() in source file rbac_api_with_domains.go as follows:

``
// GetUsersForRoleInDomain gets the users that has a role inside a domain.
func (e *Enforcer) GetUsersForRoleInDomain(name string, domain string) []string {
	res, _ := e.model["g"]["g"].RM.GetUsers(name, domain)
	return res
}
``

But when I test this function, this is nothing return, and there is nothing return when I call  ``GetUsersForRole(name string, domain string)``  either. I confirm there are some users has this role in Domain.

hsluoyz:
Fixed in: https://github.com/casbin/casbin/pull/170

CHCP:
Very good, thank you.

PR #170, opened 2018-11-30
add GetUsersForRoleInDomain and test code

CHCP:
I implement ``GetUsersForRoleInDomain``, but the test is failed. I think the function: ``GetUsersForRole`` does not work. I don't know why, I am not familiar with casbin core, would you like finding this reason?

### DIFF
--- a/rbac_api_with_domains.go
+++ b/rbac_api_with_domains.go
@@ -14,6 +14,12 @@
 
 package casbin
 
+// GetUsersForRoleInDomain gets the users that has a role inside a domain. Add by Gordon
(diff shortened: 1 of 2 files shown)


hsluoyz:
I have fixed your bug in this commit: https://github.com/casbin/casbin/commit/99df23ebbdb11ada28cf6c7e0cbae8dfdc5b8e23

BTW, it may not work as you expect in the test: when the policy rule ``g, alice, admin, domain1`` is deleted, ``alice`` will also be gone, as Casbin doesn't track a user list. It just try to extract the list of users from all ``g`` policy rules.

hsluoyz:
Fix: https://github.com/casbin/casbin/issues/167

PR #235, opened 2019-05-31
make code more elegant and efficient

Wang-Kai:
Signed-off-by: Steve <persistence2013@sina.com>

As the [golang wiki](https://github.com/golang/go/wiki/Range#gotchas) said

>the value property of range (stored here as item) is a copy of the value from items

So if we want to modify the value of `Slice`,  the for loop is more suitable in my mind.


hsluoyz:
lgtm

PR #260, opened 2019-07-09
LoadPolicyLine does not deal with duplication

obashistu:
The current utility function LoadPolicyLine does not care whether a policy line has already been added or not which in some cases can cause problems. To address this the new function  LoadPolicyLineUnique uses the model.AddPolicy function to add new policy lines. Additionally common code between the functions has been pulled into its own function.


hsluoyz:
You PR title and text both contain …  Can you correct them so I can read them all?

obashistu:
Done

hsluoyz:
The CI has failed some test cases, can you fix it?

PR #262, opened 2019-07-10
Handle zero length policy params in the enforcer code

obashistu:
- Ensure policy param args are not zero before attempting to address slice values. 

- Run Go fmt on source


hsluoyz:
lgtm.

PR #264, opened 2019-07-10
Improvement/stop using NewModel constructor

obashistu:
The misplaced and strange NewModel constructor (it panics and is only ever used to differentiate between a filepath and a text) in enforcer can be bypassed entirely by using the safer model load functions dependent on whether it's from a path or a string. 

Since it's public the function remains but the logic therein is weird.


hsluoyz:
Similar to my previous suggestion.

obashistu:
Not to worry. Looking at the code we wont progress our project using Casbin. Closed all open PR's.

Have a good day.

vtolstov:
@obashistu what you want to use instead of casbin?

obashistu:
Hi @vtolstov initially Casbin looked interesting but after reviewing the code of the Go implementation we decided it would be safer to write our own. One positive that came from looking at Casbin closely is that we've decided to use the awesome https://github.com/Knetic/govaluate library to implement flexible policies.

PR #266, opened 2019-07-10
Improvement/model load

obashistu:
- A new Load function which will attempt to load from either a file or text using some basic logic (can be extended

- A new function LoadWithConfig which allows users to handle config load errors independently of model loads

- Improved section loading


hsluoyz:
Can you remove the first commit ``Merge pull request #1 from casbin/master ``? You can rebase your repo locally and this commit will be unnecessary. So does the ``go fmt`` commit. Make your commits at the final state for easier review.

PR #269, opened 2019-07-31
Test for a logger and use string.Builder

ctxnop:
This code shows a big performance issue when inserting many policies in
a row. This is because it builds a string to log without checking for a
logger (thus can build a string without using it).
Also concatening string in a loop cause a lot a memory allocation that
has to be handled by the garbage collector, which is the real
performance hit.

Using a string.Builder avoid this issue. Also check if the logger is
enabled before doing anything.

Change-Id: Ibb63237d6065e1d1a77bb290a493bf8715b150e3
Signed-off-by: Loïc Collignon <loic.collignon@iot.bzh>

### DIFF
--- a/rbac/default-role-manager/role_manager.go
+++ b/rbac/default-role-manager/role_manager.go
@@ -15,6 +15,7 @@
 package defaultrolemanager
 
 import (
+	"strings"
 	"sync"
 
 	"github.com/casbin/casbin/v2/errors"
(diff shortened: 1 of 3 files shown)


rbac/default-role-manager/role_manager.go
```diff
@@ -196,18 +197,21 @@ func (rm *RoleManager) GetUsers(name string, domain ...string) ([]string, error)
...
-		if text := value.(*Role).toString(); text != "" {
-			if line == "" {
-				line = text
-			} else {
-				line += ", " + text
+	if log.GetLogger().IsEnabled() {
+		var sb strings.Builder
+		rm.allRoles.Range(func(_, value interface{}) bool {
+			if text := value.(*Role).toString(); text != "" {
+				if sb.Len() > 0 {
```
ctxnop: Should be "==", do not merge right now

hsluoyz:
Hi @ctxnop , please merge your commits into one for easier review.

ctxnop:
That is exactly what I plane to do, but as soon as it's 3am where I live, it's not a goot idea to do it right now

ctxnop:
Should be good now

hsluoyz:
lgtm.

PR #280, opened 2019-08-26
add missing methods in management_api to SyncedEnforcer

umq:
some methods in management_api are missing in SyncedEnforcer
those methods can fail when called while loading.
this PR adds those methods.


hsluoyz:
lgtm

PR #298, opened 2019-10-03
used model.Function wherever suitable.

asutosh97:


hsluoyz:
I like this PR as it cleans the code, except it causes import cycle: https://travis-ci.org/casbin/casbin/jobs/592983313

Can you fix it?

PR #301, opened 2019-10-06
e.GetAll<>() now gets results across all ptypes

asutosh97:


management_api.go
```diff
@@ -48,7 +48,7 @@ func (e *Enforcer) GetAllNamedActions(ptype string) []string {
 
 // GetAllRoles gets the list of roles that show up in the current policy.
 func (e *Enforcer) GetAllRoles() []string {
-	return e.GetAllNamedRoles("g")
+	return e.model.GetValuesForFieldInPolicyAllPTypes("g", 1)
```
hsluoyz: This is not `PTypes`. Maybe rename it.

asutosh97: what shall be a suitable name?

hsluoyz: `GetValuesForFieldInPolicy` would be OK.

asutosh97: there is actually a function of that name, which returns values for a specific type.

PR #308, opened 2019-10-15
Properly indent sample code in doc

shihanng:
According to this [official blog](https://blog.golang.org/godoc-documenting-go-code) one of the formatting rules for Godoc is

> Pre-formatted text must be indented relative to the surrounding comment text

This PR makes the code formatted properly in Godoc. I've skimmed through the doc and find this is the only part that need fixing. Please do let me know if I miss something. Thanks!

Here is a screenshot on Godoc that shows how it will look after the fix:

Screen Shot 2019-10-15 at 13 32 22

### DIFF
--- a/enforcer.go
+++ b/enforcer.go
@@ -47,11 +47,16 @@ type Enforcer struct {
 }
 
 // NewEnforcer creates an enforcer via file or DB.
+//
 // File:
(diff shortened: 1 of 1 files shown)


hsluoyz:
lgtm.

Issue #309, opened 2019-10-15
Please fix the threading problem of the Enforce method of the SyncedEnforcer


Orefa:
An error occurred when the multi-threaded call was made to the Enforce of the SyncedEnforcer: "fatal error: concurrent map writes" 
```
fatal error: concurrent map writes
goroutine 2811 [running]:
runtime.
throw (0x 121354f，0x15)
E:/go/src/runtime/panic. go:774 +0x79 fp=0xc0001e6910 sp=0xc0001e68e0 pc=0x433bc9
runtime.mapassign faststr (0x10c9580， 0xc00013cf30， 0x11fca6c, 0x1，0xc0001b0998)
E:/go/src/runtime/map faststr. go:211 +0x41e fp=0xc0001e6978 sp=0xc0001e6910 pc=0x41548e
github.com/casbin/casbin/v2. (*Enfor cer).enforce (0xc0000b01c0 0x0， 0x0, 0xc0007753c0, 0x4, 0x4， 0xa6, 0xc0001e6d58, 0x59f 4e9)
E:/Work/GoCndWorkspace/pkg/mod/github.com/casbin/casbin/v2@v2.1.0/enforcer. go:343 +0x1b2 fp=0xc0001e6cd8 sp=0xc0001e6978 pc=0xcd9452
github.com/casbin/casbin/v2. (*Enforcer). Enforce (...
E:/Work/GoCmdWorkspace/pkg/mod/github.com/casbin/casbin/v2@v2.1.0/enforcer. go : 486
github.com/casbin/casbin/v2. (*SyncedEnforcer).Enforce (0xc00013c900, 0x0007753c0， 0x4, 0x4, 0xc00078ab0O, 0x0， 0x0)
E:/Work/GoCndWorkspace/pkg/mod/github.com/casbin/casbin/v2@v2.1.0/enforcer synced. go:111 +0xd7 fp=0xc0001e6d68 sp=0xc000 le6cd8 pc=0xcdbb27

```

hsluoyz:
Please give more details, like your code, your fix. Thanks.

Orefa:
@hsluoyz master file  enforcer_synced.go
 line 109  e.m.RLock() => e.m.Lock()
line 110 defer e.m.RUnlock() =>defer e.m.Unlock()

hsluoyz:
Can you make a PR?

Orefa:
> 
> 
> Can you make a PR?

[...]

PR #310, opened 2019-10-16
fix the threading problem of the Enforce method of the SyncedEnforcer

Orefa:
https://github.com/casbin/casbin/issues/309#issuecomment-542487705


hsluoyz:
lgtm

Issue #333, opened 2019-11-23
LoadPolicyLine panics


mousedownmike:
The following code will panic:

```Go
emptyModel, _ := model.NewModelFromString("")
persist.LoadPolicyLine("p, alice, data1, read", emptyModel)
```
I read #50 and #51 but I'm not sure what the best approach is to handling this condition.  If `""` is not a valid Model, it seems like `model.NewModelFromString("")` should return an error.  I encountered this while creating a minimal unit test for my own code but I could see how this could occur in a production scenario where an empty string was unintentionally used to load the model.

I have a PR with an `adapter_test.go` that illustrates the error but provides no solution.  I'm happy to contribute a fix but I'm still coming up to speed so I need more info:
1. Is having an empty model a valid case?
1. If not, where should this be prevented? `NewModelFromString`, `LoadModelFromText`, `NewConfigFromText`...
1. If it is valid, should `LoadPolicyLine` return an error indicating the nil state of the Model or should it Panic with a message regarding the state of the Model.

Thanks for your time.

hsluoyz:
1. No

2. Add it here:

https://github.com/casbin/casbin/blob/51dd9a02ef850b0b5a7a3fbbeffdf5d0c20509a8/model/model.go#L142-L151

Can you make a PR to fix it?

mousedownmike:
Will do.

PR #334, opened 2019-11-23
LoadPolicyLine panic demonstration

mousedownmike:
This test demonstrates the behavior described in #333.  This does not contain a solution, it will fail and should not be merged as-is.


hsluoyz:
Got it. Closed.

PR #335, opened 2019-11-26
Required Model Section Check

mousedownmike:
For #333 this PR checks for the existence of `r`,`p`,`e` and `m` model sections, returning an error when loading if any of those is not present.

This also has a minor typo correction in a comment and a move of the empty string check in `model.AddDef()`.


persist/adapter_test.go
```diff
@@ -0,0 +1,22 @@
+package persist
+
+import (
+	"github.com/casbin/casbin/v2/model"
+	"testing"
```
hsluoyz: Sort the imports, see other files.

persist/adapter_test.go
```diff
@@ -0,0 +1,22 @@
+package persist
```
hsluoyz: License header missing.

model/model.go
```diff
@@ -140,16 +144,22 @@ func (model Model) LoadModelFromText(text string) error {
...
-	loadSection(model, cfg, "m")
-
-	loadSection(model, cfg, "g")
-
+	for s := range sectionNameMap {
+		loadSection(model, cfg, s)
+	}
+	for _, rs := range requiredSections {
+		if !model.hasSection(rs) {
+			return errors.New("model does not contain all required sections ")
```
hsluoyz: Show the missing section's name, remove trailing space.

model/model_test.go
```diff
@@ -1,9 +1,102 @@
+// Copyright 2019 The casbin Authors. All Rights Reserved.
```
hsluoyz: Put them into a new `config_test.go` file.

mousedownmike: I think you're referring to the `MockConfig` I've defined here.  I was using that to isolate the tests without using a complete `config.ConfigInterface` implementation.  If that gets moved to `config_test.go` it no longer becomes usable in this test.  I haven't found a consistent pattern for where people place their mocks in Go.  Would moving the Mock definitions to the bottom of the file be better?

mousedownmike:
Apologies for the `adapter_test.go` file getting wrapped up in this PR.  I've improved the error message contents when sections are missing.  I need clarification on your `config_test.go` comment, maybe just move the mocks to the bottom of the file?

I merged in an additional test I created for `model.AddDef`, I was using it to verify my value check change but it didn't get merged in earlier.

mousedownmike:
@hsluoyz  let me know if you need to make any additional changes here, I'd be happy to.

hsluoyz:
Hi @mousedownmike please rework your commits. Any "Merge", "Need to revert" commits should be replaced.

image

hsluoyz:
Just provide the final-version, good-looking, easy-to-review commits

mousedownmike:
That makes sense but I'm not sure how to do that with an existing PR in GitHub so I'm going to close this PR and open a new one with the commits squashed.

PR #340, opened 2019-12-07
LoadPolicyLine Required Section Check

mousedownmike:
For #333 this PR checks for the existence of r,p,e and m model sections, returning an error when loading if any of those is not present and indicates which required sections are missing.

This also has a minor typo correction in a comment and a move of the empty string check in model.AddDef().


hsluoyz:
Fix: https://github.com/casbin/casbin/issues/333

hsluoyz:
lgtm

PR #361, opened 2020-02-13
Fix syntax errors on multiline config sentences

aroldanz:
- The actual code strips the multiline char but sometimes
it leads to syntax errors because not compatible chars may
result appended.


hsluoyz:
lgtm

PR #362, opened 2020-02-16
Go idomatic stopping goroutine for auto load policy

Dasio:
Goroutine which is handling auto load policy is immediately exited after user requested to stop it, instead of sleeping.

After I commited, I realized maybe one issue?
StopAutoLoadPolicy will be blocking operation if there is no running auto load, can be solved again adding bool flag. But why user would want stop auto load policy if he didn't start it before.

### DIFF
--- a/enforcer_synced.go
+++ b/enforcer_synced.go
@@ -26,8 +26,9 @@ import (
 // SyncedEnforcer wraps Enforcer and provides synchronized access
 type SyncedEnforcer struct {
 	*Enforcer
-	m            sync.RWMutex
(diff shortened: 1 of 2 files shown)


Dasio:
I wanted to amend last commit.
Anyway, any idea how to keep code coverage? Not sure, why branch with `<- ticker.C` is not reached in CI. I tested it locally and it works

hsluoyz:
@Dasio please merge commits with the same name.

hsluoyz:

nodece:
@hsluoyz lgtm.

PR #364, opened 2020-02-20
fix(rbac_api.go): DeleteUser should remove Policies + Grouping Policies

Sefriol:
Previously it just removed Grouping policies. Now it shares same implementation as DeleteRole
casbin/node-casbin#118


hsluoyz:
lgtm

PR #369, opened 2020-02-26
Extract Enforcer Interface and make Enforcers Chainable

niallnsec:
Hi,
I'm not sure if this is in line with your plans for the project but I had a need to make these changes as part of a larger project I am working on and thought I would submit a pull request in case they are useful to others.

The purpose of the changes was to generalise the Enforcer API into an interface definition. I split the interface into two parts for convenience, the BasicEnforcer and APIEnforcer.
BasicEnforcer defines the basic set of functions for an object to be considered a valid enforcer. APIEnforcer defines the full enforcer API set, including RBAC and Management functions.

The reason for the extraction of an interface was to allow me to chain different enforcers together. Specifically, I wanted to take advantage of the new cached enforcer as well as the synced enforcer at the same time.

Rather than just creating a custom enforcer inside out of the two existing ones I thought it would be better to make the enforcers chainable. That has allowed me to also create a custom Enforcer wrapper inside of a private repository to provide audit logging. Enabling caching, audit logging or any other feature someone might want to implement can now be achieved by chaining the relevant Enforcer implementations together.

I believe (unless I have made a mistake) that none of the changes are API breaking, and only add extra functionality.

I ran benchmarks with and without this chaining ability and there was no observable difference in performance, so I believe any effects are negligible. 

I also added some example code showing how you might wish to chain enforcer implementations together.

Finally, I added an extra method GetRootEnforcer which returns the underlying Enforcer instance at the end of the chain. This is to ensure a user can still access the largest feature set if implementations further up the chain do not implement the APIEnforcer interface. To make this function I added a function to the BasicEnforcer interface called GetParentEnforcer() which retrieves the next lower enforcer in the chain, with the root Enforcer implementation returning nil to indicate the end of the chain.

No worries if this is not the sort of thing you are looking for, I realise there are some significant changes, but thought I'd share just in case :) The Casbin project has proved incredibly useful for me and my company so I am incredibly grateful for all the hard work that has gone into it!

### DIFF
--- a/enforcer.go
+++ b/enforcer.go
@@ -342,10 +342,10 @@ func (e *Enforcer) BuildRoleLinks() error {
 }
 
 // enforce use a custom matcher to decides whether a "subject" can access a "object" with the operation "action", input parameters are usually: (matcher, sub, obj, act), use model matcher by default when matcher is "".
-func (e *Enforcer) enforce(matcher string, rvals ...interface{}) (bool, error) {
+func (e *Enforcer) enforce(matcher string, rvals ...interface{}) (ok bool, err error) {
 	defer func() {
-		if err := recover(); err != nil {
-			fmt.Errorf("panic: %v", err)
+		if x := recover(); x != nil {
+			err = fmt.Errorf("panic: %v", x)
 		}
 	}()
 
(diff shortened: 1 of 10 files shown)


hsluoyz:
Hi,

Two issues for now:

No merge commits in PR:

image

Moreover, coveralls dropped too much, 14%.

hsluoyz:
@niallnsec

hsluoyz:
You should wrap the enforcer into a chain or not. It should not cause modification into Casbin core code. Enforcer should not be an interface.

PR #373, opened 2020-03-04
Fix wrong return types for methods in IEnforcer

dovics:
In IEforce, some methods are missing return value error

### DIFF
--- a/management_api.go
+++ b/management_api.go
@@ -14,6 +14,8 @@
 
 package casbin
 
+import "github.com/Knetic/govaluate"
+
 // GetAllSubjects gets the list of subjects that show up in the current policy.
 func (e *Enforcer) GetAllSubjects() []string {
 	return e.model.GetValuesForFieldInPolicyAllTypes("p", 0)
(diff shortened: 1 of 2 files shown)


management_api.go
```diff
@@ -256,6 +254,6 @@ func (e *Enforcer) RemoveFilteredNamedGroupingPolicy(ptype string, fieldIndex in
 }
 
 // AddFunction adds a customized function.
-func (e *Enforcer) AddFunction(name string, function govaluate.ExpressionFunction) {
+func (e *Enforcer) AddFunction(name string, function func(args ...interface{}) (interface{}, error)) {
```
hsluoyz: Why do this?

nodece: It seems to be for mock.

dovics: The interface cannot automatically convert equal types.  The two function types are  same.Or we can also change the type in IEnforce

hsluoyz: I prefer `govaluate.ExpressionFunction` because it's easy to know where it comes from, even if it needs to import a lib.

dovics: Ok i'll modify it right away

hsluoyz: Sorry. Does this change meaning that any user that wants to write a custom function has to import govaluate? If yes, please revert back.

hsluoyz: I don't want to cause any trouble for our users.

dovics: Of course not, the compiler will only have problems with matching function type. It will be automatically corrected in parameter passing. Users only need to do the same as before.

PR #377, opened 2020-03-08
Clean up temporarily created role

dovics:
In  `hasLink()`,  if role has not been created before, a temporary one will be created for this query, but it is not cleaned up at the end. If there are too many queries, it will take up extra memory


hsluoyz:
@nodece is this necessary?

nodece:
@hsluoyz I don't think it's necessary.

nodece:
It has a small impact on performance.

hsluoyz:
OK. I will close it for now. @dovics if you have questions, please let us know.

PR #382, opened 2020-03-12
ADD HasImplicitPermissionForUserInDomain() and HasImplicitPermissionForUser()

richsoap:
### HasPermissionAmongUsers
Both **HasImplicitPermissionForUserInDomain** and **HasImplicitPermissionForUser** need to go through a name list to find if a permission belongs to one of them. Therefore, I write a function to do these things.
I add a new role in **basic_without_resources_policy** to test this function better. 
### HasImplicitPermissionForUser
This function is for rbac-wthout-domains model. First, we get every implicit role this user can reach. And use **HasPermissionAmongUsers** to determine whether the permission is its implicit permission or not.
### HasImplicitPermissionForUserInDomain
First, I check if the domain is in the permission to avoid case like using role in domain1 to read things in domain2, because of we have no idea which param in permission is domain. The next process is the same as **HasImplicitPermissionForUser**.

### DIFF
--- a/examples/basic_without_resources_policy.csv
+++ b/examples/basic_without_resources_policy.csv
@@ -1,2 +1,3 @@
 p, alice, read
 p, bob, write
+p, carol, read
--- a/go.mod
+++ b/go.mod
@@ -1,3 +1,5 @@
 module github.com/casbin/casbin/v2
 
+go 1.14
+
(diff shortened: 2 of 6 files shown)


nodece:
Hi @richsoap , Sorry  for the late reply. Why we doesn't use `enforce`?

Issue #386, opened 2020-03-19
Add batch operations: AddPolicies() and RemovePolicies() to Management API


hsluoyz:
A lot of users require this feature.

hsluoyz:
Fixed by: https://github.com/casbin/casbin/pull/388

PR #388, opened 2020-03-19
Added multiple add and remove policies API functions.

DivyPatel9881:
Usage of the following functions is as follows :

```go
rule1 := []interface{} {"jack", "data4", "read"}
rule2 := []interface{} {"katy", "data4", "write"}
rule3 := []interface{} {"leyo", "data4", "read"}
rule4 := []interface{} {"ham", "data4", "write"}

e.AddPolicies(rule1, rule2, rule3, rule4)

e.RemovePolicies(rule1, rule2, rule3, rule4)

```

Have also added tests for these functions, all the tests pass.
Please let me know if any changes required, this is open to changes if required any.


DivyPatel9881:
Fixes #386

DivyPatel9881:
In the code, ```result``` is the individual status of added or removed policy rule, these individual results are then appended to the ```results``` list which is being returned to get statuses for each ```rule``` mentioned in the arguments.

DivyPatel9881:
Can anyone review this?

management_api.go
```diff
@@ -122,6 +122,22 @@ func (e *Enforcer) AddPolicy(params ...interface{}) (bool, error) {
 	return e.AddNamedPolicy("p", params...)
 }
 
+// AddPolicies adds authorization rules to the current policy.
+// If the rule already exists, the function returns false for the corresponding rule and the rule will not be added.
+// Otherwise the function returns true for the corresponding rule by adding the new rule.
+func (e *Enforcer) AddPolicies(rules ...interface{}) [][] interface{} {
```
nodece: I think using `[][]interface{}` is more better.
WDYT?

management_api.go
```diff
@@ -142,6 +158,20 @@ func (e *Enforcer) RemovePolicy(params ...interface{}) (bool, error) {
 	return e.RemoveNamedPolicy("p", params...)
 }
 
+// RemovePolicies removes authorization rules from the current policy.
+func (e *Enforcer) RemovePolicies(rules ...interface{}) [][] interface{} {
```
nodece: Same as above.

nodece:
You should add `addPolicies` and `removePolicies` to [internal_api.go](https://github.com/casbin/casbin/blob/master/internal_api.go).

management_api.go
DivyPatel9881: Actually, it would be better if we use like this, don't you think as the users can always use spread operator if they have rules in ```[][] interface{}```, like
```
rules = [][]interface{} {
                   []interface{} {"jack", "data4", "read"},
                   []interface{} {"katy", "data4", "write"},
                   []interface{} {"leyo", "data4", "read"},
                   []interface{} {"ham", "data4", "write"}
                }
```
And then can be used like,
```
e.AddPolicies(rules...)
e.RemovePolicies(rules...)
```

DivyPatel9881:
```You should add addPolicies and removePolicies to internal_api.go.```  Is it necessary to add those functions in internal_api, as ```AddPolicies()``` and ```RemovePolicies()``` can internally use the atomic functions of internal_api ```addPolicy()``` and ```removePolicy()``` respectively, for their implementations.
What do you think?

File: management_api.go (hunk 1 above)
Comment (DivyPatel9881) [reply management_api.go]: What do you think @nodece should I leave this as it is, or should I make changes as you said?

Comment (hsluoyz) [reply management_api.go]: What's the difference between []interface{} and [][]interface{}?

Comment (DivyPatel9881) [reply management_api.go]: Hi @hsluoyz, the ```[] interface{} ``` can consists of ```array``` or ```primitive datatypes``` like ```string```, ```int``` or ```float```, but the ```[][]interface{} array``` must only consists of any ```nested array```, it cannot contain a ```primitive data-type```.
Right.

Comment (nodece) [reply management_api.go]: I pretty, like:
```
func (e *Enforcer) AddPolicies(params [][]interface{}) (bool, error)
```
```
rules := [][]interface{}{
	{"user","/","get"},
	{"user", "/", "post"},
	{"user","/","put"},
}

e.AddPolicies(rules)
```

Comment (hsluoyz) [reply management_api.go]: Then I agree with @nodece

Comment (nodece) [reply management_api.go]: I forgot that `interface{}` can contain `arrays`.

Comment (DivyPatel9881) [reply management_api.go]: I actually tried maintaining the structure of the casbin, as every function used variadic arguments, so I thought it would be better if I use such arguments instead of passing a 2-D array. But its okay @nodece I will make the changes if @hsluoyz is also okay with this.
Thanks @nodece and @hsluoyz.

Comment (nodece) [reply management_api.go]: @hsluoyz Why don't we use `[][]string`?

Comment (nodece) [reply management_api.go]: Each policy  is `[]string`.

Comment (DivyPatel9881) [reply management_api.go]:
```
rule1 := []interface{} {"jack", "data4", "read"}
rule2 := []interface{} {"katy", "data4", "write"}
rule3 := []interface{} {"leyo", "data4", "read"}
rule4 := []interface{} {"ham", "data4", "write"}

e.AddPolicies(rule1, rule2, rule3, rule4)

e.RemovePolicies(rule1, rule2, rule3, rule4)
```
What do you think, in this case, variadic arguments won't be better?

Comment (DivyPatel9881) [reply management_api.go]: > Each policy is `[]string`.

Oh, yeah thats better, I feel.

Comment (DivyPatel9881) [reply management_api.go]: Agreed @nodece this is better signature.
```func AddPolicies(rules ...[]string) [][]interface```

Comment (DivyPatel9881) [reply management_api.go]: Then, I think this must also be changed in the management_api.go
```func (e *Enforcer) AddPolicy(params ...interface{}) (bool, error) ``` to ```func (e *Enforcer) AddPolicy(params ...string) (bool, error) ```

Comment (nodece) [reply management_api.go]: > ```
> rule1 := []interface{} {"jack", "data4", "read"}
> rule2 := []interface{} {"katy", "data4", "write"}
> rule3 := []interface{} {"leyo", "data4", "read"}
> rule4 := []interface{} {"ham", "data4", "write"}
> 
> e.AddPolicies(rule1, rule2, rule3, rule4)
> 
> e.RemovePolicies(rule1, rule2, rule3, rule4)
> ```
> 
> What do you think, in this case, variadic arguments won't be better?

It's not friendly for users. users pretty to pass 2D array.

Comment (DivyPatel9881) [reply management_api.go]: If a user wants to pass 2D array, then it can be used like this ```e.AddPolicies(rules...)```

Comment (DivyPatel9881) [reply management_api.go]: Its best of both the worlds.

Comment (DivyPatel9881) [reply management_api.go]: @nodece should I make changes from interface{} to string ?

Comment (nodece) [reply management_api.go]: If I'm a user, I want to use `e.AddPolicies(rules)` instead of `e.AddPolicies(rules...)`

Comment (nodece) [reply management_api.go]: > @nodece should I make changes from interface{} to string ?

I want to hear what @hsluoyz.

Comment (nodece) [reply management_api.go]: > @nodece should I make changes from interface{} to string ?

I want to listen from @hsluoyz .

**Comment by nodece** [comment]
> `You should add addPolicies and removePolicies to internal_api.go.` Is it necessary to add those functions in internal_api, as `AddPolicies()` and `RemovePolicies()` can internally use the atomic functions of internal_api `addPolicy()` and `removePolicy()` respectively, for their implementations.
> What do you think?

When policy changes, Casbin needs the save policy and notify other instances.  Could you improve it?

File: management_api.go (hunk 1 above)
Comment (hsluoyz) [reply management_api.go]: No one will call `e.AddPolicies(rule1, rule2, rule3, rule4)`.

`e.AddPolicies(rules)` is better than `e.AddPolicies(rules...)`

**Comment by hsluoyz** [comment]
The adapter interface needs to have optional handlers: `AddPolicies()` and `RemovePolicies()` too.

File: management_api.go (hunk 1 above)
Comment (nodece) [reply management_api.go]: The `[][]string` as parameter is better. @hsluoyz WDYT?

File: enforcer_cached.go [hunk 3]
```diff
@@ -80,7 +80,7 @@ func (e *CachedEnforcer) getCachedResult(key string) (res bool, ok bool) {
 	e.locker.RLock()
 	defer e.locker.RUnlock()
 	res, ok = e.m[key]
-	return res, ok
+	return
```
hsluoyz: Revert this.

DivyPatel9881: No, its okay, please check the function signature it already returns the parameters, the following return params are redundant.

DivyPatel9881:
```
func (e *CachedEnforcer) getCachedResult(key string) (res bool, ok bool)
```
Please check this.

DivyPatel9881: I never commit and push syntactical errors in my code.

DivyPatel9881:
I have added the functions to the model, adapter, and internal_api, as Travis CI is passing which means all the test-cases are passing, the code-climate issues are not significant as they are inevitable to getting solved based on current code model.

DivyPatel9881:
Adding more and more tests will increase the coverage.

DivyPatel9881:
@nodece please, review this.

internal_api.go
```diff
@@ -43,6 +43,41 @@ func (e *Enforcer) addPolicy(sec string, ptype string, rule []string) (bool, err
...
+				if err.Error() != notImplemented {
+					errs[i] = err
+				}
+			}
+		}
+		return rulesAdded, errs
+	}
+
+	if e.watcher !=nil && e.autoNotifyWatcher {
+		err := e.watcher.Update()
```
nodece: Please move the `watcher` to `for` outside.

internal_api.go
```diff
@@ -68,6 +103,40 @@ func (e *Enforcer) removePolicy(sec string, ptype string, rule []string) (bool,
...
+				if err.Error() != notImplemented {
+					errs[i] = err
+				}
+			}
+		}
+		return rulesRemoved, errs
+	}
+
+	if e.watcher !=nil && e.autoNotifyWatcher {
+		err := e.watcher.Update()
```
nodece: Same as above.

nodece:
@DivyPatel9881 Think about a question, what to do if adding a policy fails.

internal_api.go
DivyPatel9881: Sorry, didn't got it? Can you come again. Thank you

nodece: It means that after all the policy have been added then call the `e.watcher.update()`

DivyPatel9881:
> @DivyPatel9881 Think about a question, what to do if adding a policy fails.

Yeah, this scenario includes in my solution. It checks for all rule policies and gets individual status and errors for each addition or removal of a policy rule. Hence, the return information is quite getting track of each policy rule whether it is a valid policy rule to add or remove or not.

internal_api.go
DivyPatel9881: Yeah, thanks @nodece , its exactly what I have done.

DivyPatel9881:
@hsluoyz and @nodece, the tests attached to this PR covers each line of code written by me. Tests cover policy rules addition success and failure, removal success and failure.

management_api.go
```diff
@@ -160,6 +191,15 @@ func (e *Enforcer) RemoveNamedPolicy(ptype string, params ...interface{}) (bool,
 	return e.removePolicy("p", ptype, policy)
 }
 
+// RemoveNamedPolicies removes authorization rules from the current named policy.
+func (e *Enforcer) RemoveNamedPolicies(ptypes []string, rules [][] string) ([]bool, []error) {
```
nodece: `ptypes string` is more better.  what do you think?

persist/adapter.go
```diff
@@ -46,9 +46,15 @@ type Adapter interface {
 	// AddPolicy adds a policy rule to the storage.
 	// This is part of the Auto-Save feature.
 	AddPolicy(sec string, ptype string, rule []string) error
+	// AddPolicies adds policy rules to the storage.
+	// This is part of the Auto-Save feature.
+	AddPolicies(secs []string, ptypes []string, rules [][]string) []error
```
nodece: It's completely unnecessary.

persist/adapter.go
```diff
@@ -46,9 +46,15 @@ type Adapter interface {
...
 	// RemovePolicy removes a policy rule from the storage.
 	// This is part of the Auto-Save feature.
 	RemovePolicy(sec string, ptype string, rule []string) error
+	// RemovePolicies removes policy rules from the storage.
+	// This is part of the Auto-Save feature.
+	RemovePolicies(secs []string, ptypes []string, rules [][]string) []error
```
nodece: It's completely unnecessary.

persist/file-adapter/adapter.go
```diff
@@ -113,11 +113,29 @@ func (a *Adapter) AddPolicy(sec string, ptype string, rule []string) error {
...
+	return errs
+}
+
 // RemovePolicy removes a policy rule from the storage.
 func (a *Adapter) RemovePolicy(sec string, ptype string, rule []string) error {
 	return errors.New("not implemented")
 }
 
+// RemovePolicies removes policy rules from the storage.
+func (a *Adapter) RemovePolicies(secs []string, ptypes []string, rules [][]string) []error {
```
nodece: I feel better, something like: `func (a *Adapter) RemovePolicies(secs []string, ptypes []string, rules [][]string) []error`.

management_api.go
DivyPatel9881: What if ptype is different for some rules, in such cases we want ptype for each corresponding policy rule.

DivyPatel9881: So ```ptypes []string``` fits well.

persist/file-adapter/adapter.go
hsluoyz: Agree with @nodece , we only have two ptypes: p and g. The user call it at most twice. This is OK.

DivyPatel9881: Same reason as above, we want ```[]string```.

persist/adapter.go
hsluoyz: Make sure these two functions are optional for an adapter.

persist/adapter.go
hsluoyz: Make sure these two functions are optional for an adapter.

management_api.go
hsluoyz: Agree with @nodece , we only have two ptypes: p and g. The user call it at most twice. This is OK.

DivyPatel9881: Hey @hsluoyz what about g1, I saw it in documentation, is that a valid type?
And what if casbin's supports more ptype in future. I was thinking of future scope. If its not okay, then I can change it no problem.

persist/adapter.go
DivyPatel9881: Okay,  got it will change this.

persist/adapter.go
DivyPatel9881: Okay will make changes.

management_api.go
DivyPatel9881: So what should I do?

nodece: Uses can only add policy of the same type.

hsluoyz: Agree with @nodece

persist/adapter.go
DivyPatel9881: There is a problem, if I remove these functions from the interface, I will be unable to use these functions for any adapter unless I make type assertions.
For example, ```e.adapter.(fileadapter).AddPolicies()``` this will work well, but if used like this ```e.adapter.AddPolicies()``` will throw an error. Golang doesn't support [dynamic type assertion](https://stackoverflow.com/questions/51504706/dynamic-type-assertion-golang) and we require this feature to get implemented on more than one adapters. 
There are two solutions either we add those functions in the interface or we go with a workaround like implementing a switch-case for each adapter which gets nasty.
So, what should we do?

DivyPatel9881: Any idea?

DivyPatel9881: As we know, ```e.adapter``` is an interface, if we remove those function signatures from there, it will not be able to access the functions at run-time.

DivyPatel9881: @nodece and @hsluoyz can you think of any way out of this, if we do not add these functions?

management_api.go
DivyPatel9881: Okay, will make the changes.

persist/adapter.go
hsluoyz: Define a new interface called `AdapterEx`

nodece: We should not add new methods to the `Adapter`. Just use like `AddPolicy` is good.

DivyPatel9881: > Define a new interface called `AdapterEx`. AdapterEx is a superset of Adapter.

Nice, working on this.

DivyPatel9881: > We should not add new methods to the `Adapter`. Just use like `AddPolicy` is good.

Yeah got it, thanks

DivyPatel9881:
I have made the required changes, and have added ```AdapterEx``` interface which just getting called during type assertion in internal_api. Please review it, I think it's okay to merge now.

persist/adapter_ex.go
```diff
@@ -0,0 +1,43 @@
...
+// limitations under the License.
+
+package persist
+
+import (
+	"github.com/casbin/casbin/v2/model"
+)
+
+// AdapterEX is the interface for Casbin adapters with multiple add and remove policy functions.
+type AdapterEx interface {
```
hsluoyz: AdapterEx should extend Adapter. Do not make another copy for Adapter functions.

persist/adapter_ex.go
```diff
@@ -0,0 +1,43 @@
+// Copyright 2017 The casbin Authors. All Rights Reserved.
```
hsluoyz: It's 2020 now

persist/adapter_ex.go
```diff
@@ -0,0 +1,43 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package persist
+
+import (
+	"github.com/casbin/casbin/v2/model"
```
hsluoyz: One import doesn't have brackets

util/builtin_operators.go
```diff
@@ -197,7 +197,7 @@ func IPMatch(ip1 string, ip2 string) bool {
 		}
 
 		return objIP1.Equal(objIP2)
-	}
+	}
```
hsluoyz: What's this?

persist/file-adapter/adapter.go
```diff
@@ -113,11 +113,29 @@ func (a *Adapter) AddPolicy(sec string, ptype string, rule []string) error {
 	return errors.New("not implemented")
 }
 
+// AddPolicies adds policy rules to the storage.
+func (a *Adapter) AddPolicies(sec string, ptype string, rules [][]string) []error {
```
hsluoyz: We should only return one error, similar for others.

If one add failed, we should roll back previous ones. So it's atomic operation.

nodece:
@hsluoyz and @DivyPatel9881 Why needs the `AdapterEx`?

enforcer_interface.go
```diff
@@ -89,10 +89,14 @@ type IEnforcer interface {
 	HasPolicy(params ...interface{}) bool
 	HasNamedPolicy(ptype string, params ...interface{}) bool
 	AddPolicy(params ...interface{}) (bool, error)
+	AddPolicies(rules [][]string) ([]bool, []error)
```
hsluoyz: You missed 4 functions: AddGroupingPolicies, AddNamedGroupingPolicies, RemoveNamedGroupingPolicies, RemoveNamedGroupingPolicies

hsluoyz:
> @hsluoyz and @DivyPatel9881 Why needs the `AdapterEx`?

Because if we put them into Adapter directly, all current adapter implementations will fail, because they don't implement the interface.

hsluoyz:
We should rename `AdapterEx` to `BatchAdapter`, see how it works for `FilteredAdapter`: https://github.com/casbin/casbin/blob/master/persist/file-adapter/adapter_filtered.go

enforcer_interface.go
nodece: Use the method of `Adapter` is enough.

nodece:
I think use the method of `Adapter` is enough.

persist/file-adapter/adapter.go
nodece: Rollback should not exist. It increases complexity. We just returns an error is better.

DivyPatel9881: > We should only return one error, similar for others.
> 
> If one add failed, we should roll back previous ones. So it's atomic operation.

To add atomicity(add or remove everything or nothing) to the operations, it occurs to me, there are two solutions to tackle this :
1. We add or remove the policy rule one by one and when we get an invalid policy rule we remove or add the previously added or removed policy rules respectively.
OR
2. Introduce the function which checks the validity of the policy rules if all the policy rules are valid then we move forward for the operation or we do not and pass the error.

I think the 2nd solution is better, what do you think?

DivyPatel9881: > Rollback should not exist. It increases complexity. We just returns an error is better.

Then how do you manage to fix for previously added or removed rules from the policy?

DivyPatel9881: And also we can give support for two functions, like ```AddPolicies()```, in this case, the return type is ```[]error``` and ```AddAtomicPolicies()```, in this case, the return type is ```error```. This can be fruitful for the user.

DivyPatel9881: What do you think, @hsluoyz and @nodece ?

nodece: > > We should only return one error, similar for others.
> > If one add failed, we should roll back previous ones. So it's atomic operation.
> 
> To add atomicity(add or remove everything or nothing) to the operations, it occurs to me, there are two solutions to tackle this :
> 
> 1. We add or remove the policy rule one by one and when we get an invalid policy rule we remove or add the previously added or removed policy rules respectively.
>    OR
> 2. Introduce the function which checks the validity of the policy rules if all the policy rules are valid then we move forward for the operation or we do not and pass the error.
> 
> I think the 2nd solution is better, what do you think?

I agree it.

DivyPatel9881:
@nodece and @hsluoyz, please review it. I have added ```BatchAdapter``` interface, made operations atomic with return type ```error```, ```AddNamedGroupingPolicies()```, ```RemoveNamedGroupingPolicies()```, ```AddGroupingPolicies()```and ```RemoveGroupingPolicies()``` functions in management_api.

persist/file-adapter/adapter.go
DivyPatel9881: Okay

error_test.go
```diff
@@ -16,7 +16,6 @@ package casbin
 
 import (
 	"testing"
```
hsluoyz: Don't remove the blank line.

persist/batch_adapter.go
```diff
@@ -0,0 +1,25 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package persist
+
+// BatchAdapter is the interface for Casbin adapters with multiple add and remove policy functions.
+type BatchAdapter interface {
```
hsluoyz: BatchAdapter should extend Adapter.

util/builtin_operators.go
```diff
@@ -198,7 +198,6 @@ func IPMatch(ip1 string, ip2 string) bool {
 
 		return objIP1.Equal(objIP2)
 	}
-
 	return cidr.Contains(objIP1)
```
hsluoyz: Don't remove new line.

model/policy.go
```diff
@@ -101,6 +101,20 @@ func (model Model) AddPolicy(sec string, ptype string, rule []string) bool {
 	return false
 }
 
+// AddPolicies adds policy rules to the model.
+func (model Model) AddPolicies(sec string, ptype string, rules [][]string) bool {
```
nodece: It's necessary.

model/policy.go
```diff
@@ -113,6 +127,28 @@ func (model Model) RemovePolicy(sec string, ptype string, rule []string) bool {
 	return false
 }
 
+// RemovePolicies removes policy rules from the model.
+func (model Model) RemovePolicies(sec string, ptype string, rules [][]string) bool {
```
nodece: It's unnecessary.

internal_api.go
```diff
@@ -43,6 +45,32 @@ func (e *Enforcer) addPolicy(sec string, ptype string, rule []string) (bool, err
 	return ruleAdded, nil
 }
 
+// addPolicies adds rules to the current policy.
+// removePolicies removes rules from the current policy.
+func (e *Enforcer) addPolicies(sec string, ptype string, rules [][]string) (bool, error) {
+	rulesAdded := e.model.AddPolicies(sec, ptype, rules)
```
nodece: Should call the `AddPolicy`.

internal_api.go
```diff
@@ -68,6 +96,31 @@ func (e *Enforcer) removePolicy(sec string, ptype string, rule []string) (bool,
 	return ruleRemoved, nil
 }
 
+// removePolicies removes rules from the current policy.
+func (e *Enforcer) removePolicies(sec string, ptype string, rules [][]string) (bool, error) {
+	rulesRemoved := e.model.RemovePolicies(sec, ptype, rules)
```
nodece: Should call the `RemovePolicy`.

DivyPatel9881: @nodece, now you are asking me to change the entire structure. @hsluoyz said to me to follow this structure.

DivyPatel9881:
I haven't made those changes in this commit, ita been long back, and @hsluoyz agreed with this. @nodece the structure which you are talking about was the first commit of this PR, please check it and after that, I have changed it, what @hsluoyz said. And the current structure is correct.

DivyPatel9881:
Please don't put my entire work in vain.

DivyPatel9881:
@hsluoyz, I have made the changes.

nodece:
Sorry @DivyPatel9881. Thank you for your contribution. I have some good ideas to help you. 

The `AddPolicies` should call the `e.model.AddPolicy`, save the policy into adapter if return `true`, otherwise iteration next policy (the policy exists Casbin). If happens error, returns the index and error.

nodece:
Other look good to me.

internal_api.go
nodece: It's good.

internal_api.go
nodece: It's good.

DivyPatel9881:
> Sorry @DivyPatel9881. Thank you for your contribution. I have some good ideas to help you.
> 
> If `AdapterEx` is not added.
> 
> The `AddPolicies` should call the `e.model.AddPolicy`, save the policy into adapter if return `true`, otherwise iteration next policy (the policy exists Casbin). If happens error, returns the index and error.
> 
> This is just my opinion.

Thanks for your opinion,  but how would I introduce atomicity in this case, I think @hsluoyz is right, we must add the interface, so as to give new functionality to adapter instead of iterating through the policy rules and calling ```addPolicy()```, as you know it would be pretty much messy and difficult to implement atomicity here.

@hsluoyz what do you think?

I respect your opinion but have some flaws considering the technical feature implementation.

nodece:
Thanks for your point. You are right.

DivyPatel9881:
@hsluoyz can you merge this PR, I have added tests for each functionality and member methods mentioned in #388. The operations are atomic. The addition and removal of each valid policy are successfully done. The coverall suggests a significant increase in coverage. The codeclimate suggests to fix two issues which are inevitable, hence cannot be done anything about it.

error_test.go
```diff
@@ -16,7 +16,7 @@ package casbin
...
-
+
```
hsluoyz: No tab here

error_test.go
```diff
@@ -134,6 +134,19 @@ func TestMockAdapterErrors(t *testing.T) {
 		t.Log(err.Error())
 	}
 
+	rules := [][]string {
+			{"admin", "domain4", "data1", "read"},
+	}
+
+	_, err = e.AddPolicies(rules)
+
```
hsluoyz: no new line

model/policy.go
```diff
@@ -161,4 +197,4 @@ func (model Model) GetValuesForFieldInPolicyAllTypes(sec string, fieldIndex int)
 	util.ArrayRemoveDuplicates(&values)
 
 	return values
-}
+}
```
hsluoyz: Add new line back.

error_test.go
DivyPatel9881: Cant they be ignored.

error_test.go
DivyPatel9881: Cant they be ignored.

util/builtin_operators.go
```diff
@@ -198,7 +198,7 @@ func IPMatch(ip1 string, ip2 string) bool {
 
 		return objIP1.Equal(objIP2)
 	}
-
+
```
hsluoyz: no tab here

error_test.go
hsluoyz: image

They are showing as diff in GitHub, as you see. GitHub doesn't ignore them.

hsluoyz: I don't know what IDE you use, but we recommend Goland, to avoid such errors.

DivyPatel9881:
Can you now merge it?

hsluoyz:
Add new functions to docs

DivyPatel9881:
Okay, thanks @hsluoyz, I will add those functions in documentation.

hsluoyz:
Fix: https://github.com/casbin/casbin/issues/386

PR #391, opened 2020-03-20
fix: return panic error when enforce

jonyhy96:
The panic error needs to return to let the user know what happens.

If i write matchers like something below,I will always get false and nil result,I will never know what should be done to fix it except i debug into it.
``` conf
[matchers]
m = g(r.sub, p.sub, r.dom) && r.dom == p.dom && keyMatch(r.obj, p.obj) && regexMatch(r.act, p.act) \
|| r.sub == "root" \ # for root user
|| (p.sub == "*" && keyMatch4(r.obj, p.obj)) # for anonymous user
```

### DIFF
--- a/enforcer.go
+++ b/enforcer.go
@@ -319,7 +319,7 @@ func (e *Enforcer) EnableLog(enable bool) {
 }
 
 // EnableAutoNotifyWatcher controls whether to save a policy rule automatically notify the Watcher when it is added or removed.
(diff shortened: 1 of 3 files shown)


hsluoyz:
Wrong commits.

jonyhy96:
@hsluoyz perhaps you can tell me which commit is wrong. the last two commits is to fix the master branch's test failed issue.If you don't merge the pr which is failed in ci test then i won't need to do those commits.

PR #395, opened 2020-03-20
fix: return panic error when enforce

jonyhy96:


enforcer.go
```diff
@@ -344,10 +344,10 @@ func (e *Enforcer) BuildRoleLinks() error {
 }
 
 // enforce use a custom matcher to decides whether a "subject" can access a "object" with the operation "action", input parameters are usually: (matcher, sub, obj, act), use model matcher by default when matcher is "".
-func (e *Enforcer) enforce(matcher string, rvals ...interface{}) (bool, error) {
+func (e *Enforcer) enforce(matcher string, rvals ...interface{}) (isOk bool, err error) {
```
hsluoyz: This change is meaningless.

jonyhy96: The what the `fmt.Errorf` in defer use for?

nodece: @hsluoyz You miss it: https://github.com/casbin/casbin/pull/395/files#diff-e26d5a7454fb50c0cf57a7fd078c5d71R350

jonyhy96: @nodece Thank you for point it for me.I think if the maintainer keep close pr without further communication will reduce the contribution motivation.

hsluoyz: You can define `var err error` in the function body. We usually don't use this style: `(isOk bool, err error)`

nodece: When begins panicking, we should `recover` and return the error.

jonyhy96: Can anyone reopen this pr so that we can make further changes to this pr.

jonyhy96: > You can define `var err error` in the function body. We usually don't use this style: `(isOk bool, err error)`

If we use `var err error` at the beginning of the function and set it's value in the recover, we will not be able to return the error unless use named response like `(isOk bool, err error)`.

Because when panic happens the rest of the function(return result, err) won't be execute.
@hsluoyz

hsluoyz: I don't like `(isOk bool, err error)`

jonyhy96: > I don't like `(isOk bool, err error)`

Will you reopen this pr to let me fix this issue? If you do,what is the way you like to response the recovered error to the caller? By panic again?

hsluoyz: I don't want to see return values be defined as `(isOk bool, err error)` literally. Another writing style would be OK. You can make a new PR.

nodece: @hsluoyz, I agree with @jonyhy96 .
We should returns the error, when happened `painc`.

jonyhy96: I think panic again in defer will be a `break change` to current user.
They may haven't handle the panic of enforce.I think we can use named response here to prevent a lot changes in user's code or alert this panic behavior in readme.

If we still want to panic again i think we have three choice to go:
1. just delete this defer 
2. EnforcePanicError = errors.New("wrong matcher in conf");panic(ErrWrongMatcher)
3. panic(fmt.Errorf("wrong matcher: %w",r)) notice only go1.11^ support %w

/cc @hsluoyz @nodece

hsluoyz: OK. I take back my words.

jonyhy96: > OK. I take back my words.

Should i make some changes to this pr? Like:
```go
var ErrorWrongMatcher = errors,.New("wrong matcher in conf")

func (e *Enforcer) enforce(..){
	defer func() {
		if err := recover(); err != nil {			
			err = ErrorWrongMatcher
		}			
	}()
...
}
```
Or just leave it as is.

nodece: @jonyhy96, We just need to change `isOK` to `OK`. The others look good.

WDYH?

jonyhy96: > @jonyhy96, We just need to change `isOK` to `ok`. The others look good.
> 
> WDYH?

I will change the bool response name from isOk to ok.Besides i think the panic info of govaluate is not very clear for user's to know where exactly error is.I think it will be better we use a wrapped error `ErrorWrongMatcher ` or `fmt.Errorf("wrong matcher in conf %w",r)` but %w only support on go 1.11^.Can i use `%w` on this pr?

nodece: built-in function provides return `error`.

jonyhy96: > built-in function provides return `error`.

I don't understand this, the panic happens when `govaluate.NewEvaluableExpressionWithFunctions` is called and passed with wrong `expString` argument which is the matcher of conf.

Please provide more relate info of what do you mean.

nodece: @jonyhy96 Let's returns error to user here are well.

jonyhy96: > @jonyhy96 Let's returns error to user here are well.

I agree with you. What do you think about we use wrapped error to make user more easier to know what really going wrong.

nodece: Sounds good. Can you make a new PR?

nodece: Please use rebase to merge changes into a commit.

nodece:
@hsluoyz lgtm.

enforcer.go
hsluoyz: > but %w only support on go 1.11^.Can i use %w on this pr?

No

Issue #398, opened 2020-03-21
Include change's details when calling Watcher.Update()


urigoori:
Hello,

I’m implementing my own watcher using MySQL and when the enforcer call the watcher.update, is there a way to know what was the change? This call is happening in both addPolicy and removePolicy (and probably in other methods as well) and I need to track the change so other enforcer will get updated

Yang Luo: Why do you need to know the change?

To allow scaling, the casbin code is running on several docker containers. To sync those containers - to update other containers with changes happened in one container and to get updates from other running containers - the application is using table called “casbin_changes”. The table is holding the changes every container applied and the other containers get updates from this table (excluding the changes they initiated) . Those sync actions happen using Watcher with “SetUpdateCallback” and “update”. For this to happen the application need to serialize and de-serialize the change and need to understand what the was that change

hsluoyz:
We should implement a new WatchEx interface to add this functionality.

nodece:
Casbin should add the hooks. somethings like:  
`post-policy`:  The `post-policy` hook runs after the add/remove policy is completed.

hsluoyz:
@nodece good idea. We should put logging, watcher notification, etc. into these hooks. Or even let users define custom hooks.

hsluoyz:
We have WatcherEx to resolve this issue from this PR: https://github.com/casbin/casbin/pull/400

The docs is here: https://casbin.org/docs/en/watchers#watcherex

PR #400, opened 2020-03-22
Add WatcherEx for change's detail

dovics:
I added the WatcherEx interface, which is an enhancement of the watcher.

The user can customize the update function for `AddPolicy()` ,`RemovePolicy() `,`RemoveFilteredPolicy()`,` SavePolicy()`.

For the first three function, their parameters are the same as those passed in by the user，
But for `SavePolicy()`, Parameter is `model.Model`, I think this is a bit complicated for the user. Do you have a better way to replace it?

Or we can delete this function, But I think this function is still very important. It may be used for synchronization between all nodes

### DIFF
--- a/persist/watcher.go
+++ b/persist/watcher.go
@@ -35,10 +35,10 @@ type WatcherEx interface {
 	Watcher
 	// UpdateForAddPolicy calls the update callback of other instances to synchronize their policy.
 	// It is called after Enforcer.AddPolicy()
-	UpdateForAddPolicy(params ...interface{}) error
+	UpdateForAddPolicy(params ...string) error
 	// UPdateForRemovePolicy calls the update callback of other instances to synchronize their policy.
 	// It is called after Enforcer.RemovePolicy()
-	UpdateForRemovePolicy(params ...interface{}) error
+	UpdateForRemovePolicy(params ...string) error
 	// UpdateForRemoveFilteredPolicy calls the update callback of other instances to synchronize their policy.
 	// It is called after Enforcer.RemoveFilteredNamedGroupingPolicy()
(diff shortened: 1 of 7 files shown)


dovics:
fix #398 

The codeclimate suggests to fix two issues which are inevitable

persist/watcher.go
```diff
@@ -27,3 +29,20 @@ type Watcher interface {
 	// Close stops and releases the watcher, the callback function will not be called any more.
 	Close()
 }
+
+// WatcherEx is the strengthen for Casbin watchers.
```
hsluoyz: Start a new file: watch_ex.go

watcher_test.go
```diff
@@ -39,3 +43,34 @@ func TestSetWatcher(t *testing.T) {
 
 	e.SavePolicy() //calls watcher.Update()
 }
+
```
hsluoyz: put into watcher_ex_test.go

hsluoyz:
@nodece @urigoori please review.

nodece:
lgtm

GopherJ:
I'm wondering why don't we inherit event emitter for enforcers? It solves this type of question easily.
We just need to emit with policies' changes and call `update` with these changes.


https://github.com/casbin/casbin-rs/blob/1d3c04aaa611d7403aa6a08368c848b43c8bf7b7/src/emitter.rs#L14

hsluoyz:
@GopherJ I like this idea. An extra mechanism like WatchEx should be decoupled with Casbin core code as much as possible.

nodece:
@hsluoyz Is there any progress?

PR #401, opened 2020-03-23
Add exforceEx() for the hit rule

dovics:
Add `ExforceEx()` for the hit rule，whether it hits depends on user selected effect. User can modify it by interface `EffectorEx`

`exforceEx()` is basically a copy of `exforce()`,I know there is a lot of duplicate code here, but it's hard to reuse them without modifying `enforce()`


hsluoyz:
Enforce instead of exforce.

also code dup is not acceptable.

PR #403, opened 2020-03-24
Use fmt.Errorf() replace errors.New(fmt.Sprintf())

dovics:
golint warning:
 > should replace errors.New(fmt.Sprintf(...)) with fmt.Errorf(...)

I think the new is more concise than the old


dovics:
codeclimate's issus is inevitable

hsluoyz:
Did you replace all?

dovics:
Yes, this usage appears in only one file

hsluoyz:
lgtm

PR #409, opened 2020-03-29
Missing member functions added to the interface.

DivyPatel9881:
There were certain member function signatures missing in the enforcer interface. I have added those functions.

### DIFF
--- a/enforcer_interface.go
+++ b/enforcer_interface.go
@@ -30,12 +30,14 @@ type IEnforcer interface {
 	InitWithFile(modelPath string, policyPath string) error
 	InitWithAdapter(modelPath string, adapter persist.Adapter) error
 	InitWithModelAndAdapter(m model.Model, adapter persist.Adapter) error
+	initialize()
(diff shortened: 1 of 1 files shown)


enforcer_interface.go
```diff
@@ -30,12 +30,14 @@ type IEnforcer interface {
 	InitWithFile(modelPath string, policyPath string) error
 	InitWithAdapter(modelPath string, adapter persist.Adapter) error
 	InitWithModelAndAdapter(m model.Model, adapter persist.Adapter) error
+	initialize()
```
nodece: We should not add the private method to the `interface`. 

WDYT?

DivyPatel9881: Actually it doesn't matter if we add it or not, but if we add it, ```initialize()``` becomes a compulsory implemented member function, as it is important in any of the ```Enforcer``` variants.

nodece:
@DivyPatel9881 lgtm.

Issue #412, opened 2020-03-31
problem of running ABAC


yangshiduo:
this is my code
```go
package main
import (
        "fmt"
        "github.com/casbin/casbin"
        "github.com/casbin/casbin/model"
)
const modelText =`
[request_definition]
r=sub,obj,act
[policy_definition]
p=sub,obj,act
[policy_effect]
e=some(where(p.eft==allow))
[matchers]
m=r.sub.Type==r.obj.Type && r.sub.Method==r.obj.Method && r.sub.Name==r.obj.Name && r.sub.Read
==r.act.Read && r.sub.Write==r.act.Write
`
type User struct{
   UserName string
   Group []Group
}
type Group struct {
   Name string
   Read bool
   Write bool
   Type string
   Method string
}
type Obj struct {
   Name string
   Type string
   Method string
}
type Act struct {
   Read bool
   Write bool
}
func main(){
   m :=model.Model{}
   m.LoadModelFromText(modelText)
   e:=casbin.NewEnforcer(m)
   group1 :=Group{
        Name: "G1",
        Read: true,
        Write:true,
        Type:"user",
        Method:"get",
   }
   group2 :=Group{
        Name: "G2",
        Read: true,
        Write:true,
        Type:"user",
        Method:"get",
  }
  user1 :=User{
        UserName: "ysd",
        Group:  []Group{group1,group2},
  }
  obj :=Obj{
        Name: "G1",
        Type: "user",
        Method: "get",
  }
  act:=Act{
        Read:true,
        Write:true,
  }
  for _, v :=range user1.Group{
        flag:= e.Enforce(v,obj,act)
        if flag {
                   fmt.Println("allow")
        } else {
                   fmt.Println("not allow")
        }
  }
}
```
when i go run ABAC.go
![捕获1](https://user-images.githubusercontent.com/49757864/78108183-742feb00-7429-11ea-9c76-0f8dc140b151.PNG)

go version go1.12.5 linux/amd64

jonyhy96:
Separate your conf with `, ` due to https://github.com/casbin/casbin/blob/10eaf10a1c/model/model.go#L60 .

The code bellow will work properly.
```go
const modelText = `
[request_definition]
r = sub, obj, act
[policy_definition]
p = sub, obj, act
[policy_effect]
e = some(where (p.eft == allow))
[matchers]
[...]

PR #413, opened 2020-04-02
Added support for the CONF files having just comma (no spaces) separated values.

DivyPatel9881:
As saw in issue #412, the users are facing problems parsing ```CONF``` files. So, why don't we extend our support to just comma(no space) separated values ```,``` , not just ```, ```.

In this way, users will be able to write CONF files in both the above ways.

### DIFF
--- a/model/model.go
+++ b/model/model.go
@@ -57,9 +57,9 @@ func (model Model) AddDef(sec string, key string, value string) bool {
 	ast.Value = value
 
 	if sec == "r" || sec == "p" {
-		ast.Tokens = strings.Split(ast.Value, ", ")
+		ast.Tokens = strings.Split(ast.Value, ",")
 		for i := range ast.Tokens {
-			ast.Tokens[i] = key + "_" + ast.Tokens[i]
+			ast.Tokens[i] = key + "_" + strings.TrimSpace(ast.Tokens[i])
 		}
 	} else {
 		ast.Value = util.RemoveComments(util.EscapeAssertion(ast.Value))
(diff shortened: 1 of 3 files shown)


model/model.go
```diff
@@ -57,7 +57,8 @@ func (model Model) AddDef(sec string, key string, value string) bool {
 	ast.Value = value
 
 	if sec == "r" || sec == "p" {
-		ast.Tokens = strings.Split(ast.Value, ", ")
+		newValue := strings.Replace(ast.Value, " ", "", -1)
+		ast.Tokens = strings.Split(newValue, ",")
 		for i := range ast.Tokens {
 			ast.Tokens[i] = key + "_" + ast.Tokens[i]
```
nodece, line 63: Better to use `strings.Replace` here. 
WDYT?

jonyhy96: I think it will be more clear if we use 
```go
	act.Tokens = strings.FieldsFunc(ast.Value, func(c rune) bool {
		return !unicode.IsLetter(c) && !unicode.IsNumber(c)
	})
```
So that users can use whatever sep string they like and that won't affect the further logic.

DivyPatel9881: @jonyhy96, thanks for your suggestion, but, I think, we shouldn't allow users to change the actual syntax and semantic any further, I think this much liberty is way enough for users.

jonyhy96:
> Okay, I have made the changes, as you said @nodece.

This change will make current user's code can't work.
Perhaps he means:
```go
ast.Tokens = strings.Split(strings.Replace(ast.Value, " ", "", -1), ",")
```

model/model.go
nodece: @jonyhy96 The commits need rebase for keep clean.

nodece: Sorry @jonyhy96 , I should be at @DivyPatel9881 .

hsluoyz:
lgtm

Issue #355, opened 2020-01-28
Explain enforcement by informing matched rules


jruizaranguren:
In some scenarios it would be useful to have a way to explain why an enforcement was or wasn't satisfied.
The simplest idea could be to just inform of the matched rules that allowed satisfying (or not) the model.

Is there any way to accomplish this in casbin?

hsluoyz:
Can you make a PR to show how it works? And don't introduce too much performance downgrade after this change I hope.

hsluoyz:
Can anyone work on it?

DivyPatel9881:
I would like to work on this.

DivyPatel9881:
Tomorrow, I will be posting a working theoretical solution to this

dovics:
I have been working on this issus for a while, Now i have completed my first version #401 @jruizaranguren Can you see if my implementation does meet your expectations？

hsluoyz:
Any further progress on this?

nodece:
I propose my implementation:

In my opinion, I just want to know which policy decides whether to return `true` or `false`.

The following is signature of `ExplainEnforce`:

```golang
type ExplainState struct {
	Ok        bool     // enforce result
	HitPolicy []string // which policy decides the enforce result
}

func (e *Enforcer) ExplainEnforce(rvals ...interface{}) (explain *ExplainState, err error)
```

How to get the hit policy?

We need the `Effector.MergeEffects` returns the hit index then pass it to `model` get the policy. I found that `results` is unused in  `Effector.MergeEffects` so we can use it. Let's define the `len(results)` equal to `len(e.model["p"]["p"].Policy) + 1` then`results[len(results)-1]` as hit index.

What do you think?

dovics:
Why deal with the hit policy in `MergeEffects`? this function will also be called by normal `Enforce()`

In addition, the HitPolicy may be more than one, for example:
```
g, alice, admin
p, admin, data1, read
p, alice, data1, read
```

nodece:
Maybe the following can help to understand.
https://github.com/casbin/casbin/blob/master/effect/default_effector.go#L35
https://github.com/casbin/casbin/blob/master/effect/default_effector.go#L43

It just check `effect` is `Allow` or `Deny` then returns result.

nodece:
Don't worry. `Enforce()` unused the `results`.

hsluoyz:
@jruizaranguren solved by: https://github.com/casbin/casbin/pull/425

[...]

PR #425, opened 2020-04-16
Explain enforcement in log

dovics:
For #355
Explain why an enforcement was or wasn't satisfied.Controlled by `autoExplain` in enforce,Print the policy related to the result to the log.

such as `some(where (p.eft == allow))`, the explanation will be all allowed policy.

whether it hits depends on user selected effect. User can modify it by interface EffectorEx

### DIFF
--- a/effect/default_effector.go
+++ b/effect/default_effector.go
@@ -27,49 +27,54 @@ func NewDefaultEffector() *DefaultEffector {
 }
 
 // MergeEffects merges all matching results collected by the enforcer into a single decision.
-func (e *DefaultEffector) MergeEffects(expr string, effects []Effect, results []float64) (bool, error) {
+func (e *DefaultEffector) MergeEffects(expr string, effects []Effect, results []float64) (bool, int, error) {
 	result := false
+	explainIndex := -1
 	if expr == "some(where (p_eft == allow))" {
 		result = false
-		for _, eft := range effects {
+		for i, eft := range effects {
 			if eft == Allow {
 				result = true
+				explainIndex = i
 				break
 			}
 		}
 	} else if expr == "!some(where (p_eft == deny))" {
 		result = true
-		for _, eft := range effects {
+		for i, eft := range effects {
 			if eft == Deny {
 				result = false
+				explainIndex = i
 				break
 			}
 		}
 	} else if expr == "some(where (p_eft == allow)) && !some(where (p_eft == deny))" {
 		result = false
-		for _, eft := range effects {
+		for i, eft := range effects {
 			if eft == Allow {
 				result = true
 			} else if eft == Deny {
 				result = false
+				explainIndex = i
 				break
 			}
 		}
 	} else if expr == "priority(p_eft) || deny" {
 		result = false
-		for _, eft := range effects {
+		for i, eft := range effects {
 			if eft != Indeterminate {
 				if eft == Allow {
 					result = true
 				} else {
 					result = false
 				}
+				explainIndex = i
 				break
 			}
 		}
 	} else {
-		return false, errors.New("unsupported effect")
+		return false, -1, errors.New("unsupported effect")
 	}
 
-	return result, nil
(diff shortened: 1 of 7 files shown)


hsluoyz:
I think logs are not very friendly for users. They want to return matched rules:

```
In some scenarios it would be useful to have a way to explain why an enforcement was or wasn't satisfied.
The simplest idea could be to just inform of the matched rules that allowed satisfying (or not) the model.
```

Can we add a function called `EnforceEx()`, which is similar to `Enforce()`, but returns matched rule(s) additionally?

dovics:
the `enforce()` is too long, what do you think of changing the signature of `enforce()`
```
func (e *Enforcer) enforce(matcher string, rvals ...interface{}) (ok bool,explains [][]string,  err error)
```
or
```
func (e *Enforcer) enforce(matcher string, explains *[][]string, rvals ...interface{}) (ok bool, err error)
```

otherwise we need to split `enforce()`, I think it is difficult to do

dovics:
I tried adding `EnforceEx()` and `EnforceExWithMatch()`.

I think it is better to return all the policies related to the result. And the result and hitPolicies are returned directly in the return value

effect/default_effector.go
```diff
@@ -73,3 +73,24 @@ func (e *DefaultEffector) MergeEffects(expr string, effects []Effect, results []
 
 	return result, nil
 }
+
+// IsHit return whether it should be explain
+func (e *DefaultEffector) IsHit(expr string, eft Effect) (bool, error) {
```
nodece, line 78: Why not use `MergeEffects`? Add `IsHit` is breaking changes.

nodece:
Thanks @dovics , Must approve.

effect/default_effector.go
dovics: `results` is used to store matcher results in the case of float64.

If users use special matcher and their own designed effects, It may be used. I know this situation will be rare, but if we add feature to it, will this mislead the user？

nodece: Unused the `results` in Casbin. So we don't need to care it.

Use `MergeEffects`：
```golang
e.eft.MergeEffects(e.model["e"]["e"].Value, policyEffects, nil)
```

dovics: Thanks for your guidance, use `MergeEffects()` in the new commit

nodece: Latest implementation from https://github.com/casbin/casbin/issues/355#issuecomment-615056049.  

For your previous implementation, you just need to change `isHit, err := e.eft.MergeEffects(e.model["e"]["e"].Value, []effect.Effect{policyEffects[i]}, nil)`

nodece: Now, look good to me.

effect/default_effector.go
```diff
@@ -29,41 +29,46 @@ func NewDefaultEffector() *DefaultEffector {
...
 				break
 			}
 		}
 	} else if expr == "!some(where (p_eft == deny))" {
 		result = true
-		for _, eft := range effects {
+		for i, eft := range effects {
 			if eft == Deny {
 				result = false
+				results[len(results)-1] = float64(i)
```
hsluoyz, line 47: Don't mess with the existing `results`. If you want to pass a variable, just add a new parameter

dovics: Does adding a new parameter affect other implementations？

 what do you think of using `EffrctEx` ? We can add a function to determine what needs to be returned, I implemented it in the last commit

nodece: `results` is unused.

dovics: I think the point is don't add usage to `results`, which will make the code difficult to understand and less readable

dovics: Maybe I think too much, I will listen to @hsluoyz

nodece: Agreed @hsluoyz

nodece:
@hsluoyz Is there any progress?

PR #448, opened 2020-05-06
fix: improve KeyMatch method to avoid endless loop

### LINKED ISSUE #447: I think we should improve KeyMatch method to avoid endless loop
In KeyMatch all use `for{}` loop
```go
	re := regexp.MustCompile(`(.*)\{[^/]+\}(.*)`)
	for {
		if !strings.Contains(key2, "/{") {
			break
		}

		key2 = re.ReplaceAllString(key2, "$1([^/]+)$2")
	}
```
But in this case, it will form an endless loop
```
testKeyMatch2(t, "/alice", "/:/id", true)
testKeyMatch3(t, "/myid/using/myresid", "/{id/using/{resId}", true)
testKeyMatch4(t, "/parent/123/child/123", "/parent/{i/d}/child/{i/d}", true)
```

techoner:
fix: #447, improve KeyMatch method to avoid endless loop.

### DIFF
--- a/util/builtin_operators.go
+++ b/util/builtin_operators.go
@@ -73,14 +73,8 @@ func KeyMatchFunc(args ...interface{}) (interface{}, error) {
 func KeyMatch2(key1 string, key2 string) bool {
 	key2 = strings.Replace(key2, "/*", "/.*", -1)
 
-	re := regexp.MustCompile(`(.*):[^/]+(.*)`)
-	for {
-		if !strings.Contains(key2, "/:") {
-			break
-		}
-
-		key2 = re.ReplaceAllString(key2, "$1[^/]+$2")
-	}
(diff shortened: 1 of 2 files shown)


nodece:
Hi @techoner, We need follow semantic commit. 

Rewrite the commit like: `fix: improve KeyMatch method to avoid endless loop`.

hsluoyz:
lgtm

nodece:
Hi @hsluoyz @techoner,  semantic commit refers to git commit rather than PR title.

nodece:
The RP cannot tigger a release.

hsluoyz:
@nodece sorry. I have reverted the PR.

@techoner can you re-make a new PR?

techoner:
OK.

PR #454, opened 2020-05-10
feat: add AddRolesForUser

gadelkareem:
add AddRolesForUser

### DIFF
--- a/rbac_api.go
+++ b/rbac_api.go
@@ -55,6 +55,22 @@ func (e *Enforcer) AddRoleForUser(user string, role string) (bool, error) {
 	return e.AddGroupingPolicy(user, role)
 }
 
+// AddRolesForUser adds roles for a user.
+// Returns false if the user already has the roles (aka not affected).
+func (e *Enforcer) AddRolesForUser(user string, roles []string) (bool, error) {
+    f := false
+    for _, r := range roles {
+        b, err := e.AddGroupingPolicy(user, r)
+        if err != nil {
+            return false, err
+        }
+        if b {
+            f = true
+        }
+    }
+    return f, nil
+}
+
 // DeleteRoleForUser deletes a role for a user.
 // Returns false if the user does not have the role (aka not affected).
 func (e *Enforcer) DeleteRoleForUser(user string, role string) (bool, error) {
--- a/rbac_api_test.go
+++ b/rbac_api_test.go
@@ -126,6 +126,16 @@ func TestRoleAPI(t *testing.T) {
 	testEnforce(t, e, "bob", "data2", "write", true)
 }
 
+func TestEnforcer_AddRolesForUser(t *testing.T) {
+	e, _ := NewEnforcer("examples/rbac_model.conf", "examples/rbac_policy.csv")
+
+	e.AddRolesForUser("alice", []string{"data1_admin", "data2_admin", "data3_admin"})
+	testGetRoles(t, e, "alice", []string{"data1_admin", "data2_admin", "data3_admin"})
+	testEnforce(t, e, "alice", "data1", "read", true)
+	testEnforce(t, e, "alice", "data2", "read", true)
+	testEnforce(t, e, "alice", "data2", "write", true)
+}
+
 func testGetPermissions(t *testing.T, e *Enforcer, name string, res [][]string) {
 	t.Helper()
 	myRes := e.GetPermissionsForUser(name)


rbac_api.go
```diff
@@ -55,6 +55,22 @@ func (e *Enforcer) AddRoleForUser(user string, role string) (bool, error) {
 	return e.AddGroupingPolicy(user, role)
 }
 
+// AddRolesForUser adds roles for a user.
+// Returns false if the user already has the roles (aka not affected).
+func (e *Enforcer) AddRolesForUser(user string, roles []string) (bool, error) {
+    f := false
+    for _, r := range roles {
```
nodece, line 62: Calling `AddGroupingPolicies` is better.

nodece:
@gadelkareem Please add test cases and fix CI error.

rbac_api.go
gadelkareem: How does `AddGroupingPolicies` work? what is the structure of `rules` in this case?

nodece: See `management_api_test.go` for `AddGroupingPolicies` test cases.

gadelkareem: There might be a bug in `AddGroupingPolicies` because using the code below makes the user only has one role so it looks like the policies are being overwritten somehow.

```
func (e *Enforcer) AddRolesForUser(user string, roles []string) (bool, error) {
	var rules [][]string
	for _, r := range roles {
		rules = append(rules, []string{user, r})
	}
	return e.AddGroupingPolicies(rules)
}
```

nodece: Casbin supports a user to inherit multiple roles.

hsluoyz:
@gadelkareem fix Semantic Pull Request

rbac_api.go
gadelkareem: Though `AddGroupingPolicies` does not allow adding multiple roles to the same user.

nodece: > Though `AddGroupingPolicies` does not allow adding multiple roles to the same user.

Could you point out where it is?

gadelkareem: if you use the above code on my branch you would get an error:
```
=== RUN   TestEnforcer_AddRolesForUser
2020/05/10 16:10:20 Role links for: g
2020/05/10 16:10:20 alice < data2_admin
    TestEnforcer_AddRolesForUser: rbac_api_test.go:133: Roles for  alice :  [data2_admin]
    TestEnforcer_AddRolesForUser: rbac_api_test.go:133: Roles for  alice :  [data2_admin] , supposed to be  [data1_admin data2_admin data3_admin]
```

only one role ended up added to the user which is `data2_admin`

gadelkareem:
@hsluoyz title and commit added. Not sure why it is not fixed yet.

hsluoyz:
Replace your previous commits.

rbac_api.go
nodece: You ignore the returns of `e.AddRolesForUser`. `e.AddRolesForUser("alice", []string{"data1_admin", "data2_admin", "data3_admin"})` returns `false, nil`, so you get `[data2_admin]`.

`AddGroupingPolicies ` has  atomic operation. Due to `alice` has `data2_admin` role, so cannot add `data2_admin` to `alice` and returns `false, nil`.

gadelkareem: Then using `AddGroupingPolicy` is a better  option because we need to make those the roles are added even though one of them was already added before.

nodece: > Then using `AddGroupingPolicy` is a better option because we need to make those the roles are added even though one of them was already added before.

Sounds good. 

What do you think?  @hsluoyz @GopherJ @DivyPatel9881

DivyPatel9881: > What do you think? @hsluoyz @GopherJ @DivyPatel9881

I agree.

GopherJ:
I'm not sure of users need to know one roles exist already.  If yes then probably batch operation is better. 

Also one disadvantage on this solution is: RoleLinks will be rebuilt n times instead of just one time. 

Final point, this will cause inconsistence with the design of other batch operations.

nodece:
@gadelkareem Using `AddPolicies` or `RemovePolicies` is better in batch operation. 

> Then using `AddGroupingPolicy` is a better option because we need to make sure those roles are added even though one of them was already added before.

This is good idea. @GopherJ 

We need to improve `AddPolicies` and `RemovePolicies`.

GopherJ:
@nodece  In my opinion, it may not be so easy to improve, because db adapters need to tell us how many have actually been inserted/removed.

I also would like to achieve this but the tricky part is somehow we need to know how many have exactly been inserted/removed. 

The easy way of doing this is:

Don't allow dup

The harder way of doing this is:

Db adapters should tell us how many have been inserted/removed

nodece:
@GopherJ We can extract what data should be add/remove via `has_policy` then pass the data to adapter and model.

GopherJ:
@nodece I have doubt that we can rely on `has_policy` because what happens to db adapters can be a black box

PR #463, opened 2020-05-14
fix(synced enforcer): use casbin logger

ninedraft:
Use "github.com/casbin/casbin/v2/log" instead of "log" package.
It will increase usability -- I need to use the zap logger, but synced logger prints to default logger at the same time as the std log, which can cause a race condition.

### DIFF
--- a/enforcer_synced.go
+++ b/enforcer_synced.go
@@ -15,12 +15,12 @@
 package casbin
 
 import (
-	"log"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


nodece:
@ninedraft Please fix the CI error.

ninedraft:
@nodece done

hsluoyz:
@nodece please review.

nodece:
LGTM

Issue #453, opened 2020-05-10
There was a problem in learning


ruyi84:
I encountered a problem in the process of learning casbin, I wrote a strategy, there was an error not planned:
csv:
```csv
p, zeta, data1, read, allow
p, bob, data2, write, allow
p, zeta, data2, write, allow
p, zeta, data2, write, reject
```
conf

```ini
e = some (where (p.eft == allowed)) &&! some (where (p.eft == rejected))

m = r.sub == p.sub && r.obj == p.obj && r.act == p.act
```

However, zeta / data2 / write = true, this is why

GopherJ:
@ruyi84 You should use `deny` instead, `reject` is not casbin keyword

beyondexp:
I also encountered the same problem，but I did't use "rejected". At the beginning, I found an article introducing casbin from this address. When I tried to run it, I got the same error as above. I carefully checked for a long time, and finally found that it was a problem caused by code comments. Just delete the code comment after "p = sub, obj, act, eft" to solve this problem.
**There is still a problem.Why only the comment here will cause this problem？**
```
# Request definition
[request_definition]
r = sub, obj, act

# Policy definition
[policy_definition]
p = sub, obj, act, eft # 这里我们定义了eft，不使用默认值

# Policy effect
[policy_effect]
e = some(where (p.eft == allow)) && !some(where (p.eft == deny)) # 这里使用了deny优先

# Matchers
[matchers]
m = r.sub == p.sub && r.obj == p.obj && r.act == p.act #最简单的匹配规则。请求的参数与策略参数一致时获得策略结果
```

nodece:
Thank you @beyondexp for your report. This is a good feature, but we don't support comments after expression. 

I provide a correct model for you.
```conf
# Request definition
[request_definition]
r = sub, obj, act

# Policy definition 这里我们定义了eft，不使用默认值
[policy_definition]
p = sub, obj, act, eft 

# Policy effect 这里使用了deny优先
[policy_effect]
e = some(where (p.eft == allow)) && !some(where (p.eft == deny))

# Matchers 最简单的匹配规则。请求的参数与策略参数一致时获得策略结果
[matchers]
m = r.sub == p.sub && r.obj == p.obj && r.act == p.act 
```

GopherJ:
@nodece  it seems that we have this supported using this line no?: https://github.com/casbin/casbin/blob/0aab1e445eb0e74491982b53c2d63c57880812c4/model/model.go#L65

[...]

PR #464, opened 2020-05-19
fix: support comment after effect and matcher

dovics:
Mentioned #453


nodece:
@dovics The commit type should be `fix`.

dovics:
> @dovics The commit type should be `fix`.

Ok, I have changed

nodece:
LGTM

Issue #439, opened 2020-04-25
merge_effect quick return?


GopherJ:
Currently if we are using effect like:

```
some(where (p_eft == allow))
```

or

```
!some(where (p_eft == deny))
```

or

```
some(where (p_eft == allow)) && !some(where (p_eft == deny))
```

there isn't any optimization (quick return) it always need to iterate the whole list of policy which is not ideal because sometimes we just need to find one match and quick return the enforcing result.

hsluoyz:
I think this idea is good.

GopherJ:
implemented in: https://github.com/casbin/casbin-rs/pull/142

DivyPatel9881:
Okay, then I will implement it for casbin-core

GopherJ:
hi @DivyPatel9881 I just found that channel's overhead is out of my imagination so I reverted the PR and made https://github.com/casbin/casbin-rs/pull/145

DivyPatel9881:
Okay, I will check this out.

GopherJ:
@DivyPatel9881 I would suggest a new interface similar to this: https://github.com/casbin/casbin-rs/blob/master/src/effector.rs

nodece:
@DivyPatel9881 Is there any progress?

DivyPatel9881:
Yes, I am on it. I have added ```EffectorStream``` interface and its corresponding default implementation as ```DefaultEffectorStream``` struct. Now just some changes have to be made in the enforcer implementation. After completion of changes in Enforcer implementation, I will be posting a PR.

nodece:
@DivyPatel9881 Thanks for your contribution. It is greatly improve performance.

nodece:
@DivyPatel9881 How is your progress?

DivyPatel9881:
Almost done. Today I will be posting a PR regarding this.

GopherJ:
I think this can be closed

PR #472, opened 2020-05-27
fix: Quick return.

DivyPatel9881:
fix: #439

### DIFF
--- a/effect/default_effector.go
+++ b/effect/default_effector.go
@@ -24,7 +24,7 @@ func NewDefaultEffector() *DefaultEffector {
 	return &e
 }
 
-func (e *DefaultEffector) NewStream(expr string, cap int) (DefaultEffectorStream) {
+func (e *DefaultEffector) NewStream(expr string, cap int) (EffectorStream) {
 	if !(cap>0) {
 		panic("cap should be greater than 0")
 	}
@@ -38,7 +38,7 @@ func (e *DefaultEffector) NewStream(expr string, cap int) (DefaultEffectorStream
 		panic("unsupported effect: " + expr)
 	}
 
-	des := DefaultEffectorStream{false, res, expr, 0, cap, make([]int, 0)}
+	des := &DefaultEffectorStream{false, res, expr, 0, cap, make([]int, 0)}
 
 	return des
 }
--- a/effect/default_effector.go
+++ b/effect/default_effector.go
@@ -14,8 +14,6 @@
 
 package effect
 
-import "errors"
-
 // DefaultEffector is default effector for Casbin.
 type DefaultEffector struct {
 }
@@ -26,55 +24,27 @@ func NewDefaultEffector() *DefaultEffector {
 	return &e
 }
 
-// MergeEffects merges all matching results collected by the enforcer into a single decision.
-func (e *DefaultEffector) MergeEffects(expr string, effects []Effect, results []float64) (bool, int, error) {
-	result := false
-	explainIndex := -1
-	if expr == "some(where (p_eft == allow))" {
-		result = false
-		for i, eft := range effects {
-			if eft == Allow {
-				result = true
-				explainIndex = i
-				break
-			}
-		}
-	} else if expr == "!some(where (p_eft == deny))" {
-		result = true
-		for i, eft := range effects {
-			if eft == Deny {
-				result = false
-				explainIndex = i
-				break
-			}
-		}
-	} else if expr == "some(where (p_eft == allow)) && !some(where (p_eft == deny))" {
-		result = false
-		for i, eft := range effects {
-			if eft == Allow {
-				result = true
-			} else if eft == Deny {
-				result = false
-				explainIndex = i
-				break
-			}
-		}
-	} else if expr == "priority(p_eft) || deny" {
-		result = false
-		for i, eft := range effects {
-			if eft != Indeterminate {
-				if eft == Allow {
-					result = true
-				} else {
-					result = false
-				}
-				explainIndex = i
-				break
-			}
-		}
-	} else {
-		return false, -1, errors.New("unsupported effect")
+func (e * DefaultEffector) NewStream(expr string, cap uint) (EffectorStream) {
+	if !(cap>0) {
+		panic("cap should be greater than 0")
 	}
 
-	return result, explainIndex, nil
+	var res bool
+	if expr=="some(where (p_eft == allow))" || expr=="some(where (p_eft == allow)) && !some(where (p_eft == deny))" || expr=="priority(p_eft) || deny" {
+		res = false
+	} else if expr=="!some(where (p_eft == deny))" {
+		res = true
+	} else {
+		panic("unsupported effect: " + expr)
+	}
+
+	var des DefaultEffectorStream
+	des.done = false
+	des.res = res
+	des.expr = expr
+	des.cap = cap
+	des.idx = 0
+	des.expl = make([]uint, 0)
+
+	return des
 }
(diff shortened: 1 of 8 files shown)


hsluoyz:

effect/default_effector.go
```diff
@@ -26,55 +24,27 @@ func NewDefaultEffector() *DefaultEffector {
...
+	} else if expr=="!some(where (p_eft == deny))" {
+		res = true
 	} else {
-		return false, -1, errors.New("unsupported effect")
+		panic("unsupported effect: " + expr)
 	}
 
-	return result, explainIndex, nil
-}
+	des := DefaultEffectorStream{}
```
GopherJ, line 41: can we initialize it directly so that the syntax looks better. I mean:

```
DefaultEffectorStream {
...
}
```

effect/default_effector_stream.go
```diff
@@ -0,0 +1,82 @@
+package effect
+
+// DefaultEffectorStream is the default implementation of interface EffectorStream.
+type DefaultEffectorStream struct {
+	done bool
+	res bool
+	expr string
+	idx int
+	cap int
+	expl int
```
GopherJ, line 10: expl should be a slice because we can have multiple hitted policy for effect: `some(where (p_eft == allow)) && !some(where (p_eft == deny))` etc

GopherJ:
Otherwise it's fine for me. Ping @nodece to have a look.

Also please fix: https://codeclimate.com/github/casbin/casbin/pull/472

UnderTreeTech:
@DivyPatel9881 @GopherJ I suggest link a new benchmark report to compare with performance of preview versions.

DivyPatel9881:
I have fixed most of the issues pointed out by ```codeclimate``` just one issue to be fixed, which literally beyond fixing. I think it's ready to go. @nodece waiting for your review.

DivyPatel9881:
I have fixed most of the issues pointed out by ```codeclimate``` just one issue left to be fixed, which is literally beyond fixing. I think it's ready to go. @nodece  waiting for your review.

DivyPatel9881:
> @DivyPatel9881 @GopherJ I suggest link a new benchmark report to compare with the performance of preview versions.

Okay, nice suggestion @UnderTreeTech can you guide me on how should I benchmark the code, haven't really done before. Your inputs are totally welcome.

effect/effector_stream.go
```diff
@@ -0,0 +1,8 @@
+package effect
+
+// EffectorStream is the interface for effector stream.
+type EffectorStream interface {
+	Next() bool
```
nodece, line 5: Using `Current() bool` for keep same design as `casbin-rs`.

go.sum
```diff
@@ -1,2 +1,3 @@
 github.com/Knetic/govaluate v3.0.1-0.20171022003610-9aa49832a739+incompatible h1:1G1pk05UrOh0NlF1oeaaix1x8XzrfjIDK47TY0Zehcw=
 github.com/Knetic/govaluate v3.0.1-0.20171022003610-9aa49832a739+incompatible/go.mod h1:r7JcOSlj0wfOMncg0iLm8Leh48TZaKVeNIfJntJ2wa0=
+github.com/casbin/casbin v1.9.1 h1:ucjbS5zTrmSLtH4XogqOG920Poe6QatdXtz1FEbApeM=
```
nodece, line 3: Why add the casbin v1.9.1?

effect/default_effector.go
```diff
@@ -26,55 +24,21 @@ func NewDefaultEffector() *DefaultEffector {
...
-				if eft == Allow {
-					result = true
-				} else {
-					result = false
-				}
-				explainIndex = i
-				break
-			}
-		}
+func (e *DefaultEffector) NewStream(expr string, cap int) (DefaultEffectorStream) {
```
nodece, line 27: It must returns `EffectorStream` interface.  @hsluoyz Should we avoid using `painc` ?

effect/effector.go
```diff
@@ -26,6 +26,5 @@ const (
 
 // Effector is the interface for Casbin effectors.
 type Effector interface {
-	// MergeEffects merges all matching results collected by the enforcer into a single decision.
-	MergeEffects(expr string, effects []Effect, results []float64) (bool, int, error)
+	NewStream(expr string, cap int) (DefaultEffectorStream)
```
nodece, line 29: Must `EffectorStream` interface.

enforcer.go
```diff
@@ -460,44 +466,33 @@ func (e *Enforcer) enforce(matcher string, explains *[]string, rvals ...interfac
...
-					policyEffects[i] = effect.Indeterminate
+					eft = effect.Indeterminate
 				}
 			} else {
-				policyEffects[i] = effect.Allow
+				eft = effect.Allow
 			}
 
-			if e.model["e"]["e"].Value == "priority(p_eft) || deny" {
+			if eftStream.PushEffect(eft) {
```
nodece, line 491: A good suggestion is to record the current policy  to `explains ` when PushEffect returns `res == true && done == true`.

DivyPatel9881: I think it won't affect the performance anyways.

effect/default_effector_stream.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+func (s *DefaultEffectorStream) Next() bool {
+	if !s.done {
+		panic("done should be true")
+	}
+	return s.res
+}
+
+func (s *DefaultEffectorStream) Explain() []int {
```
nodece, line 26: Sorry, I missed the Explain. There should be only one policy that affects enforce result.

What do you think @DivyPatel9881 ?

When `res == true && done ==true` set the hit policy index to `expl`. `Explain` returns the index then.

GopherJ: @nodece  `some(where (p_eft == allow)) && !some(where (p_eft == deny))` can have multiple policies hitted

hsluoyz:
@nodece @DivyPatel9881 this PR causes breaking change. And based on semver, we have to release v3. But releasing v3 means to update all Casbin source code and even our users' import paths from v2 to v3, which is a disaster and unacceptable. See some discussions here: https://github.com/casbin/casbin/pull/478#issuecomment-637897434

Do you have a good way to avoid the import path update disaster? If not, I have to revert this PR.

hsluoyz:
This PR is reverted for now.

DivyPatel9881:
I think of no solution to this, it's like if you want to work with upgraded version update your workflow or software. And I think its kind of good, if they really want those feature which was added in v3, then only update, otherwise work with v2, as it will overburden your software with unwanted stuff. These problems are common when a specific repository is a kind of under breaking changes or under-development.

DivyPatel9881:
This won't be a problem when casbin-core has attained a certain level.

nodece:
Rename `github.com/casbin/casbin/v2/path` to `github.com/casbin/casbin/v2/path` looks ugly in git workflow, but currently is best way :(

hsluoyz:
Let's wait for Go team's response first: https://github.com/golang/go/issues/39376

divy9881:
@hsluoyz have you merged the commits of this PR, with the master

hsluoyz:
@divy9881 No. I removed the commits in this PR via git rewrite.

hsluoyz:
We need to implement it in a way that doesn't cause break change.

divy9881:
How do you propose to do so? : )

hsluoyz:
@nodece do you know such a way?

nodece:
I don't have anything way. It seems that only `github.com/casbin/casbin/v3` can be published.

PR #475, opened 2020-05-30
feat: Support pattern function in 3rd args of g

dovics:
This commit not only supports pattern function in 3rd args of g, but also can be used in policy, such as

```
p, data1_admin, *, data1, read
p, data1_admin, *, data1, write

g, alice, data1_admin, domain1
```

### DIFF
--- a/rbac/default-role-manager/role_manager.go
+++ b/rbac/default-role-manager/role_manager.go
@@ -23,23 +23,30 @@ import (
 	"github.com/casbin/casbin/v2/rbac"
 )
 
+const defaultDomain string = "casbin::default"
+const tempDomain string = "casbin::temp"
+
 type MatchingFunc func(arg1, arg2 string) bool
 
 // RoleManager provides a default implementation for the RoleManager interface
 type RoleManager struct {
-	allRoles          *sync.Map
-	maxHierarchyLevel int
-	hasPattern        bool
-	matchingFunc      MatchingFunc
+	allRoles           map[string]*sync.Map
+	maxHierarchyLevel  int
+	hasPattern         bool
+	matchingFunc       MatchingFunc
+	hasDomainPattern   bool
+	domainMatchingFunc MatchingFunc
 }
 
 // NewRoleManager is the constructor for creating an instance of the
 // default RoleManager implementation.
 func NewRoleManager(maxHierarchyLevel int) rbac.RoleManager {
 	rm := RoleManager{}
-	rm.allRoles = &sync.Map{}
+	rm.allRoles = make(map[string]*sync.Map)
+	rm.allRoles[defaultDomain] = &sync.Map{}
 	rm.maxHierarchyLevel = maxHierarchyLevel
 	rm.hasPattern = false
+	rm.hasDomainPattern = false
 
 	return &rm
 }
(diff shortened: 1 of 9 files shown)


dovics:
Looks like there are still some problems, I will fix it immediately

hsluoyz:
@dovics I think `p, data1_admin, *, data1, read` is not needed to support. We actually have keyMatch() and regexMatch() to support such requirements. Supporting * natively will only make things ambigious.

dovics:
> @dovics I think `p, data1_admin, *, data1, read` is not needed to support. We actually have keyMatch() and regexMatch() to support such requirements. Supporting * natively will only make things ambigious.

It may be that I am not expressing clearly enough, what I mean here is that we can use pattern in policy.domain，Otherwise the user needs to write

```
p, data1_admin, domain1, data1, read
p, data1_admin, domain1, data1, write
p, data1_admin, domain2, data1,read
p, data1_admin, domain2, data1, write

p, alice, data_admin1, domian1
```

The `*` can be replaced by any pattern string

nodece:
@dovics Can we only use the `AddMatchingFunc` to support pattern?

hsluoyz:
@dovics OK. I understood it now. But how to know which is `dom`? What about `p, *, data1_admin, data1, read`? Or you only recognize it from model: `p = sub, dom, obj, act`'s `dom`?

dovics:
> @dovics OK. I understood it now. But how to know which is `dom`? What about `p, *, data1_admin, data1, read`? Or you only recognize it from model: `p = sub, dom, obj, act`'s `dom`?

@hsluoyz  the matcher is 
```
g(r.sub, p.sub, r.dom) && globMatch(r.dom, p.dom) && r.obj == p.obj && r.act == p.act 
```
The data in memory is like this
```
domain1: data1_admin -> domian1: alice
```
The key to being able to write like this is `globMatch(r.dom, p.dom)`.If there is no this pr, we can write this too.  But this is not pointed out in the doc

dovics:
> @dovics Can we only use the `AddMatchingFunc` to support pattern?

I merged the two functions in the new commit

rbac/default-role-manager/role_manager.go
```diff
@@ -40,40 +41,57 @@ func NewRoleManager(maxHierarchyLevel int) rbac.RoleManager {
...
 
 // e.BuildRoleLinks must be called after AddMatchingFunc().
 //
 // example: e.GetRoleManager().(*defaultrolemanager.RoleManager).AddMatchingFunc('matcher', util.KeyMatch)
-func (rm *RoleManager) AddMatchingFunc(name string, fn MatchingFunc) {
-	rm.hasPattern = true
-	rm.matchingFunc = fn
+// If need to add Matchingfunc to the domain, add true at the end.
+// example: e.GetRoleManager().(*defaultrolemanager.RoleManager).AddMatchingFunc('matcher', util.KeyMatch, true)
+func (rm *RoleManager) AddMatchingFunc(name string, fn MatchingFunc, forDomain ...bool) error {
```
nodece, line 53: I don't want to add new parameters, but I'm not sure can we support all the args of g at the same time.

GopherJ:
can we do every thing in `AddLink` function?

dovics:
@GopherJ 
Let me first state my implementation
If call `HasLink(alice, admin, domain1)` We will get the following relationship
```
domain1::admin
       |
    *::admin
       |
    *::alice
       |
domain1:alice
```
but I just need one matchFunc, and It is easy to support case 3
```
domain1::book_admin
       |
  *::book_admin
       |
  *::/book/:id
       |
domain::/book/1
```

I added a new function to support it

In addition, I removed the part about `p, data1_admin, *, data1, read` in the unit test and examples

rbac/default-role-manager/role_manager.go
dovics: It is not yet to support the 2nd parameter, but the 1st and 3rd can be used at the same time, through the new commit

dovics: What do you think this API should be, domain and sub/obj may be different matching function, we must add them apart

nodece: Sorry @dovics , I made the wrong decision. You are right, we must use different matching function for each args.

nodece:
> Do you really think we have to deal with it in createRole()?

 @dovics  Because our previous design can't meet the current feature, so you don't have to follow the current design.

GopherJ:
@nodece @dovics 

Since I'm afraid of having bad bench on pattern(because we always need to iterate over all). I'm thinking if we can change all_roles' structure.

Here some of my thoughts
image

nodece:
I'm worried about that, too.  Your idea looks good.

dovics:
@GopherJ I think It's a good idea, but I still have a little question 

>  if a domain is a pattern domain (eg. *). we clone its graph's reference to matched domain.

if we create the matched domain first, and then create pattern domain,  we couldn't clone the pattern domain's graph to matched domain, so we still need create the pattern domain first, and then merge the two graph to matched domain.

so I think the general step should be
1. Find the domain that not use matching func
2. Add the link,
3. Find the mathced domain or pattern domain, Merge the pattern domain's graph and the matched domain's graph, and store it to the matched domain

If I'm missing something, please point it out

GopherJ:
@dovics I have been exploring my solution for a while but I didn't succeed to find a simple way to merge two graphs. simple and efficient.

dovics:
@GopherJ I don't see a good enough way either, but I think we could use the worst way to add link one by one to the new graph, which is still more efficient than this implementation. I'm going to implement the current idea what you say now, and then take longer to think of a good enough solution

hsluoyz:
@dovics @nodece please update this exciting change to our docs: https://casbin.org/docs/en/rbac#use-pattern-matching-in-rbac

nodece:
@hsluoyz  I have a PR for the feature here: https://github.com/casbin/casbin-website/pull/102

hsluoyz:
@nodece awesome!

Issue #323, opened 2019-11-11
Concurrency issues with custom adapter


rolandjitsu:
I'm not exactly sure if it's because of the custom adapter I use, but I keep getting:
```
fatal error: concurrent map writes

goroutine 998 [running]:
runtime.throw(0x5ace54, 0x15)
	/tmp/go/1.12.6/go/src/runtime/panic.go:617 +0x5c fp=0x20d18d8 sp=0x20d18c4 pc=0x3ef68
runtime.mapassign_faststr(0x548b60, 0x21bf0e0, 0x5a286e, 0x1, 0x21c6368)
	/tmp/go/1.12.6/go/src/runtime/map_faststr.go:291 +0x424 fp=0x20d1908 sp=0x20d18d8 pc=0x239f8
github.com/casbin/casbin/v2.(*Enforcer).enforce(0x21bcc80, 0x0, 0x0, 0x21cef20, 0x3, 0x3, 0x214920c, 0x5a37da, 0x4)
	/home/pi/go/pkg/mod/github.com/casbin/casbin/v2@v2.1.0/enforcer.go:343 +0x128 fp=0x20d1adc sp=0x20d1908 pc=0x17bd4c
github.com/casbin/casbin/v2.(*Enforcer).Enforce(...)
	/home/pi/go/pkg/mod/github.com/casbin/casbin/v2@v2.1.0/enforcer.go:486
api/auth.(*CasbinConfig).CheckPermission(0x21c8cb0, 0x677dc0, 0x2228410, 0x83, 0x90, 0x0)
	/tmp/go/1.12.6/packages/src/api/auth/casbin_middleware.go:67 +0x154 fp=0x20d1b20 sp=0x20d1adc pc=0x48cb2c
api/auth.CasbinMiddlewareWithConfig.func1.1(0x677dc0, 0x2228410, 0x2228410, 0x21cee01)
	/tmp/go/1.12.6/packages/src/api/auth/casbin_middleware.go:51 +0x64 fp=0x20d1b44 sp=0x20d1b20 pc=0x48e928
api/auth.TokenMiddlewareWithConfig.func1.1(0x677dc0, 0x2228410, 0x4, 0x545890)
	/tmp/go/1.12.6/packages/src/api/auth/token_middleware.go:47 +0xf4 fp=0x20d1b68 sp=0x20d1b44 pc=0x48ed98
[...]

PR #479, opened 2020-06-02
perf: Use read lock in synced enforcer

postulha:
Use read lock in `Enforce()` method instead of write lock in synced enforcer.  

Write lock blocks other readers and writers, while read lock do not block other readers. So calling `SyncedEnforcer.Enforce()` can be a performance bottleneck when using `SyncedEnforcer` as singleton, for example.

Previously the read lock was in this method, and in [this PR](https://github.com/casbin/casbin/pull/310) it was changed to write lock due to error related to concurrent map writes. But the error was caused by bug in nested `Enforcer` instance, not synced enforcer. And this bug has been already fixed [here](https://github.com/casbin/casbin/issues/323).  

So we can use read lock again.

### DIFF
--- a/enforcer_synced.go
+++ b/enforcer_synced.go
@@ -125,8 +125,8 @@ func (e *SyncedEnforcer) BuildRoleLinks() error {
 
 // Enforce decides whether a "subject" can access a "object" with the operation "action", input parameters are usually: (sub, obj, act).
 func (e *SyncedEnforcer) Enforce(rvals ...interface{}) (bool, error) {
-	e.m.Lock()
-	defer e.m.Unlock()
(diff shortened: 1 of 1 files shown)


hsluoyz:
@nodece please review.

nodece:
LGTM

nodece:
@postulha Please fix the Semantic Pull Request.

postulha:
@nodece Done

nodece:
Hi @postulha, due to we revert #472 , so you need reset your code.  Sorry for the inconvenience caused to you.

nodece:
LGTM

PR #480, opened 2020-06-03
fix: Add Static Check for Enforcer Implementations

will7200:
Adds static check for Synced Enforcers and Cached
Fix SyncedEnforcer to correctly implement EnforcerInterface

### DIFF
--- a/enforcer_interface.go
+++ b/enforcer_interface.go
@@ -23,6 +23,8 @@ import (
 )
 
 var _ IEnforcer = &Enforcer{}
+var _ IEnforcer = &SyncedEnforcer{}
+var _ IEnforcer = &CachedEnforcer{}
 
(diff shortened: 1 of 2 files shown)


enforcer_interface.go
```diff
@@ -23,6 +23,8 @@ import (
 )
 
 var _ IEnforcer = &Enforcer{}
+var _ IEnforcer = &SyncedEnforcer{}
+var _ IEnforcer = &CachedEnforcer{}
```
nodece, line 27: I'm not sure if correct. @hsluoyz

hsluoyz: This is correct, see: https://stackoverflow.com/questions/10498547/ensure-a-type-implements-an-interface-at-compile-time-in-go

PR #485, opened 2020-06-05
fix: Necessary code changes in buildIncrementalRoleLinks.

DivyPatel9881:
There is no need to record ```effects``` in ```AddPolicies``` and ```RemovePolicies``` as we know that the operations are atomic, so either all rules are successful or none.

### DIFF
--- a/internal_api.go
+++ b/internal_api.go
@@ -61,13 +61,13 @@ func (e *Enforcer) addPolicy(sec string, ptype string, rule []string) (bool, err
 // addPolicies adds rules to the current policy.
 // removePolicies removes rules from the current policy.
 func (e *Enforcer) addPolicies(sec string, ptype string, rules [][]string) (bool, error) {
-	rulesAdded, effects := e.model.AddPolicies(sec, ptype, rules)
+	rulesAdded := e.model.AddPolicies(sec, ptype, rules)
 	if !rulesAdded {
 		return rulesAdded, nil
 	}
 
 	if sec == "g" {
-		err := e.BuildIncrementalRoleLinks(model.PolicyAdd, ptype, effects)
+		err := e.BuildIncrementalRoleLinks(model.PolicyAdd, ptype, rules)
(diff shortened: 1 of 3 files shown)


nodece:
Hi @DivyPatel9881 , I forgot where we discussed the question. 

For add action, we should add policy that are not in model, ignore existing policies.
For remove action, we should remove policy that are in model, Ignore nonexistent policies. 

So we should record `effects`.

DivyPatel9881:
@nodece, then what you are suggesting is that AddPolicies and RemovePolicies operations are no longer atomic?
@hsluoyz are you okay with this? While updating this operations, we had a discussion that this operations must be atomic, right?

GopherJ:
@DivyPatel9881 @nodece  non-atmoic requires adapters being able to return correct number of inserted/removed policies (ignoring those exist already).

It's a nice feature but will be breaking changes (adapter interface)

nodece:
adapter should be atomic.

DivyPatel9881:
Yeah, so if the operations are atomic then there is no need to record effects in ```AddPolicies``` and ```RemovePolicies```, but ```RemoveFilteredPolicy``` needs recording effects.
So, I think if it is okay, then this PR is ready to merge.

GopherJ:
adapter can be non-atomic as what we've discussed but in this case, it needs to return back what it has done. (how many policies have it inserted/deleted)

DivyPatel9881:
>adapter can be non-atomic as what we've discussed but in this case, it needs to return back what it has done. (how many policies have it inserted/deleted) so that model can do the same thing.

@GopherJ, I completely understand what you are saying, we can implement the operations, non-atomic by recording successfull rules, but we shouldn't, we had discussed this before here, #388. The operations should be necessarily atomic in this case.

GopherJ:
@DivyPatel9881 It should be the same thing no? If adapter can return what it has done by using transaction.

PR #488, opened 2020-06-09
fix: issue#469

RobotHuang:


role_manager_b_test.go
```diff
@@ -0,0 +1,95 @@
...
+
+	// 1000 users.
+	gPolicies := make([][]string, 0)
+	for i := 0; i < 1000; i++ {
+		gPolicies = append(gPolicies, []string{fmt.Sprintf("user%d", i), fmt.Sprintf("group%d", i/10)})
+	}
+
+	e.BuildRoleLinks()
+
+	rm := defaultrolemanager.NewRoleManager(1)
```
nodece, line 30: It should be `e.GetRoleManager()`

role_manager_b_test.go
```diff
@@ -0,0 +1,95 @@
...
+
+	for i := 0; i < b.N; i++ {
+		for j := 0; j < 100; j++ {
```
nodece, line 32: I think it's better to use random number, like `rm.HasLink("user501", fmt.Sprintf("group%d", rand.Intn(100)))`. 

What do you think?

role_manager_b_test.go
```diff
@@ -0,0 +1,94 @@
...
+		gPolicies = append(gPolicies, []string{fmt.Sprintf("user%d", i), fmt.Sprintf("group%d", i/10)})
+	}
+
+	e.BuildRoleLinks()
+
+	rm := e.GetRoleManager()
+
+	for i := 0; i < b.N; i++ {
+		for j := 0; j < 1000; j++ {
+			rm.HasLink("user501", fmt.Sprintf("group%d", j))
```
nodece, line 62: the same as above

role_manager_b_test.go
```diff
@@ -0,0 +1,94 @@
...
+		gPolicies = append(gPolicies, []string{fmt.Sprintf("user%d", i), fmt.Sprintf("group%d", i/10)})
+	}
+
+	e.BuildRoleLinks()
+
+	rm := e.GetRoleManager()
+	
+	for i := 0; i < b.N; i++ {
+		for j := 0; j < 10000; j++ {
+			rm.HasLink("user501", fmt.Sprintf("group%d", j))
```
nodece, line 91: the same as the above.

role_manager_b_test.go
RobotHuang: I reference https://github.com/casbin/casbin-rs/blob/b3bd688fd7fbb3b0c497b0ad255e129ed549b945/benches/benchmark.rs#L144, and he seems to use a loop. And he wants to compare the casbin's bench with casbin-rs's bench

dovics:
Hi @RobotHuang @nodece If the benchmark test should in `rbac/default-role-manager/role_manager_b_test.go`

and we shouldn't use `Enforcer` to addLink,  the `pPolicy` is useless, I think `rm.AddLink(fmt.Sprintf("user%d", i), fmt.Sprintf("group%d", i/10))` is better 

What do you think?

role_manager_b_test.go
```diff
@@ -0,0 +1,94 @@
...
+	// 100000 users.
+	gPolicies := make([][]string, 0)
+	for i := 0; i < 100000; i++ {
+		gPolicies = append(gPolicies, []string{fmt.Sprintf("user%d", i), fmt.Sprintf("group%d", i/10)})
+	}
+
+	e.BuildRoleLinks()
+
+	rm := e.GetRoleManager()
+
```
dovics, line 88: Should here be `b.ResetTimer()` ？we just need the time for `HasLink`

role_manager_b_test.go
```diff
@@ -0,0 +1,94 @@
...
+	// 1000 users.
+	gPolicies := make([][]string, 0)
+	for i := 0; i < 1000; i++ {
+		gPolicies = append(gPolicies, []string{fmt.Sprintf("user%d", i), fmt.Sprintf("group%d", i/10)})
+	}
+
+	e.BuildRoleLinks()
+
+	rm := e.GetRoleManager()
+
```
dovics, line 30: Should here be b.ResetTimer() ？we just need the time for HasLink

nodece: Yes, please add `b.ResetTimer()` here.

nodece:
@RobotHuang @dovics We just need to create a role_manager var then test that `AddLink`, `DeleteLink`, `HasLink`, and so on.

role_manager_b_test.go
GopherJ: loop there is to simulate enforce.

Because what we need to know is: *how much time has been taken up by role manager during an enforce*.

PR #490, opened 2020-06-11
fix: Perform adapter persist actions before model changes

### LINKED ISSUE #415: Fix:  adding policy to adapter before trying to add into model
Recently I'm having issue with the following code in `internal_api.go`

```go
func (e *Enforcer) addPolicy(sec string, ptype string, rule []string) (bool, error) {
	ruleAdded := e.model.AddPolicy(sec, ptype, rule)
	if !ruleAdded {
		return ruleAdded, nil
	}

	if e.adapter != nil && e.autoSave {
		if err := e.adapter.AddPolicy(sec, ptype, rule); err != nil {
			if err.Error() != notImplemented {
				return ruleAdded, err
			}
		}
	}

	if e.watcher !=nil && e.autoNotifyWatcher {
		err := e.watcher.Update()
		if err != nil {
			return ruleAdded, err
		}
	}

	return ruleAdded, nil
}
```

It seems logic, but it may cause bugs while having network or io issues. For example it fails to add policy into adapter for some reason but it succeeds to add policy into model. Then it return an error to say that we don't succeed to add policy but actually we already added policies into model and it will influence the enforce result.

To solve this, shall we add policy into adapter first? I mean when auto save has been enabled.

speza:
Should resolve #415.

> If the model changes are made first but the adapter fails (network, code error, anything really), the in-memory model and adapter store will be out of sync. The next time the adapter is instantiated it will call LoadPolicy(x) in the store and the changes made in-memory will obviously be lost.

This change moves the adapter action to be first. The mutation of in-memory state is done only if the store does not error (not implemented should be fine still).

### DIFF
--- a/internal_api.go
+++ b/internal_api.go
@@ -25,6 +25,14 @@ const (
 
 // addPolicy adds a rule to the current policy.
 func (e *Enforcer) addPolicy(sec string, ptype string, rule []string) (bool, error) {
+	if e.adapter != nil && e.autoSave {
+		if err := e.adapter.AddPolicy(sec, ptype, rule); err != nil {
+			if err.Error() != notImplemented {
+				return false, err
+			}
+		}
+	}
+
 	ruleAdded := e.model.AddPolicy(sec, ptype, rule)
 	if !ruleAdded {
 		return ruleAdded, nil
@@ -34,14 +42,6 @@ func (e *Enforcer) addPolicy(sec string, ptype string, rule []string) (bool, err
 		err := e.BuildIncrementalRoleLinks(model.PolicyAdd, ptype, [][]string{rule})
 		if err != nil {
 			return ruleAdded, err
-		}
-	}
-
-	if e.adapter != nil && e.autoSave {
-		if err := e.adapter.AddPolicy(sec, ptype, rule); err != nil {
-			if err.Error() != notImplemented {
-				return ruleAdded, err
-			}
 		}
 	}
 
@@ -60,6 +60,14 @@ func (e *Enforcer) addPolicy(sec string, ptype string, rule []string) (bool, err
 
 // addPolicies adds rules to the current policy.
 func (e *Enforcer) addPolicies(sec string, ptype string, rules [][]string) (bool, error) {
+	if e.adapter != nil && e.autoSave {
+		if err := e.adapter.(persist.BatchAdapter).AddPolicies(sec, ptype, rules); err != nil {
+			if err.Error() != notImplemented {
+				return false, err
+			}
+		}
+	}
+
 	rulesAdded := e.model.AddPolicies(sec, ptype, rules)
 	if !rulesAdded {
 		return rulesAdded, nil
@@ -69,14 +77,6 @@ func (e *Enforcer) addPolicies(sec string, ptype string, rules [][]string) (bool
 		err := e.BuildIncrementalRoleLinks(model.PolicyAdd, ptype, rules)
 		if err != nil {
 			return rulesAdded, err
-		}
-	}
-
-	if e.adapter != nil && e.autoSave {
-		if err := e.adapter.(persist.BatchAdapter).AddPolicies(sec, ptype, rules); err != nil {
-			if err.Error() != notImplemented {
-				return rulesAdded, err
-			}
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


hsluoyz:
@GopherJ @nodece @xcaptain please review.

internal_api.go
```diff
@@ -25,6 +25,14 @@ const (
 
 // addPolicy adds a rule to the current policy.
 func (e *Enforcer) addPolicy(sec string, ptype string, rule []string) (bool, error) {
+	if e.adapter != nil && e.autoSave {
```
nodece, line 28: I recommend call model.HasPolicy to check if the policy can be added to adapter.

internal_api.go
```diff
@@ -160,6 +160,14 @@ func (e *Enforcer) removePolicies(sec string, ptype string, rules [][]string) (b
 
 // removeFilteredPolicy removes rules based on field filters from the current policy.
 func (e *Enforcer) removeFilteredPolicy(sec string, ptype string, fieldIndex int, fieldValues ...string) (bool, error) {
+	if e.adapter != nil && e.autoSave {
```
nodece, line 163: the same as above

internal_api.go
```diff
@@ -128,6 +128,14 @@ func (e *Enforcer) removePolicy(sec string, ptype string, rule []string) (bool,
 
 // removePolicies removes rules from the current policy.
 func (e *Enforcer) removePolicies(sec string, ptype string, rules [][]string) (bool, error) {
+	if e.adapter != nil && e.autoSave {
```
nodece, line 131: the same as above

internal_api.go
```diff
@@ -92,6 +92,14 @@ func (e *Enforcer) addPolicies(sec string, ptype string, rules [][]string) (bool
 
 // removePolicy removes a rule from the current policy.
 func (e *Enforcer) removePolicy(sec string, ptype string, rule []string) (bool, error) {
+	if e.adapter != nil && e.autoSave {
```
nodece, line 95: the same as above

internal_api.go
```diff
@@ -60,6 +60,14 @@ func (e *Enforcer) addPolicy(sec string, ptype string, rule []string) (bool, err
 
 // addPolicies adds rules to the current policy.
 func (e *Enforcer) addPolicies(sec string, ptype string, rules [][]string) (bool, error) {
+	if e.adapter != nil && e.autoSave {
```
nodece, line 63: the same as above

internal_api.go
speza: @nodece my only question around this is that this check is already done in the model mutation code. So we'd be duplicating that work. In the case of AddPolicies (or another plural action) we'd be potentially iterating over everything to check again. That isn't great.

We could move the ` model.HasPolicy(...)` check outside of the functions in policy.go?

nodece: I know it's  duplicate work. If not checked, It will add duplicate policy :(  
`adapter` may have added a unique index on table, but I'm not sure.

nodece: > We could move the model.HasPolicy(...) check outside of the functions in policy.go?

You can try.

speza: I think there is 2 easy options.

1. We move all `model.HasPolicy(...)` checks outside of `policy.go` into the calling code. We would also have to drop the `policy.go` mutation returning a `bool` as it doesn't make sense anymore.

2. Push adapter actions into `policy.go`. This doesn't feel great to me, but it is another option.

nodece: I have an idea, Let's add a method returns effects policy in policy.go.

speza: Not sure I follow @nodece? Do you mean call a func from within `policy.go` that will save to the adapter?

nodece: Sounds good,  could you try?

speza: I will have a look a little bit later today mate :)

nodece: @speza Looks good. 

cc @hsluoyz

speza: @nodece I don't think its particularly pretty! But it allows us to achieve calling the persist adapter without having to pass the whole enforcer or adatper through. Problem is that codeclimate dislikes it! I fear that doing anything nicer at this stage will require a lot more refactoring.

nodece: It's breaking model struct, we also can add a new method in model, like `AddPolicyWith persist`.

nodece:
> have we already switched to `O(1)` for inserting/removing policy?

Not supported :(  

we need hash the policy then add it to map.

internal_api.go
speza: I think that would increase code duplication quite a bit.

speza: Ok, I think the latest change is nicer @nodece @hsluoyz. This now leaves `policy.go` as the model mutator only. I moved the mutation pre-conditions up into `internal_api.go`. Personally, I think this is nicer as we're not mixing adapter logic in the `policy.go` model methods.

PR #495, opened 2020-06-16
Fix: Avoid typecast adapter to BatchAdapter for RemovePolicy

sanoopsugathan:
Avoid typecast to BatchAdapter for RemovePolicy


nodece:
LGTM

Issue #499, opened 2020-06-19
performance:  switch to O(1) for inserting/removing policy 


nodece:
## Summary
https://github.com/casbin/casbin/issues/455#issuecomment-646519444

PR #500, opened 2020-06-20
fix: use map to store policy index

dovics:
fix: #499

### DIFF
--- a/model/assertion.go
+++ b/model/assertion.go
@@ -17,6 +17,7 @@ package model
 import (
 	"errors"
 	"strings"
+	"sync"
 
 	"github.com/casbin/casbin/v2/log"
 	"github.com/casbin/casbin/v2/rbac"
@@ -31,9 +32,12 @@ type Assertion struct {
 	Policy    [][]string
 	PolicyMap map[string]int
 	RM        rbac.RoleManager
+	Mutex     sync.RWMutex
 }
 
 func (ast *Assertion) buildIncrementalRoleLinks(rm rbac.RoleManager, op PolicyOp, rules [][]string) error {
+	ast.Mutex.RLock()
+	defer ast.Mutex.RUnlock()
 	ast.RM = rm
 	count := strings.Count(ast.Value, "_")
 	if count < 2 {
@@ -65,6 +69,8 @@ func (ast *Assertion) buildIncrementalRoleLinks(rm rbac.RoleManager, op PolicyOp
 }
 
 func (ast *Assertion) buildRoleLinks(rm rbac.RoleManager) error {
+	ast.Mutex.RLock()
+	defer ast.Mutex.RUnlock()
 	ast.RM = rm
 	count := strings.Count(ast.Value, "_")
 	if count < 2 {
--- a/model/assertion.go
+++ b/model/assertion.go
@@ -25,11 +25,12 @@ import (
 // Assertion represents an expression in a section of the model.
 // For example: r = sub, obj, act
 type Assertion struct {
-	Key    string
-	Value  string
-	Tokens []string
-	Policy [][]string
-	RM     rbac.RoleManager
+	Key       string
+	Value     string
+	Tokens    []string
+	Policy    [][]string
+	PolicyMap map[string]int
+	RM        rbac.RoleManager
 }
 
 func (ast *Assertion) buildIncrementalRoleLinks(rm rbac.RoleManager, op PolicyOp, rules [][]string) error {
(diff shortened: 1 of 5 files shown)


hsluoyz:
@nodece @GopherJ @UnderTreeTech please review.

model/policy.go
```diff
@@ -123,34 +124,37 @@ func (model Model) HasPolicies(sec string, ptype string, rules [][]string) bool
...
 
 // RemovePolicies removes policy rules from the model.
 func (model Model) RemovePolicies(sec string, ptype string, rules [][]string) bool {
-	for j := 0; j < len(rules); j++ {
-		for i, r := range model[sec][ptype].Policy {
-			if util.ArrayEquals(rules[j], r) {
-				model[sec][ptype].Policy = append(model[sec][ptype].Policy[:i], model[sec][ptype].Policy[i+1:]...)
-			}
+	for _, rule := range rules {
+		if ok := model.RemovePolicy(sec, ptype, rule); !ok {
```
nodece, line 156: I think the return value should be ignored.

nodece:
Hi @dovics , Could you add benchmark test for the changes?  I want to know if the performance has improved.

model/policy.go
```diff
@@ -123,34 +124,37 @@ func (model Model) HasPolicies(sec string, ptype string, rules [][]string) bool
...
-			return true
-		}
+	index, ok := model[sec][ptype].PolicyMap[strings.Join(rule, DefaultSep)]
+	if !ok {
+		return false
 	}
 
-	return false
+	model[sec][ptype].Policy = append(model[sec][ptype].Policy[:index], model[sec][ptype].Policy[index+1:]...)
+	delete(model[sec][ptype].PolicyMap, strings.Join(rule, DefaultSep))
```
UnderTreeTech, line 145: concurrent delete map element will panic

UnderTreeTech:
It's exactly a great improvement, thanks for your work.

model/policy.go
nodece: Caspin should use thread safe instance by default.

dovics:
The Benchmark previous
Screen Shot 2020-06-20 at 12 25 43 PM

The Benchmark later
Screen Shot 2020-06-20 at 12 28 20 PM

UnderTreeTech:
> The Benchmark previous
> Screen Shot 2020-06-20 at 12 25 43 PM
> 
> The Benchmark later
> Screen Shot 2020-06-20 at 12 28 20 PM

Exciting!

model/policy.go
dovics: I think the issue of thread safety should be provided by SyncedEnforcer. Before, the model was not thread safe. Even if we guarantee that map is thread-safe, policy slice still requires addition another lock

model/assertion.go
```diff
@@ -25,11 +25,12 @@ import (
...
-	Value  string
-	Tokens []string
-	Policy [][]string
-	RM     rbac.RoleManager
+	Key       string
+	Value     string
+	Tokens    []string
+	Policy    [][]string
+	PolicyMap map[string]int
+	RM        rbac.RoleManager
```
nodece, line 33: Could you add `sync.RWMutex` to the Assertion?

model/policy.go
nodece: @dovics you're right.

model/assertion.go
dovics: ok, I will add it

dovics: I still think this mutex has some duplicates with SyncedEnforce

nodece: I see, we can new a  PR to improve it.

model/policy.go
```diff
@@ -122,46 +140,54 @@ func (model Model) HasPolicies(sec string, ptype string, rules [][]string) bool
...
+	defer model[sec][ptype].Mutex.Unlock()
 	model[sec][ptype].Policy = append(model[sec][ptype].Policy, rule)
+	model[sec][ptype].PolicyMap[strings.Join(rule, DefaultSep)] = len(model[sec][ptype].Policy) - 1
 }
 
 // AddPolicies adds policy rules to the model.
 func (model Model) AddPolicies(sec string, ptype string, rules [][]string) {
-	for i := 0; i < len(rules); i++ {
-		model[sec][ptype].Policy = append(model[sec][ptype].Policy, rules[i])
+	for _, rule := range rules {
```
nodece, line 151: Add lock here.

nodece:
@dovics Did you miss the https://github.com/casbin/casbin/pull/500/files#diff-abe2d810a9c1f824dbc2aa7b8e3afd26R252?

PR #512, opened 2020-07-05
add missing functions to SyncedEnforcer

PhilippSeitz:
add AddGroupingPolicies, AddNamedGroupingPolicies, AddPolicies & AddNamedPolicies to SyncedEnforcer

btw. it can be pretty confusing to used the `SyncedEnforcer` assuming all functions are synced.

https://github.com/casbin/casbin/blob/7841fd31a667614446024033dc03185dbd793a88/enforcer_interface.go#L25-L27

this won't fail since

https://github.com/casbin/casbin/blob/7841fd31a667614446024033dc03185dbd793a88/enforcer_synced.go#L28-L33

i would suggest to wrap every single function instead and keep the enforcer private

```go
type SyncedEnforcer struct {
	e               *Enforcer
	m               sync.RWMutex
	stopAutoLoad    chan struct{}
	autoLoadRunning int32
}
```

### DIFF
--- a/enforcer_synced.go
+++ b/enforcer_synced.go
@@ -375,3 +375,39 @@ func (e *SyncedEnforcer) AddFunction(name string, function govaluate.ExpressionF
 	defer e.m.Unlock()
 	e.Enforcer.AddFunction(name, function)
 }
+
+// AddGroupingPolicies adds role inheritance rulea to the current policy.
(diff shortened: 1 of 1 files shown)


nodece:
@PhilippSeitz LGTM, We added the plan to v3.

PR #514, opened 2020-07-07
fix: add explain index on allow

PhilippSeitz:


hsluoyz:
@PhilippSeitz please fix the semantic PR error:

image

GopherJ:
Seems since certain commit we have only one explainIndex available, but it's not the case for `some(p.eft == allow) && !some(p.eft == deny)` effect.

GopherJ:
we have lost this pr: https://github.com/casbin/casbin/pull/472

nodece:
We have planed to support the feature on v3, so close the PR.

PR #515, opened 2020-07-08
perf: Parallelize Policy Evaluation

aguss787:
Fix: https://github.com/casbin/casbin/issues/516

In uses cases that policy evaluation is bottle necked in IO operation (such as database query/API calls), it is much faster to parallelize the evaluation. This trade speed in other cases, due to regeneration of several entity. However, performance cost is small compared to the benefit that we get from parallelizing the entire evaluation, such as big number of policy or IO-heavy operation.

### DIFF
--- a/enforcer.go
+++ b/enforcer.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"runtime"
 	"strings"
 	"sync"
 
@@ -176,7 +177,7 @@ func (e *Enforcer) initialize() {
 	e.autoSave = true
 	e.autoBuildRoleLinks = true
 	e.autoNotifyWatcher = true
-	e.parallel = false
+	e.parallel = true
 }
 
 // LoadModel reloads the model from the model CONF file.
@@ -426,30 +427,52 @@ func (e *Enforcer) enforce(matcher string, explains *[]string, rvals ...interfac
 		if e.parallel {
 			ctx, cancel := context.WithCancel(context.Background())
 
+			n := runtime.GOMAXPROCS(0)
+
 			var wg sync.WaitGroup
-			wg.Add(len(e.model["p"]["p"].Policy))
-			errs := make(chan error, len(e.model["p"]["p"].Policy))
+			wg.Add(n)
 
-			for i, pvals := range e.model["p"]["p"].Policy {
-				go func(i int, pvals []string, parameters enforceParameters, expression *govaluate.EvaluableExpression) {
-					defer wg.Done()
-
-					select {
-					case <-ctx.Done():
-						return
-					default:
-					}
-
-					parameters.pVals = pvals
-					var err error
-					policyEffects[i], matcherResults[i], err = e.evaluatePolicy(parameters, expression, expString, functions)
-					if err != nil {
-						errs <- err
-						cancel()
-					}
-				}(i, pvals, parameters, expression)
+			type param struct {
+				i          int
+				parameters enforceParameters
+				expression *govaluate.EvaluableExpression
 			}
 
+			inputs := make(chan param, n)
+			errs := make(chan error, len(e.model["p"]["p"].Policy))
+
+			for i := 0; i < n; i++ {
+				go func() {
+					defer wg.Done()
+
+					for input := range inputs {
+						select {
+						case <-ctx.Done():
+							return
+						default:
+						}
+
+						var err error
+						policyEffects[input.i], matcherResults[input.i], err = e.evaluatePolicy(input.parameters, input.expression, expString, functions)
+						if err != nil {
+							errs <- err
+							cancel()
+						}
+					}
+				}()
+			}
+
+			for i, pvals := range e.model["p"]["p"].Policy {
+				input := param{
+					i:          i,
+					parameters: parameters,
+					expression: expression,
+				}
+				input.parameters.pVals = pvals
+				inputs <- input
+			}
+
+			close(inputs)
 			wg.Wait()
 
 			close(errs)
--- a/enforcer.go
+++ b/enforcer.go
@@ -15,9 +15,11 @@
 package casbin
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"strings"
+	"sync"
 
 	"github.com/Knetic/govaluate"
 	"github.com/casbin/casbin/v2/effect"
@@ -45,6 +47,7 @@ type Enforcer struct {
 	autoSave           bool
 	autoBuildRoleLinks bool
 	autoNotifyWatcher  bool
+	parallel           bool
 }
 
 // NewEnforcer creates an enforcer via file or DB.
@@ -173,6 +176,7 @@ func (e *Enforcer) initialize() {
 	e.autoSave = true
 	e.autoBuildRoleLinks = true
 	e.autoNotifyWatcher = true
+	e.parallel = false
 }
 
 // LoadModel reloads the model from the model CONF file.
@@ -334,6 +338,11 @@ func (e *Enforcer) EnableAutoSave(autoSave bool) {
 	e.autoSave = autoSave
 }
 
+// EnableParallelEnforcing controls whether to evaluate the policies in parallel when enforcing.
+func (e *Enforcer) EnableParallelEnforcing(parallel bool) {
+	e.parallel = parallel
+}
+
 // EnableAutoBuildRoleLinks controls whether to rebuild the role inheritance relations when a role is added or deleted.
 func (e *Enforcer) EnableAutoBuildRoleLinks(autoBuildRoleLinks bool) {
 	e.autoBuildRoleLinks = autoBuildRoleLinks
@@ -366,13 +375,8 @@ func (e *Enforcer) enforce(matcher string, explains *[]string, rvals ...interfac
 		return true, nil
 	}
 
-	functions := e.fm.GetFunctions()
-	if _, ok := e.model["g"]; ok {
-		for key, ast := range e.model["g"] {
-			rm := ast.RM
-			functions[key] = util.GenerateGFunction(rm)
-		}
-	}
+	functions := e.generateFunctions()
+
 	var expString string
 	if matcher == "" {
 		expString = e.model["m"]["m"].Value
@@ -382,7 +386,6 @@ func (e *Enforcer) enforce(matcher string, explains *[]string, rvals ...interfac
 
 	var expression *govaluate.EvaluableExpression
 	hasEval := util.HasEval(expString)
-
 	if !hasEval {
 		expression, err = govaluate.NewEvaluableExpressionWithFunctions(expString, functions)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


enforcer.go
```diff
@@ -334,6 +337,11 @@ func (e *Enforcer) EnableAutoSave(autoSave bool) {
 	e.autoSave = autoSave
 }
 
+// EnableAutoSave controls whether to save a policy rule automatically to the adapter when it is added or removed.
```
GopherJ, line 340: The comment here is incorrect

enforcer.go
```diff
@@ -556,6 +529,96 @@ func (e *Enforcer) enforce(matcher string, explains *[]string, rvals ...interfac
...
+
+		expression, err = govaluate.NewEvaluableExpressionWithFunctions(expWithRule, functions)
+		if err != nil {
+			return nil, fmt.Errorf("p.sub_rule should satisfy the syntax of matcher: %s", err)
+		}
+	}
+	return expression, nil
+}
+
+func (e *Enforcer) generateFunctions() map[string]govaluate.ExpressionFunction {
```
GopherJ, line 610: Go prefers uppercase for the first character

enforcer.go
aguss787: Whoops, my bad. forgot to change it when copying it.
I wrote the comment to match the function name, I will change the description if it is changed.

enforcer.go
```diff
@@ -334,6 +337,11 @@ func (e *Enforcer) EnableAutoSave(autoSave bool) {
 	e.autoSave = autoSave
 }
 
+// EnableAutoSave controls whether to save a policy rule automatically to the adapter when it is added or removed.
+func (e *Enforcer) EnableParallelPolicyEvaluation(parallel bool) {
```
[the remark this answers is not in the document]
nodece: This is a good idea.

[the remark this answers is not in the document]
aguss787: Okay! :+1:

enforcer.go
```diff
@@ -419,78 +421,50 @@ func (e *Enforcer) enforce(matcher string, explains *[]string, rvals ...interfac
...
 			}
 
-			result, err := expression.Eval(parameters)
-			// log.LogPrint("Result: ", result)
+			go func() {
+				wg.Wait()
+				close(errs)
+			}()
+
+			var err error
```
nodece, line 447: I don't thinks this is correct.  We need wait evaluatePolicy is done then check errors.

nodece: please use `ctx, cancel := context.WithCancel(context.Background())` to control logic.

enforcer.go
```diff
@@ -419,78 +421,50 @@ func (e *Enforcer) enforce(matcher string, explains *[]string, rvals ...interfac
...
+			for cerr := range errs {
```
nodece, line 448: We just need to get first err, somethings like: `err := <- errrs`.

aguss787: Okay :+1:

enforcer.go
aguss787: I update it according to your example. But, I'm not really sure if I understand your review correctly. Kindly check it again :)

enforcer.go
```diff
@@ -419,78 +422,54 @@ func (e *Enforcer) enforce(matcher string, explains *[]string, rvals ...interfac
...
 			}
 
-			result, err := expression.Eval(parameters)
-			// log.LogPrint("Result: ", result)
+			wg.Wait()
 
-			if err != nil {
+			select {
+			case err := <-errs:
 				return false, err
```
nodece, line 457: I think use the following code instead of your code is better.
```golang
close(errs)
err := <-errs
if err != nil {
    return false, err
}

Comment (aguss787) [reply enforcer.go:458]: Okay, updated.

**Comment by hsluoyz** [comment]
@aguss787 can you post the result of our built-in benchmark tests to see how much performance is improved after this PR?

**Comment by aguss787** [comment]
Hi @hsluoyz 

Here is the benchmark w/o parallelization:
```
goos: linux
goarch: amd64
pkg: github.com/casbin/casbin/v2
BenchmarkRaw
BenchmarkRaw-8                          	48629660	        24.8 ns/op
BenchmarkBasicModel
BenchmarkBasicModel-8                   	   99822	     12190 ns/op
BenchmarkRBACModel
BenchmarkRBACModel-8                    	   66802	     17588 ns/op
BenchmarkRBACModelSmall
BenchmarkRBACModelSmall-8               	    7240	    160548 ns/op
BenchmarkRBACModelMedium
BenchmarkRBACModelMedium-8              	     945	   1253905 ns/op
BenchmarkRBACModelLarge
BenchmarkRBACModelLarge-8               	      61	  19974960 ns/op
BenchmarkRBACModelWithResourceRoles
BenchmarkRBACModelWithResourceRoles-8   	   57669	     20385 ns/op
BenchmarkRBACModelWithDomains
BenchmarkRBACModelWithDomains-8         	   44973	     25011 ns/op
BenchmarkABACModel
BenchmarkABACModel-8                    	  200053	      6088 ns/op
BenchmarkKeyMatchModel
BenchmarkKeyMatchModel-8                	   55480	     21724 ns/op
BenchmarkRBACModelWithDeny
BenchmarkRBACModelWithDeny-8            	   62979	     18602 ns/op
BenchmarkPriorityModel
BenchmarkPriorityModel-8                	   88984	     13805 ns/op
```

and here is with parallelization:

```
goos: linux
goarch: amd64
pkg: github.com/casbin/casbin/v2
BenchmarkRaw
BenchmarkRaw-8                          	49546116	        24.7 ns/op
BenchmarkBasicModel
BenchmarkBasicModel-8                   	   47355	     24429 ns/op
BenchmarkRBACModel
BenchmarkRBACModel-8                    	   25101	     46916 ns/op
BenchmarkRBACModelSmall
BenchmarkRBACModelSmall-8               	    3139	    424567 ns/op
BenchmarkRBACModelMedium
BenchmarkRBACModelMedium-8              	     274	   3828752 ns/op
BenchmarkRBACModelLarge
BenchmarkRBACModelLarge-8               	      31	  38290352 ns/op
BenchmarkRBACModelWithResourceRoles
BenchmarkRBACModelWithResourceRoles-8   	   24807	     46726 ns/op
BenchmarkRBACModelWithDomains
BenchmarkRBACModelWithDomains-8         	   20874	     55602 ns/op
BenchmarkABACModel
BenchmarkABACModel-8                    	  196093	      6196 ns/op
BenchmarkKeyMatchModel
BenchmarkKeyMatchModel-8                	   25377	     47220 ns/op
BenchmarkRBACModelWithDeny
BenchmarkRBACModelWithDeny-8            	   21535	     54006 ns/op
BenchmarkPriorityModel
BenchmarkPriorityModel-8                	   16716	     69853 ns/op
PASS
```

for fast to evaluate policies, it seems to be slower than the sequential one (but it is still useful for slow evaluated spolicies).
when I trace it, I found that for half of the time, it only run in one thread:

go

do you have any idea on whats happened here?

**Comment by nodece** [comment]
great job, the feature has greatly improved the performance for Casbin.

**Comment by zhiyuan-mojie** [comment]
Do you have a try about parallelizing policy evaluation with a grouped policy by policy slice index handled by  goroutines pool? every goroutine deal with part of the policy. for example, there are 100 policies handle by 10 goroutines and the the ninety-first policy match the request. at this condition, we can get the result quickly and every goroutines maybe only handle one policy .

**Comment by nodece** [comment]
current, the number of goroutine equals is the numer of policy, we must be able to control the number of goroutine.  when we have too many policy, I worry the performance will decline.

**Comment by hsluoyz** [comment]
I agree with @nodece

**Comment by aguss787** [comment]
Sure, I will try those approach. I did try making a go routine pool, but it also have that same problem. I will try it again and put the result here.

**Comment by aguss787** [comment]
Sure, I will try those approach. I did try making a go routine pool, but it also have that same problem. I will try it again and put the result here.

**Comment by zhiyuan-mojie** [comment]
> current, the number of goroutine equals is the numer of policy, we must be able to control the number of goroutine. when we have too many policy, Casbin will create same number of goroutine, I worry the performance will decline.

we can use the maximum number of CPUs that can be executing
simultaneously  and   get the number through runtime.GOMAXPROCS(0)

**Comment by nodece** [comment]
I'm developing v3.0.0,  I'm interesting the feature, Could you make a new PR that merge it into release-3.0.0 branch? I will improve the performance. 

What do you think? @aguss787

**Comment by GopherJ** [comment]
fasthttp has its workpool implementation which can be used as reference. https://github.com/valyala/fasthttp/blob/master/workerpool.go

**Comment by aguss787** [comment]
Sorry for the delay, it has been a busy week for me.

@nodece sure, I'll do it

@GopherJ sure! I will take a look at it

**Comment by aguss787** [comment]
@GopherJ I tried to do similar thing with the example that you gave me, I think it is better (?)
```
goos: linux
goarch: amd64
pkg: github.com/casbin/casbin/v2
BenchmarkRaw
BenchmarkRaw-8                          	47808964	        23.9 ns/op
BenchmarkBasicModel
BenchmarkBasicModel-8                   	   52347	     22968 ns/op
BenchmarkRBACModel
BenchmarkRBACModel-8                    	   28047	     42985 ns/op
BenchmarkRBACModelSmall
BenchmarkRBACModelSmall-8               	    5056	    211381 ns/op
BenchmarkRBACModelMedium
BenchmarkRBACModelMedium-8              	     790	   1472552 ns/op
BenchmarkRBACModelLarge
BenchmarkRBACModelLarge-8               	      61	  18470174 ns/op
BenchmarkRBACModelWithResourceRoles
BenchmarkRBACModelWithResourceRoles-8   	   26574	     46162 ns/op
BenchmarkRBACModelWithDomains
BenchmarkRBACModelWithDomains-8         	   24559	     49239 ns/op
BenchmarkABACModel
BenchmarkABACModel-8                    	  215286	      5562 ns/op
BenchmarkKeyMatchModel
BenchmarkKeyMatchModel-8                	   27358	     45205 ns/op
BenchmarkRBACModelWithDeny
BenchmarkRBACModelWithDeny-8            	   27560	     43470 ns/op
BenchmarkPriorityModel
BenchmarkPriorityModel-8                	   25513	     46236 ns/op
PASS
```

**Comment by GopherJ** [comment]
I think your benchmark test suite is not correct, parallel enforcing will only have advantage when there are really big number of policies, but in your test suite it's not the case. That's why I believe a threshold is necessary.

File: enforcer.go [hunk 7]
```diff
@@ -419,78 +423,76 @@ func (e *Enforcer) enforce(matcher string, explains *[]string, rvals ...interfac
...
-					} else {
-						return false, errors.New("please make sure rule exists in policy when using eval() in matcher")
+			inputs := make(chan param, n)
+			errs := make(chan error, len(e.model["p"]["p"].Policy))
+
+			for i := 0; i < n; i++ {
+				go func() {
+					defer wg.Done()
+
+					for input := range inputs {
```
nodece, line 448: I don't think this is correct. You create N goroutine to wait data from other goroutine, but we only have a datasource. N goroutine to do the same thing.  What do you think?

nodece: The following is my implement:
```golang
n := runtime.GOMAXPROCS(0)
jobC := make(chan struct{}, n)
for i, pvals := range e.model["p"]["p"].Policy {
	jobC <- struct{}{}

	go func(done chan struct{}) {
		defer func() {
			<-done
		}()
		// do eval policy
	}(jobC)
}
````

nodece:
thank you for your contribution, it's important for us, the PR has been inactive, so I decided to close the PR, but I will migrate it to Casbin v3.

Issue #518, opened 2020-07-14
allowing the parsing of policy file to deal with commas inside columns


misu200:
From forum: https://forum.casbin.org/t/13

If I have in Model .conf like:
`matchers m = eval(p.sub_rule) && r.obj == p.obj && r.act == p.act`

then in policy file the following stuff works: 
`p, my_custom_func(r.sub.Age), /data1, read`

BUT not the following kind of stuff where my custom function has at least 2 parameters:
` p, my_custom_func(r.sub.Age,r.sub.MaxAge), /data1, read`

....because the policy file is splitted by the ',' (COMMA) and my_custom_func takes 2 parameters ..

Casbin error: " panic: invalid policy size: ...."

hsluoyz:
@00LT00 @dovics can anyone take this issue?

GopherJ:
@misu200 There are some options you can use:

1. inline your custom function and make sure it's an expression without comma seperator, in the first version of `scale ABAC feature` we would like to only support this format.

2. try to use `""` to contain your `my_custom_func(r.sub.Age,r.sub.MaxAge)` to see if it works. I believe still no, because we need to consider removing `\"` to make it works.

hsluoyz:
I have a good way: use regex's split (https://yourbasic.org/golang/split-string-into-slice/#split-by-whitespace-and-newline) to split on all commas except the ones inside brackets:

a, b, c --> [a, b, c]
a, func(b, c) --> [a, func(b, c)]

Can anyone implement it?

hsluoyz:
@misu200 fixed in: https://github.com/casbin/casbin/pull/521

PR #521, opened 2020-07-15
fix: should remove escaped double quotes

GopherJ:
Signed-off-by: Cheng JIANG <jiang.cheng@vip.163.com>

Try Fix: #518

### DIFF
--- a/log/default_logger.go
+++ b/log/default_logger.go
@@ -25,15 +25,15 @@ type DefaultLogger struct {
 }
 
 func (l *DefaultLogger) EnableLog(enable bool) {
-    i := 0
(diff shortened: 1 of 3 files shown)


GopherJ:
I believe this will still not fix issue, because it'll probably split by "," which breaks things

hsluoyz:
I like the lint part.. But the trim call is not fixing the issue. Can I ask to keep the lint part only?

GopherJ:
I believe it already fixed the issue

hsluoyz:
Oh. I missed the 2nd commit. Never mind.

PR #548, opened 2020-07-25
fix: Call Enforcer method ClearPolicy.

divy9881:
Calling the Enforcer method 'ClearPolicy' would be more relevant than calling the model's method.

### DIFF
--- a/enforcer.go
+++ b/enforcer.go
@@ -243,7 +243,7 @@ func (e *Enforcer) ClearPolicy() {
 
 // LoadPolicy reloads the policy from file/database.
 func (e *Enforcer) LoadPolicy() error {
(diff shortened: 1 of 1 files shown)


nodece:
`e.ClearPolicy()` will save policy to storage container, :( so I decide to close the PR.

hsluoyz:
How is Go Casbin handling it? I'd like it to be similar with Go.

nodece:
We need to call `adapter.SavePolicy` in `e. ClearPolicy`, I think @divy9881 should fix the code.

hsluoyz:
Go ahead!

nodece:
No activity for a long time, closed.

PR #558, opened 2020-08-06
fix: update enforcer logic when there is no policy found

lucas-57blocks:
When there is no policy found:
1. Provide better error message when using eval() in matcher.
2. It is unnecessary to throw a null pointer error when use EnforceEx function, "explainIndex" is return by interface, should check this value, make sure that this value does not break the rest of the logic.


hsluoyz:

nodece (approved):
lgtm

Issue #560, opened 2020-08-07
Domain pattern and LoadFilteredPolicy


DarkSorrow:
Hello,

I'm still trying to implement something with domains.
I've been using the domain pattern because i might have a lot of domains and wanted to have the right to create some superdomain.

I'm trying to use an adapter such as this one https://github.com/casbin/casbin-pg-adapter
So far it seems to work however it seems that casbin erase everything when you start a loadFilteredPolicy, the problem is that the user from the domain pattern get erased as well so each time i make a loadFilterPolicy i lose the super users but also on the higher picture, each time i will want to start a domain it means that my enforcer will lose the rights existing for all other domains.
Is there some better way to just load a subset of policies and keep them alive until refresh by a new loadFilter on the domain or is it something that would require some custom code on the adapter instead?

hsluoyz:
@nodece @00LT00

hsluoyz:
@DarkSorrow if this is the adapter issue, can you send the issue here instead? https://github.com/casbin/casbin-pg-adapter/issues

DarkSorrow:
Well i was thinking about doing it on the adapter and checked most adapter but they all work a bit the same, they seems to erase previous value, i somehow thing that i can bypass this by making my own query and translation but i don't know if there would be some possibilities of having something in the core that erase the policyline filtered only instead of the whole policy line. Because in the end the filtered loading is a bit problematic if you plan to have multiple refresh on a domain and such.

But then maybe there is something that i don't see? My use case is having pattern domain matching that can be loaded and "reloaded" when changed by a client

hsluoyz:
@DarkSorrow can you make a PR?

wingdog:
I have the same issue and this is where clear policy happens:
https://github.com/casbin/casbin/blob/master/enforcer.go#L263

For me it seems to be more like a bug than a feature request.

DarkSorrow:
@hsluoyz i'll try to put what i made but it's not well tested yet cause i'm on holidays xD. I kinda fork the repo tho because i'm using pgx for postgres but i should be able to isolate the code otherwise i'll just publish my adapter i suppose

But still i think it could be nice to have just a "clearFilteredPolicy" or something like that to remove only the filtered policy and keep the rest still in the enforcer :x do you think it's doable? I can try to take a look too maybe

nodece:
@DarkSorrow  I think you want to keep old role_manager then load filtered data by call LoadFilteredPolicy.

[...]

PR #571, opened 2020-08-18
Added option to disabled clear policies

DarkSorrow:
Adding an option to disable the clear policies when we load filtered one
This should be transparent and keep old code and current existing adapter working the same way
https://github.com/casbin/casbin/issues/560


DarkSorrow:
Okay i'm not sure what i'm missing with the Semantic PR, wasn't it just the title the problem :x?

GopherJ:
I think add a field to `Filter` is better

```
Filter {
...,
ClearExistingPolicy: false
}
```

Or just add a new method `LoadIncrementalFilteredPolicy` etc

DarkSorrow:
I was thinking of adding a new method too before but i was thinking that for those that update the package the new function will give the adapter an error when they load it no?
After both is fine with me but the reason i added it in the main enforcer is because the piece of code that calls the ClearPolicy was also outside the filterAdapter itself and adding something new to the adapter would make all current adapter not working if they update no?

GopherJ:
no, you don't  need to have LoadIncrementalFilteredPolicy for adapter because it's an enforcer thing

DarkSorrow:
Okay i'm sorry for the DCO i know what i need to do but the semantic Pull Request i don't get it =/

PR #587, opened 2020-09-02
feat: add mutex to model

dovics:

### DIFF
--- a/enforcer.go
+++ b/enforcer.go
@@ -35,7 +35,7 @@ import (
 // Enforcer is the main interface for authorization enforcement and policy management.
 type Enforcer struct {
 	modelPath string
-	model     model.Model
+	model     *model.Model
 	fm        model.FunctionMap
 	eft       effect.Effector
 
@@ -99,7 +99,7 @@ func NewEnforcer(params ...interface{}) (*Enforcer, error) {
 			case string:
 				return nil, errors.New("invalid parameters for enforcer")
 			default:
-				err := e.InitWithModelAndAdapter(p0.(model.Model), params[1].(persist.Adapter))
+				err := e.InitWithModelAndAdapter(p0.(*model.Model), params[1].(persist.Adapter))
 				if err != nil {
 					return nil, err
 				}
@@ -113,7 +113,7 @@ func NewEnforcer(params ...interface{}) (*Enforcer, error) {
 				return nil, err
 			}
 		default:
-			err := e.InitWithModelAndAdapter(p0.(model.Model), nil)
+			err := e.InitWithModelAndAdapter(p0.(*model.Model), nil)
 			if err != nil {
 				return nil, err
 			}
@@ -150,7 +150,7 @@ func (e *Enforcer) InitWithAdapter(modelPath string, adapter persist.Adapter) er
 }
 
 // InitWithModelAndAdapter initializes an enforcer with a model and a database adapter.
-func (e *Enforcer) InitWithModelAndAdapter(m model.Model, adapter persist.Adapter) error {
+func (e *Enforcer) InitWithModelAndAdapter(m *model.Model, adapter persist.Adapter) error {
 	e.adapter = adapter
 
 	e.model = m
@@ -200,12 +200,12 @@ func (e *Enforcer) LoadModel() error {
 }
 
 // GetModel gets the current model.
-func (e *Enforcer) GetModel() model.Model {
+func (e *Enforcer) GetModel() *model.Model {
 	return e.model
 }
 
 // SetModel sets the current model.
-func (e *Enforcer) SetModel(m model.Model) {
+func (e *Enforcer) SetModel(m *model.Model) {
 	e.model = m
 	e.fm = model.LoadFunctionMap()
 
@@ -432,16 +432,10 @@ func (e *Enforcer) enforce(matcher string, explains *[][]string, rvals ...interf
 		return true, nil
 	}
 
-	functions := e.fm.GetFunctions()
-	if _, ok := e.model["g"]; ok {
-		for key, ast := range e.model["g"] {
-			rm := ast.RM
-			functions[key] = util.GenerateGFunction(rm)
-		}
-	}
+	functions := e.model.GenerateFunctions(e.fm)
 	var expString string
 	if matcher == "" {
-		expString = e.model["m"]["m"].Value
+		expString = e.model.GetMatcher()
 	} else {
 		expString = util.RemoveComments(util.EscapeAssertion(matcher))
 	}
@@ -456,14 +450,8 @@ func (e *Enforcer) enforce(matcher string, explains *[][]string, rvals ...interf
 		}
 	}
 
-	rTokens := make(map[string]int, len(e.model["r"]["r"].Tokens))
-	for i, token := range e.model["r"]["r"].Tokens {
-		rTokens[token] = i
-	}
-	pTokens := make(map[string]int, len(e.model["p"]["p"].Tokens))
-	for i, token := range e.model["p"]["p"].Tokens {
-		pTokens[token] = i
-	}
+	rTokens := e.model.GetTokens("r", "r")
+	pTokens := e.model.GetTokens("p", "p")
 
 	parameters := enforceParameters{
 		rTokens: rTokens,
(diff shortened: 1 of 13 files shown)


hsluoyz:
@nodece @GopherJ @divy9881 please reivew.

enforcer.go
```diff
@@ -433,15 +433,17 @@ func (e *Enforcer) enforce(matcher string, explains *[][]string, rvals ...interf
 	}
 
 	functions := e.fm.GetFunctions()
-	if _, ok := e.model["g"]; ok {
-		for key, ast := range e.model["g"] {
+	e.model.Mutex.RLock()
+	defer e.model.Mutex.RUnlock()
+	if _, ok := e.model.Data["g"]; ok {
```
nodece, line 438: could you pack the `e.model.Data["g"]` to method?

enforcer.go
```diff
@@ -433,15 +433,17 @@ func (e *Enforcer) enforce(matcher string, explains *[][]string, rvals ...interf
...
+		for key, ast := range e.model.Data["g"] {
 			rm := ast.RM
 			functions[key] = util.GenerateGFunction(rm)
 		}
 	}
 	var expString string
 	if matcher == "" {
-		expString = e.model["m"]["m"].Value
+		expString = e.model.Data["m"]["m"].Value
```
nodece, line 446: +1

enforcer.go
```diff
@@ -456,12 +458,12 @@ func (e *Enforcer) enforce(matcher string, explains *[][]string, rvals ...interf
 		}
 	}
 
-	rTokens := make(map[string]int, len(e.model["r"]["r"].Tokens))
-	for i, token := range e.model["r"]["r"].Tokens {
+	rTokens := make(map[string]int, len(e.model.Data["r"]["r"].Tokens))
```
nodece, line 461: +1

model/policy.go
```diff
@@ -122,59 +138,69 @@ func (model Model) HasPolicies(sec string, ptype string, rules [][]string) bool
...
 	}
 
-	model[sec][ptype].Policy = append(model[sec][ptype].Policy[:index], model[sec][ptype].Policy[index+1:]...)
-	delete(model[sec][ptype].PolicyMap, strings.Join(rule, DefaultSep))
-	for i := index; i < len(model[sec][ptype].Policy); i++ {
-		model[sec][ptype].PolicyMap[strings.Join(model[sec][ptype].Policy[i], DefaultSep)] = i
+	model.Data[sec][ptype].Policy = append(model.Data[sec][ptype].Policy[:index], model.Data[sec][ptype].Policy[index+1:]...)
+	delete(model.Data[sec][ptype].PolicyMap, strings.Join(rule, DefaultSep))
+	for i := index; i < len(model.Data[sec][ptype].Policy); i++ {
+		model.Data[sec][ptype].PolicyMap[strings.Join(model.Data[sec][ptype].Policy[i], DefaultSep)] = i
```
GopherJ, line 175: It looks like an expensive operation, I'm wondering why dont we implement a class hashset using map.

Or use something like swap remove,  swap the policy with the last one and then remove only the last one. this way we avoid touching the others

model/policy.go
```diff
@@ -122,59 +138,69 @@ func (model Model) HasPolicies(sec string, ptype string, rules [][]string) bool
...
 			continue
 		}
 
-		model[sec][ptype].Policy = append(model[sec][ptype].Policy[:index], model[sec][ptype].Policy[index+1:]...)
-		delete(model[sec][ptype].PolicyMap, strings.Join(rule, DefaultSep))
-		for i := index; i < len(model[sec][ptype].Policy); i++ {
-			model[sec][ptype].PolicyMap[strings.Join(model[sec][ptype].Policy[i], DefaultSep)] = i
+		model.Data[sec][ptype].Policy = append(model.Data[sec][ptype].Policy[:index], model.Data[sec][ptype].Policy[index+1:]...)
+		delete(model.Data[sec][ptype].PolicyMap, strings.Join(rule, DefaultSep))
+		for i := index; i < len(model.Data[sec][ptype].Policy); i++ {
```
GopherJ, line 193: same here, looks like expensive operation

GopherJ (changes requested):

model/policy.go
dovics: The problem is that the result of casbin enforce is affected by the order of the policy, so we must store the order of the policy through slices, and this map is to speed up querying the policy position. If we want to change to the classic hashset, we need to give up this feature.

And this operation is to update the data in the map after removing the policy, so we must modify the position of each policy in the map one by one

GopherJ: We can use linked hash map to keep order and implement a hashset with it.  I believe it solves this problem. 

Also even we insist on this solution, we need to use at least a hash function instead if Join directly the policy which can be heavy to store.  It doubles the used memory

GopherJ: BTW, 

> The problem is that the result of casbin enforce is affected by the order of the policy, so we must store the order of the policy through slices, and this map is to speed up querying the policy position. If we want to change to the classic hashset, the result of the enforcer will be uncertain.


when we are enforcing, there isn't any change on policy and policy order (thanks lock).  I still think we can do swap remove instead of simple remove. This drops down from O(n) to O(1)

dovics: For swap remove，suppose we have such a policy
```
g, alice, data1_deny_group

p, bob, data2, read, deny
p, alice, data1, read, allow
p, data1_deny_group, data1, write, deny
p, data1_deny_group, data1, read, deny
// enforce(alice, data1, read) =>  true
```

If we call `remove(bob, data2, deny)`, then it should become

```
p, data1_deny_group, data1, read, deny
p, alice, data1, read, allow
p, data1_deny_group, data1, write, deny
// enforce(alice, data1, read) =>  false
```
I think it’s not good to remove one policy but affect the other

dovics: I remember that we once mentioned the linked hashmap when doing this https://github.com/casbin/casbin/pull/500, we should listen to the opinions of other people @hsluoyz @nodece  @divy9881 

I totally agree with storing the hash value, I will add it in another pr

model/model.go
```diff
@@ -18,14 +18,20 @@ import (
...
 	"github.com/casbin/casbin/v3/config"
 	"github.com/casbin/casbin/v3/log"
 	"github.com/casbin/casbin/v3/util"
 )
 
 // Model represents the whole access control model.
-type Model map[string]AssertionMap
+type Model struct {
+	Data  map[string]AssertionMap
+	Mutex sync.RWMutex
```
nodece, line 33: it should be private, we are not sure who is using the lock.

model/model.go
```diff
@@ -144,7 +153,7 @@ func (model Model) LoadModelFromText(text string) error {
 	return model.loadModelFromConfig(cfg)
 }
 
-func (model Model) loadModelFromConfig(cfg config.ConfigInterface) error {
+func (model *Model) loadModelFromConfig(cfg config.ConfigInterface) error {
 	for s := range sectionNameMap {
```
nodece, line 157: there are should add lock.

model/model.go
```diff
@@ -160,17 +169,28 @@ func (model Model) loadModelFromConfig(cfg config.ConfigInterface) error {
 	return nil
 }
 
-func (model Model) hasSection(sec string) bool {
-	section := model[sec]
+func (model *Model) hasSection(sec string) bool {
+	model.Mutex.RLock()
```
nodece, line 173: The lock seems unnecessary.

model/model.go
```diff
@@ -41,13 +47,15 @@ var sectionNameMap = map[string]string{
...
 	value := cfg.String(sectionNameMap[sec] + "::" + key)
 	return model.AddDef(sec, key, value)
 }
 
 // AddDef adds an assertion to the model.
-func (model Model) AddDef(sec string, key string, value string) bool {
+func (model *Model) AddDef(sec string, key string, value string) bool {
+	model.Mutex.Lock()
+	defer model.Mutex.Unlock()
 	if value == "" {
```
nodece, line 59: We should extract the following code to `addDef()`.

model/model.go
```diff
@@ -41,13 +47,15 @@ var sectionNameMap = map[string]string{
 // Minimal required sections for a model to be valid
 var requiredSections = []string{"r", "p", "e", "m"}
 
-func loadAssertion(model Model, cfg config.ConfigInterface, sec string, key string) bool {
+func loadAssertion(model *Model, cfg config.ConfigInterface, sec string, key string) bool {
 	value := cfg.String(sectionNameMap[sec] + "::" + key)
 	return model.AddDef(sec, key, value)
```
nodece, line 52: there are call `model.addDef(sec, key, value)`.

model/model.go
```diff
@@ -18,14 +18,20 @@ import (
...
 
 	"github.com/casbin/casbin/v3/config"
 	"github.com/casbin/casbin/v3/log"
 	"github.com/casbin/casbin/v3/util"
 )
 
 // Model represents the whole access control model.
-type Model map[string]AssertionMap
+type Model struct {
+	Data  map[string]AssertionMap
```
nodece, line 32: `Data` and `Mutex` should be private.

nodece: We should avoid users using `Data` directly.

model/policy.go
nodece: It is very important to explore a new structure to store policy, we can implement the feature in next PR.

model/policy.go
GopherJ: then we need a real hash set which keeps order. Go should have one.

model/policy.go
GopherJ: then we need a real hash set which keeps order. Go should have one.

model/model.go
dovics: I made them private, but in order to get policy when enforce, I added `GetAssertion()`. But this may also allow users to operate `Assertion`, what are your suggestions?

GopherJ:
@nodece @dovics I think we need to hide `Data` and `Mutex`, we shouldn't let users to call `Mutex` manually. This will probably be a big change to do because in adapter currently we just use  stuff like `model["p"]["p]` etc

nodece (approved):

PR #589, opened 2020-09-05
fix: Add ClearPolicy to Adapter

dovics:

### DIFF
--- a/persist/adapter.go
+++ b/persist/adapter.go
@@ -48,6 +48,8 @@ type Adapter interface {
 	LoadPolicy(model *model.Model) error
 	// SavePolicy saves all policy rules to the storage.
 	SavePolicy(model *model.Model) error
+	// ClearPolicy remove all policy rules from the storage.
+	ClearPolicy() error
 
 	// AddPolicy adds a policy rule to the storage.
 	// This is part of the Auto-Save feature.
--- a/enforcer.go
+++ b/enforcer.go
@@ -249,14 +249,14 @@ func (e *Enforcer) SetEffector(eft effect.Effector) {
 	e.eft = eft
 }
 
-func (e *Enforcer) IsAudoLoadRunning() bool {
+func (e *Enforcer) IsAutoLoadRunning() bool {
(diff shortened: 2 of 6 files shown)


hsluoyz:
Why we have this PR? Can you give us a pointer about its context?

dovics:
The main job is to allow `adapter.ClearPolicy` to be called before `model.clearPolicy`. This reason is the same as https://github.com/casbin/casbin/issues/415.
> If the model changes are made first but the adapter fails (network, code error, anything really), the in-memory model and adapter store will be out of sync. The next time the adapter is instantiated it will call LoadPolicy(x) in the store and the changes made in-memory will obviously be lost.

Also I fixed some typo by the way.

persist/adapter.go
```diff
@@ -48,6 +48,8 @@ type Adapter interface {
 	LoadPolicy(model *model.Model) error
 	// SavePolicy saves all policy rules to the storage.
 	SavePolicy(model *model.Model) error
+	// ClearPolicy remove all policy rules from the storage.
+	ClearPolicy() error
```
hsluoyz, line 52: This PR adds `ClearPolicy()` to the adapter interface and will cause compile error, which is a BREAKING change for ALL current adapter implementations.

dovics: Yes, I know, but https://github.com/casbin/casbin/pull/587 has caused the adapter to be re-implemented, because the model has become private, we can only operate through public func, I think it’s time finished this feature in casbin v3

nodece: I submitted a PR #590 for compatible with V2. @dovics

dovics: It is not fully compatible with the v2 adapter, because the way of operating the model needs to be changed from `model["p"]["p"]` to using public functions.

And do you think this modification is unnecessary? It may cause model is inconsistent with adapter.

nodece: @dovics  please refer to https://github.com/casbin/casbin/pull/592#discussion_r502783238.

hsluoyz:
Any update?

Issue #581, opened 2020-08-21
[Feature] Log actions, AddPolicy, RemovePolicy ...


GopherJ:
**Want to prioritize this issue? Try:**

[![issuehunt-to-marktext](https://github.com/BoostIO/issuehunt-materials/raw/master/v1/issuehunt-button-v1.svg)](https://issuehunt.io/r/casbin/casbin)

------

**Is your feature request related to a problem? Please describe.**
A clear and concise description of what the problem is. Ex. I'm always frustrated when casbin golang cannot have a clear and concise log on what's happening on the following events

- Enforce
- AddPolicy
- RemovePolicy
- AddPolicies
- RemovePolicies
- RemoveFilteredPolicy
- ClearCache

Currently this is possible with `watcher` but users need to implement their own logger, the built in logger doesn't support

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

hsluoyz:
Fixed by: https://github.com/casbin/casbin/pull/591

PR #591, opened 2020-09-07
fix: Improve logger for management api

dovics:
fix: https://github.com/casbin/casbin/issues/581

### DIFF
--- a/enforcer.go
+++ b/enforcer.go
@@ -299,7 +299,15 @@ func (e *Enforcer) ClearPolicy() error {
 	// TODO: implement ClearPolicy in adapter and move model.ClearPolicy after adapter.ClearPolicy
 	e.model.ClearPolicy()
 
-	return e.adapter.SavePolicy(e.model)
+	if err := e.adapter.SavePolicy(e.model); err != nil {
+		return err
+	}
+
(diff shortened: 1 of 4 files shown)


nodece:
Looks good to me, @hsluoyz, @GopherJ please take a look.

nodece (approved):

GopherJ:
https://github.com/casbin/casbin/pull/591/files#diff-72816617037c9418525dfa491f4c5f77L52-L66

I have a feeling that this place will be buggy. How will it handle ABAC struct?

For example, if I enforce two times:

1. enforce(Person {name: "alice"}, "data1", "read")
2. enforce(Person {name: "bob"}, "data1", "read")

It seems the cache key will be the same, because it'll skip `Person{..}` which is not a `string`, @nodece @dovics How do you think?

GopherJ:
Ok sorry I understood it'll not use cache if ABAC is founded

GopherJ (approved):

dovics:
> Why no use defer in this case to claim ABAC is not cached? Currently if it's not cached then there is no log. Do we need a log when it's not cached?

There will be a log here, it comes from https://github.com/casbin/casbin/blob/master/enforcer.go#L570

PR #595, opened 2020-09-09
fix: Separate internal_api

dovics:

Now the return value of internal_api is too much, this function need to processe include model, adapter, dispatcher, watcher and logger. I think this function should be simplified.

When the dispatcher updates the policy, it needs to update model and adapter one by one and call `BuildIncrementalRoleLinks()`, which makes the development of the plug-in more complicated, so I separated the main component model and adapter as a new api that can be called from outside.

### DIFF
--- a/enforcer.go
+++ b/enforcer.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/Knetic/govaluate"
 	"github.com/casbin/casbin/v3/effect"
+	"github.com/casbin/casbin/v3/internal"
 	"github.com/casbin/casbin/v3/log"
 	"github.com/casbin/casbin/v3/model"
 	"github.com/casbin/casbin/v3/persist"
@@ -44,6 +45,7 @@ type Enforcer struct {
 	dispatcher persist.Dispatcher
 	rm         rbac.RoleManager
 
+	internal             internal.PolicyManager
 	enabled              bool
 	autoSave             bool
 	autoBuildRoleLinks   bool
@@ -156,9 +158,8 @@ func (e *Enforcer) InitWithModelAndAdapter(m *model.Model, adapter persist.Adapt
 	e.model = m
 	e.model.PrintModel()
 	e.fm = model.LoadFunctionMap()
-
 	e.initialize()
-
+	e.internal = internal.NewPolicyManager(m, adapter, e.rm)
 	// Do not initialize the full policy when using a filtered adapter
 	fa, ok := e.adapter.(persist.FilteredAdapter)
 	if e.adapter != nil && (!ok || ok && !fa.IsFiltered()) {
@@ -208,7 +209,7 @@ func (e *Enforcer) GetModel() *model.Model {
 func (e *Enforcer) SetModel(m *model.Model) {
 	e.model = m
 	e.fm = model.LoadFunctionMap()
-
+	e.internal = internal.NewPolicyManager(m, e.adapter, e.rm)
 	e.initialize()
 }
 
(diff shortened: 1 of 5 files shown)


nodece:
@dovics I think we can extract the those methods to new dir, because of `internal_api.go` has too many methods.

dovics:
> @dovics I think we can extract the those methods to new dir, because of internal_api.go has too many methods.

@nodece This means that we need a new struct to control the model and adapter, I think this is not a good choice

If you want `internal_api.go` not to be so big, we can create a new file to separate these codes

nodece:
@dovics I don't hope that enforcer has `AddPolicyInternal`, `RemovePolicyInternal` and so on. It's going to make developers wonder what the difference is.

internal/internal.go
```diff
@@ -0,0 +1,171 @@
...
+
+import (
+	"github.com/casbin/casbin/v3/errors"
+	"github.com/casbin/casbin/v3/model"
+	"github.com/casbin/casbin/v3/persist"
+	"github.com/casbin/casbin/v3/rbac"
+)
+
+// InternalController is the internal controller for model and adapter
+type InternalController interface {
```
nodece, line 11: The name looks not clear, I think that we should rename InternalController to `PolicyManager`. 

@dovics @GopherJ @hsluoyz What do you think?

nodece: We should return the affected data when adding/removing multiple policies.

enforcer.go
```diff
@@ -220,6 +221,17 @@ func (e *Enforcer) GetAdapter() persist.Adapter {
 // SetAdapter sets the current adapter.
 func (e *Enforcer) SetAdapter(adapter persist.Adapter) {
 	e.adapter = adapter
+	e.internal = internal.NewInternalController(e.model, adapter, e.rm)
+}
+
+// SetInternalController sets the current internal controller.
+func (e *Enforcer) SetInternalController(internal internal.InternalController) {
```
nodece, line 228: It is unnecessary.

enforcer.go
```diff
@@ -220,6 +221,12 @@ func (e *Enforcer) GetAdapter() persist.Adapter {
 // SetAdapter sets the current adapter.
 func (e *Enforcer) SetAdapter(adapter persist.Adapter) {
 	e.adapter = adapter
+	e.internal = internal.NewPolicyManager(e.model, adapter, e.rm)
+}
+
+// GetInternalController gets the current internal controller.
+func (e *Enforcer) GetInternalController() internal.PolicyManager {
```
nodece, line 228: Rename GetInternalController  to PolicyManager.

go.mod
```diff
@@ -1,5 +1,8 @@
 module github.com/casbin/casbin/v3
 
-require github.com/Knetic/govaluate v3.0.1-0.20171022003610-9aa49832a739+incompatible
+require (
+	github.com/Knetic/govaluate v3.0.1-0.20171022003610-9aa49832a739+incompatible
+	github.com/casbin/casbin/v2 v2.11.3 // indirect
```
nodece, line 5: why add casbin/v2 to go.mod?

nodece (approved):
LGTM, please take a look @hsluoyz, @GopherJ

internal/internal.go
hsluoyz: Why not call it `InternalEnforcer`?

hsluoyz:
@nodece why not change `AddPolicyInternal, RemovePolicyInternal` to private?

internal/internal.go
nodece: @hsluoyz At present, casbin supports updating policy via enforcer.AddPolicy/RemovePolicy or get model way. For get model way, after updating the policy, users need to call `adapter`, `watcher`, this is complex for users, we have to simplify the operation, so we extract a layer for other plug-ins/user to update the policy, this is also because of casbin-raft. when the `dispatcher` is enabled, the ManagerAPI provided by casbin will forward the request to casbin-raft when policy changes.

nodece:
> @nodece why not change `AddPolicyInternal, RemovePolicyInternal` to private?

It must be public. casbin-raft needs it.

nodece:
@hsluoyz Do you have any suggestions?

nodece (approved):

PR #602, opened 2020-09-15
Update model_test.go

jamesislebron:
update test file avoid unexported error when registered func.


jamesislebron:
Well.
Finally I realized that  model_test.go and enforce.go are in the same package.

Issue #568, opened 2020-08-15
SyncedEnforcer missing functions & CachedEnforcer should overload functions to clear cache


GopherJ:
[<img alt="Issuehunt badges" src="https://img.shields.io/badge/IssueHunt-%241%20Funded-%2300A156.svg" />](https://issuehunt.io/r/casbin/casbin/issues/568)

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $1.00)

- [<img src='https://avatars0.githubusercontent.com/u/3787410?v=4' alt='hsluoyz' width=24 height=24> hsluoyz](https://issuehunt.io/u/hsluoyz) ($1.00)

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/casbin/casbin/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
</details>

GopherJ:
If we compare `enforcer_synced.go` with `enforcer_interface.go`, there are couple of functions missing, that means those functions may cause thread safety problems


CachedEnforcer currently don't clear cache automatically, we should overload functions

GopherJ:
Example: https://github.com/casbin/casbin-cpp/pull/61

dschmidt:
`CasbinJsGetPermissionForUser(...)` accepts only `*Enforcer`, it would be cool to be able to pass a `*SyncedEnforcer` to it.
For that at least `GetImplicitPermissionsForUser(...)` is missing in the `SyncedEnforcer`.

hsluoyz:
@00LT00 can you fix this issue?

/cc @nodece

00LT00:
@hsluoyz @dschmidt  plz review #605

00LT00:
@hsluoyz @nodece @dschmidt  Plz review #607 and #608

dschmidt:
> @hsluoyz @nodece @dschmidt Plz review #607 and #608

commented on #607, no idea about #608 :)

hsluoyz:
@00LT00 congrats! Please claim your $1 funding at: https://issuehunt.io/r/casbin/casbin/issues/568

PR #607, opened 2020-09-18
feat: Add CasbinJsGetPermissionForUserSynced for SyncedEnforcer

00LT00:

Fix: https://github.com/casbin/casbin/issues/568


frontend.go
```diff
@@ -30,3 +30,16 @@ func CasbinJsGetPermissionForUser(e *Enforcer, user string) ([]byte, error) {
...
+	if err != nil {
+		return nil, err
+	}
+	permission := make(map[string][]string)
+	for i := 0; i < len(policy); i++ {
+		permission[policy[i][2]] = append(permission[policy[i][2]], policy[i][1])
+	}
+	b, _ := json.Marshal(permission)
+	return b, nil
+}
```
dschmidt, line 45: I don't think the code should be copied just to accept an additional type.

I would either use the existing `IEnforcer` interface or add a simple interface in `frontend.go` that just requires the functions used in `CasbinJsGetPermissionForUser`, i.e. just `GetImplicitPermissionsForUser` right now.

enforcer_synced.go
```diff
@@ -418,3 +418,45 @@ func (e *SyncedEnforcer) AddNamedPolicies(ptype string, rules [][]string) (bool,
...
+
+// GetImplicitUsersForPermission gets implicit users for a permission.
+// For example:
+// p, admin, data1, read
+// p, bob, data1, read
+// g, alice, admin
+//
+// GetImplicitUsersForPermission("data1", "read") will get: ["alice", "bob"].
+// Note: only users will be returned, roles (2nd arg in "g") will be excluded.
+func (e *SyncedEnforcer) GetImplicitPermissionsForUser(user string, domain ...string) ([][]string, error) {
```
dschmidt, line 430: The documentation is for `GetImplicitUsersForPermission` but the function is called `GetImplicitPermissionsForUser`

frontend_test.go
```diff
@@ -91,3 +91,67 @@ func TestCasbinJsGetPermissionForUser(t *testing.T) {
...
+	}
+	if contains(perm, "data_not_exist") {
+		t.Errorf("Test error: Bob can access a non-existing data")
+	}
+
+	_, ok = bob_target["rm_rf"]
+	if ok {
+		t.Errorf("Someone can have a non-existing action (rm -rf)")
+	}
+}
```
dschmidt, line 157: This becomes obsolete as well without duplicating the whole of `CasbinJsGetPermissionForUser`.

It would be nice to add a test that just makes sure that `CasbinJsGetPermissionForUser` can be used with a `EnforcerSynced`

dschmidt (changes requested):

frontend.go
```diff
@@ -16,9 +16,14 @@ package casbin
...
 	"encoding/json"
+	"errors"
 )
 
-func CasbinJsGetPermissionForUser(e *Enforcer, user string) ([]byte, error) {
+func CasbinJsGetPermissionForUser(e IEnforcer, user string) ([]byte, error) {
+	_, ok := e.(IEnforcer)
+	if !ok {
+		return nil, errors.New("Unsupported type")
+	}
```
dschmidt, line 26: I believe this is dead code, if `e` is not of type `IEnforcer` the code won't compile. AFAICT you can just remove this check.

frontend_test.go
```diff
@@ -91,3 +91,67 @@ func TestCasbinJsGetPermissionForUser(t *testing.T) {
 		t.Errorf("Someone can have a non-existing action (rm -rf)")
 	}
 }
+
+func TestCasbinJsGetPermissionForUser2(t *testing.T) {
+	e, err := NewSyncedEnforcer("examples/rbac_model.conf", "examples/rbac_policy.csv")
+	if err != nil {
+		panic(err)
+	}
+	target_str, _ := CasbinJsGetPermissionForUser(e, "alice")
```
dschmidt, line 100: I think you can drop the rest of the test from here on, if it's just a copy of the other test.

Up until here it makes sure that CasbinJsGetPermissionForUser can be used with a synced enforcer.

dschmidt (changes requested):

frontend.go
00LT00: I'm sorry I forgot to use `interface{}` as a go developer...

dschmidt: No need to be sorry and also no need to use interface{} parameters here :)

frontend.go
00LT00: You are right.

frontend_test.go
00LT00: You are right.

00LT00:
plz review again @dschmidt

dschmidt (approved):
Cool, looking good IMHO :)

frontend_test.go
hsluoyz: Please assert the return value.

hsluoyz:
@kingiw please review.

dschmidt:
@kingiw ping :)

kingiw (approved):

kingiw:
Sorry for my late response.

PR #608, opened 2020-09-18
fix: Clear cache after update policies.

00LT00:

#568

### DIFF
--- a/enforcer_cached.go
+++ b/enforcer_cached.go
@@ -22,10 +22,9 @@ import (
 // CachedEnforcer wraps Enforcer and provides decision cache
 type CachedEnforcer struct {
 	*Enforcer
-	m              map[string]bool
-	enableCache    bool
-	autoClearCache bool
-	locker         *sync.RWMutex
+	m           map[string]bool
+	enableCache bool
+	locker      *sync.RWMutex
 }
 
 // NewCachedEnforcer creates a cached enforcer via file or DB.
@@ -38,7 +37,6 @@ func NewCachedEnforcer(params ...interface{}) (*CachedEnforcer, error) {
 	}
 
 	e.enableCache = true
-	e.autoClearCache = true
 	e.m = make(map[string]bool)
 	e.locker = new(sync.RWMutex)
 	return e, nil
@@ -47,11 +45,6 @@ func NewCachedEnforcer(params ...interface{}) (*CachedEnforcer, error) {
 // EnableCache determines whether to enable cache on Enforce(). When enableCache is enabled, cached result (true | false) will be returned for previous decisions.
 func (e *CachedEnforcer) EnableCache(enableCache bool) {
 	e.enableCache = enableCache
-}
-
-// AutoClear determines whether to clear cache after any operation for policy. if AutoClear is true, cache will be clear after update policy.
-func (e *CachedEnforcer) AutoClearCache(autoClearCache bool) {
-	e.autoClearCache = autoClearCache
 }
 
 // Enforce decides whether a "subject" can access a "object" with the operation "action", input parameters are usually: (sub, obj, act).
@@ -107,9 +100,7 @@ func (e *CachedEnforcer) InvalidateCache() {
 // Otherwise the function returns true by adding the new rule.
 func (e *CachedEnforcer) AddPolicy(params ...interface{}) (bool, error) {
 	ok, err := e.AddNamedPolicy("p", params...)
-	if e.autoClearCache {
-		e.InvalidateCache()
-	}
+	e.InvalidateCache()
 	return ok, err
 }
 
@@ -118,9 +109,7 @@ func (e *CachedEnforcer) AddPolicy(params ...interface{}) (bool, error) {
 // Otherwise the function returns true for the corresponding rule by adding the new rule.
 func (e *CachedEnforcer) AddPolicies(rules [][]string) (bool, error) {
 	ok, err := e.AddNamedPolicies("p", rules)
-	if e.autoClearCache {
-		e.InvalidateCache()
-	}
+	e.InvalidateCache()
 	return ok, err
 }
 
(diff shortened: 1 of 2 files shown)


hsluoyz:
@nodece @GopherJ @dovics please review.

enforcer_cached.go
```diff
@@ -22,9 +22,10 @@ import (
 // CachedEnforcer wraps Enforcer and provides decision cache
 type CachedEnforcer struct {
 	*Enforcer
-	m           map[string]bool
-	enableCache bool
-	locker      *sync.RWMutex
+	m              map[string]bool
+	enableCache    bool
```
GopherJ, line 26: `enableCache` is not necessary because cache must be cleaned

enforcer_cached.go
```diff
@@ -47,6 +49,11 @@ func (e *CachedEnforcer) EnableCache(enableCache bool) {
 	e.enableCache = enableCache
 }
 
+// AutoClear determines whether to clear cache after any operation for policy. if AutoClear is true, cache will be clear after update policy.
+func (e *CachedEnforcer) AutoClearCache(autoClearCache bool) {
```
GopherJ, line 53: not necessary

00LT00: I think there may be users who are used to the way. Without this, will some users be troubled?

nodece: when policy changes, we must clear cache, so `AutoClearCache`  is unnecessary.

00LT00: ok

00LT00:
plz review @GopherJ @nodece

enforcer_cached.go
```diff
@@ -93,3 +93,36 @@ func (e *CachedEnforcer) setCachedResult(key string, res bool) {
...
+
+// RemovePolicy removes an authorization rule from the current policy.
+func (e *CachedEnforcer) RemovePolicy(params ...interface{}) (bool, error) {
+	ok, err := e.RemoveNamedPolicy("p", params...)
+	e.InvalidateCache()
+	return ok, err
+}
+
+// RemovePolicies removes authorization rules from the current policy.
+func (e *CachedEnforcer) RemovePolicies(rules [][]string) (bool, error) {
```
GopherJ, line 124: How about `RemoveFilteredPolicy`?

GopherJ (approved):

enforcer_cached.go
```diff
@@ -93,3 +93,36 @@ func (e *CachedEnforcer) setCachedResult(key string, res bool) {
...
 	e.m = make(map[string]bool)
 }
+
+/* Clear cache after update Policy*/
+// AddPolicy adds an authorization rule to the current policy.
+// If the rule already exists, the function returns false and the rule will not be added.
+// Otherwise the function returns true by adding the new rule.
+func (e *CachedEnforcer) AddPolicy(params ...interface{}) (bool, error) {
+	ok, err := e.AddNamedPolicy("p", params...)
+	e.InvalidateCache()
```
GopherJ, line 103: also please check `if ok`, avoid unecessary clear is important

enforcer_cached.go
```diff
@@ -93,3 +93,36 @@ func (e *CachedEnforcer) setCachedResult(key string, res bool) {
...
+	return ok, err
+}
+
+// AddPolicies adds authorization rules to the current policy.
+// If the rule already exists, the function returns false for the corresponding rule and the rule will not be added.
+// Otherwise the function returns true for the corresponding rule by adding the new rule.
+func (e *CachedEnforcer) AddPolicies(rules [][]string) (bool, error) {
+	ok, err := e.AddNamedPolicies("p", rules)
```
GopherJ, line 111: check pls `ok`

enforcer_cached.go
```diff
@@ -93,3 +93,36 @@ func (e *CachedEnforcer) setCachedResult(key string, res bool) {
...
+	e.InvalidateCache()
+	return ok, err
+}
+
+// RemovePolicy removes an authorization rule from the current policy.
+func (e *CachedEnforcer) RemovePolicy(params ...interface{}) (bool, error) {
+	ok, err := e.RemoveNamedPolicy("p", params...)
```
GopherJ, line 118: check pls `ok`

enforcer_cached.go
```diff
@@ -93,3 +93,36 @@ func (e *CachedEnforcer) setCachedResult(key string, res bool) {
...
+	e.InvalidateCache()
+	return ok, err
+}
+
+// RemovePolicies removes authorization rules from the current policy.
+func (e *CachedEnforcer) RemovePolicies(rules [][]string) (bool, error) {
+	ok, err := e.RemoveNamedPolicies("p", rules)
```
GopherJ, line 125: check pls `ok`

enforcer_cached.go
00LT00: sure

enforcer_cached.go
00LT00: I don't quite understand. Does it mean to add this function?

GopherJ: you need to clear cache in RmoveFilteredPolicy

GopherJ:
I think this PR won't solve problems, because it works only when users call

1. AddPolicy
2. AddPolicies
3. RemovePolicy
4. RemovePolicies

how can we ensure it overwrites all the management/rbac apis? 

if users call e. AddRoleForUser, it doesn't find this function in cachedEnforcer 

so it'll still call EEnforcer's function

GopherJ:
See maybe how casbin-cpp and casbin-rs solved this issue

00LT00:
ok

00LT00:
But where the casbin-cpp's code. I see it doesn't clear cache after change policy. Or do you mean that cpp is also facing this problem now?

image

And the test is:
image

GopherJ:
@divy9881 @nodece Any idea on this?

Issue #603, opened 2020-09-16
[Question] eval support in policy


lucas-57blocks:
With the code in pr [fix: compatible RBAC in ABAC](https://github.com/casbin/casbin/pull/583), change the replace role of eval string in policy, why? How to support scenario  below? #583

@hsluoyz @GopherJ @nodece 

**Model:**
```ini
[request_definition]
r = user, url, method

[policy_definition]
p = role, ext, url, method

[policy_effect]
e = some(where (p.eft == allow))

[matchers]
m = (r.user.Role == p.role && eval(p.ext)) && keyMatch2(r.url, p.url) && regexMatch(r.method, p.method)
```

**Policy:**

```
p, employee, true, /user/profile, (GET)
p, employee, r.user.Age > 18, /user/game, (GET
```

**Request:**

```
{employee 10}, /user/profile, GET ---> false (expected: true)
```

hsluoyz:
@00LT00 can you reply to this question?

/cc @nodece

00LT00:
@lucas-57blocks Now we don't support `r.user.Role == p.role`  writing directly in Conf. So you can try 
Policy:
```
p, employee, r.user.Role == p.role, /user/profile, GET
p, employee, r.user.Age > 18, /user/game, GET
```
Model:
```
[matchers]
m = (eval(p.ext)) && keyMatch2(r.url, p.url) && regexMatch(r.method, p.method)
```
test.go
```go
type testUser struct {
	Age int
	Role string
}

func newTestUser(name int, role string) testUser {
	s := testUser{}
	s.Age = name
	s.Role = role
	return s
}

func Test(t *testing.T) {
	e, _ := NewEnforcer("examples/abac_with_rbac.conf", "examples/abac_with_rbac.csv")

	user := newTestUser(10, "employee")

	ok, err := e.Enforce(user, "/user/profile", "GET")
	if err != nil || ok != true {
		t.Fatal("expected Enforce returns true, nil, but got", ok, err)
	}
}
```
And it's ABAC in fact.

lucas-57blocks:
@00LT00  Now, the code is
```go
if strings.Contains(rule, ">") || strings.Contains(rule, "<") || strings.Contains(rule, "=") {
    expWithRule = util.ReplaceEval(expWithRule, rule)
} else {
    expWithRule = util.ReplaceEval(expWithRule, "false")
}
```
[...]

PR #609, opened 2020-09-19
feat: Add bool for ABAC

00LT00:
#603

### DIFF
--- a/enforcer.go
+++ b/enforcer.go
@@ -446,7 +446,7 @@ func (e *Enforcer) enforce(matcher string, explains *[]string, rvals ...interfac
 				for _, ruleName := range ruleNames {
 					if j, ok := parameters.pTokens[ruleName]; ok {
 						rule := util.EscapeAssertion(pvals[j])
-						if strings.Contains(rule, ">") || strings.Contains(rule, "<") || strings.Contains(rule, "=") {
+						if strings.Contains(rule, ">") || strings.Contains(rule, "<") || strings.Contains(rule, "=") || strings.Contains(rule, "true") {
 							expWithRule = util.ReplaceEval(expWithRule, rule)
 						} else {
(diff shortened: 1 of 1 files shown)


00LT00:
plz review @hsluoyz @lucas-57blocks  @nodece

GopherJ (changes requested):
Please iterate only one time instead

lucas-57blocks:
> Please iterate only one time instead if we insist on this solution.
> 
> I personally don't like this solution, it's users' responsibility to ensure expression being given in policy.

I couldn't agree more.

nodece:
The feature is compatible RBAC in ABAC: #583

enforcer.go
```diff
@@ -446,7 +446,7 @@ func (e *Enforcer) enforce(matcher string, explains *[]string, rvals ...interfac
 				for _, ruleName := range ruleNames {
 					if j, ok := parameters.pTokens[ruleName]; ok {
 						rule := util.EscapeAssertion(pvals[j])
-						if strings.Contains(rule, ">") || strings.Contains(rule, "<") || strings.Contains(rule, "=") {
+						if strings.Contains(rule, ">") || strings.Contains(rule, "<") || strings.Contains(rule, "=") || strings.Contains(rule, "true") {
```
nodece, line 449: miss `false`.

nodece:
@00LT00 It is better to use `fix`.

enforcer.go
00LT00: `false` will return "false" directly through the other branches. I think it is meaningless.

hsluoyz:
What does this PR mean?

00LT00:
just for the #603 and #583

nodece:
#583 has been rollback, so closing the PR.

Issue #614, opened 2020-09-30
[Question]Is it possible to ignore func (rm *RoleManager) PrintRoles() when i disable log


holayao:
------

**What's your scenario? What do you want to achieve?**
Your answer here

i have been  study with GC recently , and i found func (rm *RoleManager) PrintRoles() took up a lot of heap memory due to this line as shown in the red box 
![image](https://user-images.githubusercontent.com/41869842/94640989-a22bb480-0312-11eb-89b6-d787d90a615d.png)
so i wonder if this function can be ingored when i disable log to avoid large memory allocs

hsluoyz:
@nodece @00LT00

00LT00:
@holayao I noticed that you are still using the old version of casbin. So I suggest you should update the version for the many feature.
And you can how to turn off the log [here](https://casbin.org/docs/en/log-error#logging).
```go
e.EnableLog(false)
```

holayao:
I have already disable log feature,but this function will be executed when call buildRoleLinks, as > [here](https://github.com/casbin/casbin/blob/c730e2689d1f9e60206fbe30f0995d48899b6760/enforcer.go#L253)

it looks like we must enable function buildRoleLinks,or when i add or delete a role ,it wont work

hsluoyz:
We can wrap the log printing in `buildRoleLinks()` with an EnableLog flag check. @00LT00 or @holayao can you make a PR?

holayao:
> We can wrap the log printing in `buildRoleLinks()` with an EnableLog flag check. @00LT00 or @holayao can you make a PR?

holayao:
@hsluoyz of course,it's my pleasure

PR #615, opened 2020-10-02
fix: wrap the log printing in buildRoleLinks()

00LT00:

Fix: https://github.com/casbin/casbin/issues/614

### DIFF
--- a/model/assertion.go
+++ b/model/assertion.go
@@ -83,6 +83,8 @@ func (ast *Assertion) buildRoleLinks(rm rbac.RoleManager) error {
 		}
 	}
 
-	log.LogPrint("Role links for: " + ast.Key)
+	if log.GetLogger().IsEnabled() {
+		log.LogPrint("Role links for: " + ast.Key)
+	}
 	return ast.RM.PrintRoles()
 }
--- a/model/assertion.go
+++ b/model/assertion.go
@@ -83,8 +83,7 @@ func (ast *Assertion) buildRoleLinks(rm rbac.RoleManager) error {
 		}
 	}
 
-	if log.GetLogger().IsEnabled() {
-		log.LogPrint("Role links for: " + ast.Key)
-	}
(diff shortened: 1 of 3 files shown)


hsluoyz:
@holayao hi, @00LT00 already made a PR. Please review.

model/policy.go
```diff
@@ -43,6 +43,9 @@ func (model Model) BuildIncrementalRoleLinks(rm rbac.RoleManager, op PolicyOp, s
 
 // BuildRoleLinks initializes the roles in RBAC.
 func (model Model) BuildRoleLinks(rm rbac.RoleManager) error {
+	if log.GetLogger().IsEnabled() {
```
nodece, line 46: I think you should add `log.GetLogger().IsEnabled()` to `PrintPolicy()`.

model/assertion.go
```diff
@@ -83,6 +83,8 @@ func (ast *Assertion) buildRoleLinks(rm rbac.RoleManager) error {
 		}
 	}
 
-	log.LogPrint("Role links for: " + ast.Key)
+	if log.GetLogger().IsEnabled() {
```
nodece, line 86: unnecessary.

holayao:
@nodece I was on vacation the other day, and I'm sorry I didn't review the code. I dont think this PR fixs the problem i am facing . The line which takes up a large heap is in this function  as [here](https://github.com/casbin/casbin/blob/d0d65d828c784211a47c23f9cc63b7429da0287e/model/assertion.go#L88). And i seems like this fuction does nothing but print log. So is it possible to check log flag before execute this function?  Thanks

nodece:
@holayao We have added checks.
https://github.com/casbin/casbin/blob/d0d65d828c784211a47c23f9cc63b7429da0287e/rbac/default-role-manager/role_manager.go#L231

holayao:
@nodece  Sorry, i didn't notice that .Thank you so much!

PR #624, opened 2020-10-15
fix: Error in README

Fi5t:
A method `GetImplicitRolesForUser` has a such signature `func (e *Enforcer) GetImplicitRolesForUser(name string, domain ...string) []string `, and therefore returns only one value.


hsluoyz:
@Fi5t it has a 2nd error return value:

https://github.com/casbin/casbin/blob/eff1f230f9f635086ad5e1930926cb4d198ba5f9/rbac_api.go#L155-L189

hsluoyz:
Closed as invalid.

Issue #234, opened 2019-05-30
Dose casbin will add `Audit Log` ability ?


Wang-Kai:
As a ACL(or RBAC) system, tracking of authorization grants and denials is important. So dose casbin plan to add `Audit ability` to output access or denials log ? 

Just like [Ladon Audit Log](https://github.com/ory/ladon#audit-log-warden)

hsluoyz:
Hi @Wang-Kai , we will discuss about this feature. What Go logging library do you recommend to use?

vtolstov:
May be provide interface? For example I'm use for logs loki by grafana.

пт, 14 июн. 2019 г., 17:23 Yang Luo <notifications@github.com>:

Wang-Kai:
> Hi @Wang-Kai , we will discuss about this feature. What Go logging library do you recommend to use?

I recommend [apex/log](https://github.com/apex/log) strongly, it is simple and flexible.

mdjunior:
Hi guys,
This is a very important contribution.

The main auditLog operations should stay in [enforcer](https://github.com/casbin/casbin/blob/master/enforcer.go#L300), right? Looking at [line 432](https://github.com/casbin/casbin/blob/a5a1667e25abdbc5a4414ea9b6b889764174b7ec/enforcer.go#L432) `rvals[0]` has the name of the policy that matches for a true result.

Should we add an auditLog interface [here](https://github.com/casbin/casbin/blob/master/log/logger.go), right?

Could this interface get a `map[string]string` for the information to be structured?

hsluoyz:
Hi @Wang-Kai , Casbin has already provided a Logger interface, you can implement your own Logger with it: https://github.com/casbin/casbin/blob/master/log/logger.go

hsluoyz:
@mdjunior , we already have logging the enforce result:

https://github.com/casbin/casbin/blob/a5a1667e25abdbc5a4414ea9b6b889764174b7ec/enforcer.go#L430-L441

I don't quite get it by ``map[string]string``.

mdjunior:
This event can be a:
```
map[string]interface{
    "logType":"Request",
    "policy": rvals[0],
    "vals": fmt.Sprintf("%v",rvals[1:]),
    "result": result
}
```

In this way, an event can be transformed into a string or to other data structures. The user can manipulate this for your case.

Using the current `LogPrint` interface, the client would have to do the parser of the text.

obashistu:
[...]

PR #626, opened 2020-10-19
feat: Improve logger

kocoler:
Fix: https://github.com/casbin/casbin/issues/234

Improve the logger function.

### DIFF
--- a/enforcer.go
+++ b/enforcer.go
@@ -561,7 +561,12 @@ func (e *Enforcer) enforce(matcher string, explains *[]string, rvals ...interfac
 
 		}
 
-		log.LogPrint(reqStr.String())
+		logType := log.LogTypeGrantedAccessRequest
+		if !result {
+			logType = log.LogTypeRejectedAccessRequest
+		}
+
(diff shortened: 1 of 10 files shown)


hsluoyz:
@kocoler fix CI.

kocoler:
@hsluoyz already fixed.

hsluoyz:
@nodece @GopherJ @dovics please review.

nodece:
@kocoler I suggest that you add an EnforceLog interface to the log folder, you don't need to change the currently log, EnforceLog only records enforce log. 

@hsluoyz what do you think?

hsluoyz:
I think we should have a clear idea about why we are refactoring the logging.

From my opinion, we already have a logger interface, and an default logger that implements the logger. Next we should focus on how to implement other loggers that use the No.1, No.2 open-source Golang loggers. In such process, we may found places to improve inside our logger interface. Then we iterate to improve it. I think this is how the entire process works.

nodece:
@hsluoyz sounds great.  It is important that improve the currently log interface. A good idea that provides different level log.

Issue #613, opened 2020-09-29
[Feature] add the new UpdateGroupingPolicy() API


GopherJ:
**Want to prioritize this issue? Try:**

[![issuehunt-to-marktext](https://github.com/BoostIO/issuehunt-materials/raw/master/v1/issuehunt-button-v1.svg)](https://issuehunt.io/r/casbin/casbin)

------

**Is your feature request related to a problem? Please describe.**
A clear and concise description of what the problem is. Ex. I'm always frustrated when when I would like to rename a role and there isn't API for this operation yet.

**Describe the solution you'd like**

1. rename role in policy (can be in the first or the second placeholder)

```
p, admin, data1, read
g, alice, admin
```
becomes:
```
p, administrator, data1, read
g, alice, administrator
```

2. rename role in rbac manager for all domains

```
{ admin: Role { name: "admin", roles: [] }
```
becomes
```
{ administrator: Role { name: "administrator", roles: [] }
```

**Describe alternatives you've considered**

**Additional context**

liujiajun:
Hi all, I would like to work on this issue if no one's taking.

liujiajun:
Since this is my first time contributing to the project, I was wondering anyone could help check the proposal below:

### Proposal 
Add a new RBAC API `EditRoleName(oldName, newName, roleType string) (bool, error)`, where `roleType` can be either `resource` or `subject`.  `rowType` is needed because we need to know which field to update in policy rules. Alternatively, we can use something like `index` instead to achieve greater flexibility.

### Implementation
- Get filtered policies based on `oldName`. Save it to a temporary variable.
- Remove filtered policies based on `oldName`
- Update `oldName` to `newName` in the saved variable.
- Add policies using the saved variable

Admittedly, this is not the most efficient implementation, but it is easy to implement since all the above mentioned operations have APIs.

hsluoyz:
@GopherJ why don't we use `UpdatePolicy()` (https://github.com/casbin/casbin/pull/592) to update the roles in the g policy?

nodece:
It has been implemented by #592.

00LT00:
@nodece Plz reopen this issues, I just thought of the #592 couldn't resolve this issues. Because it just update `model[p]`.

hsluoyz:
@00LT00 we should have `Enforcer.UpdateGroupingPolicy()` series functions like add and remove.

Can you add this too?

00LT00:
I am writing, but something has been delayed in the past two weeks

[...]

PR #631, opened 2020-10-27
feat: Rename role

00LT00:
#613

### DIFF
--- a/management_api.go
+++ b/management_api.go
@@ -301,8 +301,11 @@ func (e *Enforcer) UpdateGroupingPolicy(oldName string, newName string) (bool, e
 }
 
 func (e *Enforcer) UpdateNamedGroupingPolicy(ptype string, oldName string, newName string) (bool, error) {
-	var err error
 	roles, err := e.GetRolesForUser(oldName)
+	if err != nil {
+		return false, err
+	}
+	names, err := e.GetUsersForRole(oldName)
 	if err != nil {
 		return false, err
 	}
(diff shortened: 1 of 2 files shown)


00LT00:
plz review @hsluoyz @nodece @GopherJ

management_api_test.go
```diff
@@ -253,4 +253,12 @@ func TestModifyGroupingPolicyAPI(t *testing.T) {
 	testGetUsers(t, e, []string{}, "data1_admin")
 	testGetUsers(t, e, []string{}, "data2_admin")
 	testGetUsers(t, e, []string{"eve"}, "data3_admin")
+	_, _ = e.AddGroupingPolicy("alice", "eve")
```
hsluoyz, line 256: @00LT00 `eve` is usually a user name, not a role name.

management_api.go
```diff
@@ -296,6 +296,33 @@ func (e *Enforcer) RemoveNamedGroupingPolicies(ptype string, rules [][]string) (
...
+	roles, err := e.GetRolesForUser(oldName)
+	if err != nil {
+		return false, err
+	}
+	names, err := e.GetUsersForRole(oldName)
+	if err != nil {
+		return false, err
+	}
+	for _, role := range roles {
+		if flag, err := e.updatePolicy("g", ptype, []string{oldName, role}, []string{newName, role}); err != nil || !flag {
```
GopherJ, line 313: it doesn't seems correct to me because here we are assuming: `g` only have two parameters but it's not always this case

00LT00: Do you mean we should add UpdateRoleInDomain? 
Or is it like UpdatePolicy?
```
func (e *Enforcer) UpdateNamedPolicy(ptype string, p1 []string, p2 []string) (bool, error) {
	return e.updatePolicy("p", ptype, p1, p2)
}
```

00LT00: I think the new logic could make sure this problem.

nodece (approved):
LGTM

nodece:
@00LT00 your commit message looks confusing.

management_api.go
hsluoyz: Any update?

Issue #639, opened 2020-11-16
function "LoadModel" is not synchronized in structure [SyncedEnforcer]


sangalao:
See content of file : 
https://github.com/casbin/casbin/blob/master/enforcer_synced.go

function "LoadModel" is not present, and thus not synchronized in this structure.
Therefore, thread concurrency issues may occur while calling [LoadModel] functions and other methods through different go routines

hsluoyz:
@sangalao can you make a PR to fix it?

hsluoyz:
@nodece

closetool:
I'll try to fix it, thanks

PR #669, opened 2020-12-21
fix: LoadModel concurrency

closetool:
The SyncedEnforcer is used on the condition of concurrency, so a lock is needed for preventing data race. We use a read-write lock for better performance. Cause LoadModel is the read action, so I called m.Lock() instead of RLock(), after read operation, release the lock. Model is used to parse the policy, so it is nessessary to call LoadPolicy after LoadModel.

Fix: https://github.com/casbin/casbin/issues/639


hsluoyz:
Use more words to describe the commit message and PR title.

hsluoyz:
@closetool plz fix:

image

hsluoyz:
@sangalao please review.

Issue #663, opened 2020-12-16
[Question]strSlice


maxiloEmmmm:
**Want to prioritize this issue? Try:**

[![issuehunt-to-marktext](https://github.com/BoostIO/issuehunt-materials/raw/master/v1/issuehunt-button-v1.svg)](https://issuehunt.io/r/casbin/casbin)

------

**What's your scenario? What do you want to achieve?**
Your answer here
version: github.com/casbin/casbin/v2 v2.17.0

```go
rule := []string{"role1", "obj", "act"}
e.AddPolicy(rule)
//that's ok
rule[0] = "role2"
e.AddPolicy(rule)
//adapter data ok
//e.model["p"]["p"].Policy is invaild, role1,obj,act not exist role2,obj,act exist
pass, _ := e.Enforce("role1", "obj", "act")
//pass is false
```
there will make role2 replace role1 for e.model["p"]["p"].Policy
https://github.com/casbin/casbin/blob/11b54bfb32cc2fff2bbb46fba1ffbbd26ec3cc34/management_api.go#L135-L138
Q1:
why not deal with here for `params[0].([]string)`
https://github.com/casbin/casbin/blob/11b54bfb32cc2fff2bbb46fba1ffbbd26ec3cc34/management_api.go#L139-L142

hsluoyz:
Closed for now.

maxiloEmmmm:
slice会造成引用问题你看不到么？
麻烦能好好看看问题？

maxiloEmmmm:
@hsluoyz  thk

hsluoyz:
Fixed in: https://github.com/casbin/casbin/pull/671

PR #671, opened 2020-12-23
fix: slice reference in AddNamePolicy

closetool:
AddNamePolicy uses the slice param passed in, when the param changing, and the policy will change too

Fix: https://github.com/casbin/casbin/issues/663

### DIFF
--- a/management_api.go
+++ b/management_api.go
@@ -134,6 +134,7 @@ func (e *Enforcer) AddPolicies(rules [][]string) (bool, error) {
 // Otherwise the function returns true by adding the new rule.
 func (e *Enforcer) AddNamedPolicy(ptype string, params ...interface{}) (bool, error) {
 	if strSlice, ok := params[0].([]string); len(params) == 1 && ok {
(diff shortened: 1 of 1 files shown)


management_api.go
```diff
@@ -134,6 +134,7 @@ func (e *Enforcer) AddPolicies(rules [][]string) (bool, error) {
 // Otherwise the function returns true by adding the new rule.
 func (e *Enforcer) AddNamedPolicy(ptype string, params ...interface{}) (bool, error) {
 	if strSlice, ok := params[0].([]string); len(params) == 1 && ok {
+		strSlice = append(make([]string,0,len(strSlice)),strSlice...)
```
hsluoyz, line 137: code format

management_api_test.go
```diff
@@ -184,18 +184,23 @@ func TestModifyPolicyAPI(t *testing.T) {
 	_, _ = e.RemoveNamedPolicy("p", namedPolicy)
 	_, _ = e.AddNamedPolicy("p", namedPolicy)
 
+	// for #663
```
hsluoyz, line 187: No need to add test

PR #672, opened 2020-12-28
fix: Make CachedEnforcer actually thread-safe

tchap:
Checked `CachedEnforcer`, seemed like it is supposed to be thread-safe since there is a mutex, but it isn't.

That should be now fixed.


hsluoyz:
lgtm

PR #724, opened 2021-03-07
fix: remove default role manager domain map

closetool:

### DIFF
--- a/rbac/default-role-manager/role_manager.go
+++ b/rbac/default-role-manager/role_manager.go
@@ -15,6 +15,7 @@
 package defaultrolemanager
 
 import (
+	"fmt"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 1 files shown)


closetool:
@hsluoyz plz review

rbac/default-role-manager/role_manager.go
```diff
@@ -24,13 +25,15 @@ import (
 	"github.com/casbin/casbin/v2/util"
 )
 
-const defaultDomain string = "casbin::default"
+const defaultDomain string = "casbin_default"
```
hsluoyz, line 28: Why `defaultDomain` is not empty string?

I think a global user alice should be ``::alice``

hsluoyz (changes requested):

hsluoyz:
@closetool can you post the comparison data of our benchmark test: https://casbin.org/docs/en/benchmark ?

hsluoyz (approved):

hsluoyz:
The performance diff:

image

Issue #728, opened 2021-03-09
Stronger UpdatePolicies() API


hsluoyz:
Here are some concerns about the API:

- https://github.com/casbin/casbin/issues/688#issuecomment-792598307
- https://github.com/casbin/casbin/issues/688#issuecomment-792716863

We should support:

1. The old policy size can be different with the new policy size.
2. The old policy supports "*" in its element to support a group of pattern-matched rules, like `GetFilteredPolicy()`

hsluoyz:
@closetool

hsluoyz:
@closetool merged, plz add the new API after: https://casbin.org/docs/en/management-api#updatepolicies

PR #729, opened 2021-03-09
feat: update policies enhance

closetool:

Fix: #728


dahu33:
@closetool, @hsluoyz, thanks a lot for working on this! While this is going in the right direction, it still relies on `SavePolicy()` which we are trying to avoid for its performance penalty.

Instead, I would propose to add an `UpdateFilteredPolicy(fieldIndex int, fieldValues[], newRules [][]string) error` function to the management API but also to the `UpdatableAdapter` interface so we can preserve atomicity between the model and the storage while using the `AutoSave` feature. Would that make sense?

closetool:
@dahu33 @arafat-java @hsluoyz plz review

internal_api.go
```diff
@@ -228,6 +236,24 @@ func (e *Enforcer) updatePolicies(sec string, ptype string, oldRules [][]string,
...
+	}
+
+	if ok, err := e.addPolicies(sec, ptype, newRules); err != nil {
+		return ok, err
+	}
+
+	e.EnableAutoSave(autoSave)
+	return true, nil
+}
+
```
dahu33, line 256: This implementation do not preserve atomicity. Example: if the call to `e.removePolicies` succeed but the call to `e.addPolicies` fails, we end-up with a partial update / an inconsistent state.

Also, to preserve atomicity between the in-memory policies and the policy storage (adapter), I don't see any other solution than extending the adapter interface (see my previous comment in this PR).

dahu33 (changes requested):

arafat-java:
I reviewed the pr and there are a couple of issues

1. It does not remove the policies on the object that are not passed in the new set. 
     E.g. if data1 previously had alice and bob and now in the UpdatePolicies I pass alice, jack and jill... it will add jack and jill but will not remove bob. So now the policies on data1 would be alice, bob, jack and jill which is incorrect
2. In the code I could see that the deletion and insertion are not part of the same transaction. Because in the code i see that at the beginning on line 241 it says e.EnableAutoSave(true) and I think that should be e.EnableAutoSave(false) for the entire think to happen within 1 transaction

internal_api.go
hsluoyz: @closetool

hsluoyz:
@closetool plz reply to all the questions in this PR. Make sure the concerns are resolved.

Issue #735, opened 2021-03-12
[Bug] Unable to add a domain matching function on default role manager


giefferre:
**Describe the bug**

The current implementation of the `defaultrolemanager` package makes impossible to set a custom domain matching function, even if the method `AddDomainMatchingFunc` is public.

**To Reproduce**

```go
package main

import (
	"log"

	"github.com/casbin/casbin/v2"
	defaultrolemanager "github.com/casbin/casbin/v2/rbac/default-role-manager"
)

func main() {
	rm := defaultrolemanager.NewRoleManager(10)

	e, err := casbin.NewEnforcer("model.conf", "policies.csv")
	if err != nil {
		log.Fatal(err)
	}

	rm.AddDomainMatchingFunc(
		"somename",
		func(pattern, domain string) bool {
			return false
		},
	)

	// implementation here...	
}
```

Results in
```
rm.AddDomainMatchingFunc undefined (type rbac.RoleManager has no field or method AddDomainMatchingFunc)compiler
```

**Expected behavior**

The method should be available on `defaultrolemanager.RoleManager` objects.

PR #736, opened 2021-03-12
fix: Unable to add domain matching func

giefferre:
This branch fixes casbin/casbin#735

The problem in the code was that the `defaultrolemanager.NewRoleManager` constructor method signature defines a `rbac.RoleManager` instead of a `*RoleManager`.

As side effect, this also involved the import of `rbac`, which is of no use in the `defaultrolemanager` package.

The important aspect to take care of, in fact, is that `defaultrolemanager.RoleManager` adheres to the `rbac.RoleManager`.
This is granted by the language itself.

This modification allows to avoid the type casting.


hsluoyz:
@closetool plz review.

closetool (approved):

Issue #744, opened 2021-03-29
Add GetDomainsForUser() API


hsluoyz:
See: https://github.com/casbin/Casbin.NET/issues/148#issuecomment-809174721

hsluoyz:
@closetool @RobotHuang

RobotHuang:
Does it return all domains one user has?

Sagilio:
See: https://github.com/casbin/Casbin.NET/issues/147#issue-841030349

PR #746, opened 2021-03-30
feat: Add GetDomainsForUser() API

RobotHuang:
Fix: https://github.com/casbin/casbin/issues/744

### DIFF
--- a/rbac/default-role-manager/role_manager.go
+++ b/rbac/default-role-manager/role_manager.go
@@ -300,6 +300,28 @@ func (rm *RoleManager) PrintRoles() error {
 	return nil
 }
 
+// GetDomains gets domains that a user has
+func (rm *RoleManager) GetDomains(name string) ([]string, error) {
+	var domains []string
+	for domain := range rm.domains {
+		if rm.hasAnyRole(name, domain)  {
+			domains = append(domains, domain)
+		}
+	}
+	domains = util.RemoveDuplicateElement(domains)
+	return domains, nil
+}
+
+func (rm *RoleManager) hasAnyRole(name string, domain string) bool {
+	patternDomain := rm.getPatternDomain(domain)
+	for _, domain := range patternDomain {
+		if rm.roles.hasRole(domain, name, rm.matchingFunc) {
+			return true
+		}
+	}
+	return false
+}
+
 // Roles represents all roles in a domain
 type Roles struct {
 	sync.Map
--- a/examples/rbac_with_domains_policy2.csv
+++ b/examples/rbac_with_domains_policy2.csv
@@ -1,0 +1,9 @@
+p, admin, domain1, data1, read
+p, admin, domain1, data1, write
+p, admin, domain2, data2, read
+p, admin, domain2, data2, write
+p, user, domain3, data2, read
+g, alice, admin, domain1
+g, alice, admin, domain2
+g, bob, admin, domain2
+g, bob, user, domain3
--- a/model_test.go
+++ b/model_test.go
@@ -347,6 +347,9 @@ func (rm *testCustomRoleManager) GetRoles(name string, domain ...string) ([]stri
 func (rm *testCustomRoleManager) GetUsers(name string, domain ...string) ([]string, error) {
 	return []string{}, nil
 }
+func (rm *testCustomRoleManager) GetDomains(name string) ([]string, error) {
+	return []string{}, nil
+}
 func (rm *testCustomRoleManager) PrintRoles() error           { return nil }
 func (rm *testCustomRoleManager) SetLogger(logger log.Logger) {}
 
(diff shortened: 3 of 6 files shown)


hsluoyz:
@thoraj please review.

hsluoyz:
@thoraj can you just see if the test case output is expected for the model and policy? If that's good to go, we will port the exactly the same logic to .NET

image

rbac/default-role-manager/role_manager.go
```diff
@@ -300,6 +300,22 @@ func (rm *RoleManager) PrintRoles() error {
 	return nil
 }
 
+// GetDomains gets domains that a user has
+func (rm *RoleManager) GetDomains(name string) ([]string, error) {
+	var domains []string
+	for domain := range rm.domains {
+		roles, err := rm.GetRoles(name, domain)
```
Sagilio, line 307: This API need not know the role list, I think to use an API only return bool like `hasAnyRole` is better.

hsluoyz: @Sagilio but do we have such an API? `hasAnyRole()`

hsluoyz: @RobotHuang

thoraj: I had a look, and I also tested the same model and policy using my rather clumsy workaround for getting the domains. I got the same result as shown in your tests. 

But since the information can he extracted from grouping policy alone, the tests will not cover cases where:
- The domain association is directly given by a policy
- The domain association is indirect and follows from the grouping policy (rbac)

E.g. what would the result be if we added a policy: 
```
p, carol, domain3, data2, read
```
And asked for Carol's roles?  (I discovered my workaround failed to produce: carol: domain3) 

So a question then is should the domains only be derived from grouping policies, or also from direct policies?

Sagilio: I mean we can add a new func like this, it can pass the test:
```go
// GetDomains gets domains that a user has
func (rm *RoleManager) GetDomains(name string) ([]string, error) {
	var domains []string
	for domain := range rm.domains {
		if rm.hasAnyRole(name, domain)  {
			domains = append(domains, domain)
		}
	}
	domains = util.RemoveDuplicateElement(domains)
	return domains, nil
}

func (rm *RoleManager) hasAnyRole(name string, domain string) bool {
	patternDomain := rm.getPatternDomain(domain)
	for _, domain := range patternDomain {
		if rm.roles.hasRole(domain, name, rm.matchingFunc) {
			return true
		}
	}
	return false
}
```

hsluoyz: @RobotHuang @Sagilio can you answer the question proposed by @thoraj in this thread?

RobotHuang: @thoraj The result of `p, carol, domain3, data2, read`  is `domain3` when we ask the domains of carol and we have tested it. Could u provide some cases about `The domain association is indirect and follows from the grouping policy (rbac)`?

hsluoyz: @thoraj

thoraj: >Could u provide some cases about The domain association is indirect and follows from the grouping policy (rbac)?

```
g, alice, role1
g, role1, role2

p, role2, dom1, obj1, act1
```

Would GetDomains("alice") => dom1 ?

hsluoyz: @RobotHuang

RobotHuang: @thoraj In docs, the role_definition should be `g = _, _, _`. You could refer https://casbin.org/docs/en/rbac-with-domains.

thoraj: Hmm. OK. For rbac_with_domains that makes sense I guess. 

But I'm experimenting with a model.
```
[request_definition]
r = sub, dom, obj, act

[policy_definition]
p = sub, dom, obj, act

[role_definition]
g = _, _, _  	# users and role hierarcies
g2 = _, _		# organization hierarchies
g3 = _, _		# functional hierarchies

[policy_effect]
e = some(where (p.eft == allow))

[matchers]
m = g(r.sub, p.sub, r.dom) && g2(r.obj, p.obj) && g3(r.act, p.act) && r.dom == p.dom
```

Here only user and role hierarchy `"g"` uses domains. The org and functional hierarchies do not. 

Is it OK to mix like this? I would hope that as long as permission policies use the `"g"` hierarchy (which do have domain), this would be OK (in the sense that GetDomains() will return the right values)

hsluoyz: @thoraj it's OK. Different RBAC systems like g, g2, g3 are totally separated. Any one can have domain or not.

thoraj: Ok. Thanks. Good to know. 

I agree that my example of indirect domain attribution is contrived since roles must have domains for a "**rbac with domain**" model. 

So I think the suggested implementation would work well
Looking forward to testing this in the casbin.net version  👍

hsluoyz:
@RobotHuang please add the new APIs to the official docs

Issue #748, opened 2021-04-02
[Bug] Unable to Enforce policies after LoadPolicy


giefferre:
**Describe the bug**

After the execution of `e.LoadPolicy()` method where `e` is either a Casbin Enforcer or a Casbin Synced Enforcer, the previously Enforced policies are not enforced anymore.

**To Reproduce**

```go
func main() {
	casbinEnforcer, err := casbin.NewEnforcer("model.conf", "policies.csv")
	if err != nil {
		log.Fatal(err)
	}

	allowed, err := casbinEnforcer.Enforce("mySubject", "myObject", "anAction")
	if err != nil {
		log.Fatal(err)
	}
	log.Println(allowed) // true
	
	casbinEnforcer.LoadPolicy()

	allowed, err := casbinEnforcer.Enforce("mySubject", "myObject", "anAction")
	if err != nil {
		log.Fatal(err)
	}
	log.Println(allowed) // false <---- THIS SHOULD BE TRUE AS PREVIOUS EXECUTION
}
```

**Expected behavior**

The Enforce is expected to return the same result even reloading the same policies

**Additional context**

This replicates on different adapters.

hsluoyz:
@giefferre can you provide:

1. go env
2. go version
3. Casbin version
4. contents of "model.conf", "policies.csv"

giefferre:
Complete example follows

```go
package main

import (
	"log"

	"github.com/casbin/casbin/v2"
	"github.com/casbin/casbin/v2/util"
)

func main() {
	casbinEnforcer, err := casbin.NewEnforcer("model.conf", "policies.csv")
	if err != nil {
		log.Fatal(err)
	}

	casbinEnforcer.AddNamedDomainMatchingFunc("g", "", util.KeyMatch2)
	casbinEnforcer.AddNamedMatchingFunc("g", "", util.KeyMatch2)

	allowed, err := casbinEnforcer.Enforce("alice", "office/service", "write")
	if err != nil {
		log.Fatal(err)
	}
	log.Println(allowed) // true

	casbinEnforcer.LoadPolicy()

	allowed, err = casbinEnforcer.Enforce("alice", "office/service", "write")
	if err != nil {
		log.Fatal(err)
	}
	log.Println(allowed) // false <---- THIS SHOULD BE TRUE AS PREVIOUS EXECUTION
}
```

ACTUAL RESULT
```
2021/04/02 15:07:58 true
2021/04/02 15:07:58 false
```

model.conf

```
[...]

PR #749, opened 2021-04-03
fix: enforcer action changed after calling `LoadPolicy`(#748)

closetool:

fix: #748

### DIFF
--- a/enforcer.go
+++ b/enforcer.go
@@ -280,7 +280,9 @@ func (e *Enforcer) LoadPolicy() error {
 		return err
 	}
 
-	e.initRmMap()
+	if err := e.clearRmMap(); err != nil {
+		return err
+	}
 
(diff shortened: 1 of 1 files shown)


closetool:
@hsluoyz @giefferre plz reiview

hsluoyz:
@closetool can you explain what caused this bug and why made such change?

closetool:
@hsluoyz sure
In previous code, calling `LoadPolicy` resulted in replacement of rolemanger, `DomainMatchingFunc` and `MatchingFunc` will be deleted too, so we have to reset matching function.
Now, I just call the `Clear` method, and matching function will be reversed.

hsluoyz:
@closetool great, just as I expected in: https://github.com/casbin/casbin/issues/748#issuecomment-812523355

Issue #742, opened 2021-03-25
[Feature] Migrating from `Knetic/govaluate` to `antonmedv/expr`


fr3fou:
**Is your feature request related to a problem? Please describe.**
The Go implementation of Casbin uses https://github.com/Knetic/govaluate, which at this point seems to be abandoned. Last commit was made on Oct 22, 2017 and the files haven't been updated in 4-5 years 
![image](https://user-images.githubusercontent.com/1344906/112514065-9ab70680-8d9d-11eb-9129-dd48f2993316.png)
There are as well some long-standing bugs that are annoying to deal with, such as 

![image](https://user-images.githubusercontent.com/1344906/112514400-f1bcdb80-8d9d-11eb-8e99-63932f33cfd5.png)
https://github.com/Knetic/govaluate/issues/79

**Describe the solution you'd like**
Casbin should migrate to a more maintained library, such as https://github.com/antonmedv/expr

**Describe alternatives you've considered**
 Forking https://github.com/Knetic/govaluate and fixing the bugs seems like an option, but it might be easier to just migrate.

hsluoyz:
@fr3fou what about https://github.com/PaesslerAG/gval ?

fr3fou:
> @fr3fou what about https://github.com/PaesslerAG/gval ?

also seems fine, we should pick whichever will take the least time to migrate to

tson1111:
Hi, I want to have a try on this issue! I'll look at https://github.com/PaesslerAG/gval

hsluoyz:
@tson1111 https://github.com/PaesslerAG/gval is just an option. Please choose the best one from all options.

tson1111:
@hsluoyz Cool. I’ll do some research.

tson1111:
@hsluoyz Hi, I took a look at some evaluation expression packages. It turns out that [expr](https://github.com/antonmedv/expr) has better performance (comparison [here](https://github.com/antonmedv/golang-expression-evaluation-comparison)) while [gval](https://github.com/PaesslerAG/gval) is better-documented and seems easier to migrate to. Since performance isn't our major concern and [gval](https://github.com/PaesslerAG/gval) is completely fine (https://github.com/PaesslerAG/gval#performance), [gval](https://github.com/PaesslerAG/gval) seems like a better choice. If it looks go to you, I'll start to code.

hsluoyz:
[...]

PR #758, opened 2021-04-11
feat: Migrate expression evaluation library from `govaluate` to `expr`.

tson1111:
As issue #742 mentioned, we want a more actively-maintained and efficient expression evaluation library. [expr](https://github.com/antonmedv/expr) may be a better choice compared to the current library [govaluate](https://github.com/Knetic/govaluate).


hsluoyz:
@tson1111 

1. Is this a breaking change?
2. Can you evaluate the effect of the breaking change?

PR #766, opened 2021-04-19
try to fix #762

### LINKED ISSUE #762: [Bug]Using priority-model Won't get any implicit permission
**Describe the bug**
using priority model, and has the policy as follow:
p, 10, data_group, data, read, allow

when u try to get implicit permission with GetImplicitPermissionsForUser("data_group")
it will return nil

for the function GetPermissionsForUser, it write e.GetFilteredPolicy(0, args...)
but index=0 is the priority field.

will it be a bug, or if i use the wrong method to get a user's permission?

yockii:
try to find priority index when adding policy and sorting priority based policy


hsluoyz:
@yockii plz fix:

image

hsluoyz:
@closetool plz review.

model/policy.go
```diff
@@ -147,24 +147,35 @@ func (model Model) HasPolicies(sec string, ptype string, rules [][]string) bool
...
-			if err != nil {
-				break
-			}
-			if idx > idxInsert {
-				assertion.Policy[i] = assertion.Policy[i-1]
-			} else {
-				break
+
+	// find priority index
+	priorityIndex := -1
```
hsluoyz, line 152: Can we refactor this code into a function? like `getPriorityIndex()`?

```go
priorityIndex := -1
	for index, token := range assertion.Tokens {
		if token == fmt.Sprintf("%s_priority", ptype) {
			priorityIndex = index
			break
		}
	}
```

model/policy.go
```diff
@@ -147,24 +147,35 @@ func (model Model) HasPolicies(sec string, ptype string, rules [][]string) bool
...
+	for index, token := range assertion.Tokens {
+		if token == fmt.Sprintf("%s_priority", ptype) {
+			priorityIndex = index
+			break
+		}
+	}
+
+	if priorityIndex >= 0 {
+		if idxInsert, err := strconv.ParseUint(rule[priorityIndex], 10, 32); sec == "p" && err == nil {
```
hsluoyz, line 161: Define a util function `pasrseUint()` to contain `strconv.ParseUint(rule[priorityIndex], 10, 32)`

hsluoyz, line 161: Why not put `sec == "p"` at first? So we can early return, because only p rules can have priority.

hsluoyz (changes requested):

model/policy.go
closetool: @hsluoyz @yockii Can we define a global variable `priorityIndex` which has default value `-1`, and it'll be initialled in `SortPoliciesByPriority` ?

model/policy.go
closetool: Maybe `strings.Atoi` could replace it

closetool (changes requested):
@yockii plz fix above

yockii:
> @yockii plz fix above

got it.

yockii:
> @yockii plz fix:
> 
> image

how to fix this ? as i have updated the pr title and commit comment begined with 'fix: ', but still got a failed semantic pull request?

hsluoyz:
@yockii this is not `fix: xxx`:

image

model/policy.go
```diff
@@ -147,32 +146,34 @@ func (model Model) HasPolicies(sec string, ptype string, rules [][]string) bool
...
-			if err != nil {
-				break
-			}
-			if idx > idxInsert {
-				assertion.Policy[i] = assertion.Policy[i-1]
-			} else {
-				break
+
+	if sec == "p" && assertion.priorityIndex >= 0 {
+		if idxInsert, err := strconv.Atoi(rule[assertion.priorityIndex]); sec == "p" && err == nil {
```
hsluoyz, line 151: no `sec == "p"` again

hsluoyz (changes requested):

hsluoyz (approved):

Issue #768, opened 2021-04-21
[Bug] several: invalid memory address or nil pointer dereference


arigon:
**Describe the bug**
There are several nil pointer dereferences which are not properly checked. I will create some PRs linked to this issue, but it will take some time to find most of them. :-)

**To Reproduce**
For example: Trying to add a grouping policy without any existing policy.

**Additional context**
`_, ok := model[sec][ptype].PolicyMap[strings.Join(rule, DefaultSep)]`
If model[sec] has no hit, it will panic. If model[sec][ptype] has no hit it will panic as well.

hsluoyz:
Won't fix due to: https://github.com/casbin/casbin/pull/769#issuecomment-826354184

PR #769, opened 2021-04-21
added assertions to prevent nil pointer exceptions

arigon:
Fix: https://github.com/casbin/casbin/issues/768

### DIFF
--- a/model/policy.go
+++ b/model/policy.go
@@ -36,7 +36,11 @@ const DefaultSep = ","
 // BuildIncrementalRoleLinks provides incremental build the role inheritance relations.
 func (model Model) BuildIncrementalRoleLinks(rmMap map[string]rbac.RoleManager, op PolicyOp, sec string, ptype string, rules [][]string) error {
 	if sec == "g" {
-		return model[sec][ptype].buildIncrementalRoleLinks(rmMap[ptype], op, rules)
+		if _, ok := model[sec]; ok {
+			if _, ok := model[sec][ptype]; ok {
+				return model[sec][ptype].buildIncrementalRoleLinks(rmMap[ptype], op, rules)
+			}
+		}
 	}
 	return nil
 }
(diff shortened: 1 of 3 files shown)


model/policy.go
```diff
@@ -144,28 +163,31 @@ func (model Model) HasPolicies(sec string, ptype string, rules [][]string) bool
...
-			for ; i > 0; i-- {
-				idx, err := strconv.Atoi(assertion.Policy[i-1][assertion.priorityIndex])
-				if err != nil {
-					break
-				}
-				if idx > idxInsert {
-					assertion.Policy[i] = assertion.Policy[i-1]
-				} else {
-					break
+	if _, ok := model[sec]; ok {
```
hsluoyz, line 166: @arigon I'm not sure it's good to silently mute the non-existing keys, because we don't return error in this API and the caller has no mean to know that some sec or ptype doesn't work.

The caller has the reponsibility to make sure the input sec and ptype are working in his scenario (model). Or we will generate Go panic to alert him.

Or do you have a scenario for this change?

hsluoyz (changes requested):

hsluoyz:
@arigon plz fix: 
image

link your github:

image

arigon:
> @arigon plz fix:
> image
> 
> link your github:
> 
> image

How can I fix this?

model/policy.go
arigon: The problem is in fact that the API is not returning an error. Generating a panic is against best practice.

See https://golang.org/doc/effective_go#panic 
> This is only an example but real library functions should avoid panic. If the problem can be masked or worked around, it's always better to let things continue to run rather than taking down the whole program.

So this should only be a temporary solution to prevent panics at all when no policy is loaded. The final solution should include returning an error but it might be a breaking change.

hsluoyz:
The caller is responsible to make the code correct and avoid panic. Casbin doesn't do endless and tedious checking against user input. And this PR makes the code harder to read with deeper if. So closed here.

Issue #775, opened 2021-04-27
[Question] WatcherEx methods should have ptype parameter.


gonsuke:
Hi team,

I'm looking into WatcherEx mechanisms and the [redis-watcher](https://github.com/casbin/redis-watcher) implementation.
As far as I understand so far, there are 2 ways of trigger of `UpdateForAddPolicy` for example.
One is `Enforcer.AddNamedPolicy`, https://github.com/casbin/casbin/blob/ab6b8e7e136ff05b251abc374e7be38fe9c80e13/management_api.go#L145   and the other is `Enforcer.AddNamedGroupingPolicy`. https://github.com/casbin/casbin/blob/ab6b8e7e136ff05b251abc374e7be38fe9c80e13/management_api.go#L259

Now I'm writing `UpdateForAddPolicy` and other WatherEx event handlers on my application, and I'm trying to call [LoadPolicyLine](https://github.com/casbin/casbin/blob/ab6b8e7e136ff05b251abc374e7be38fe9c80e13/persist/adapter.go#L25-L44) so that the application can load only policy which is delivered from WatcherEx without querying to backend storage.

However, WatcherEx methods don't have `ptype` parameter, so it's unable to construct policy line for `LoadPolicyLine`.
Any thoughts? Should we have `ptype` parameter for each WatcherEx methods?

hsluoyz:
@closetool

PR #776, opened 2021-04-27
feat: add `ptype` parameter for WatcherEx methods

gonsuke:
Fix: https://github.com/casbin/casbin/issues/775

As I mentioned in the issue
https://github.com/casbin/casbin/issues/775
we'd like to have `ptype` parameter in WatcherEx methods. 

This pr is a proposal, and aparently this will break exiting WatcherEx implementation though...

### DIFF
--- a/internal_api.go
+++ b/internal_api.go
@@ -58,7 +58,7 @@ func (e *Enforcer) addPolicy(sec string, ptype string, rule []string) (bool, err
 	if e.watcher != nil && e.autoNotifyWatcher {
 		var err error
 		if watcher, ok := e.watcher.(persist.WatcherEx); ok {
-			err = watcher.UpdateForAddPolicy(rule...)
+			err = watcher.UpdateForAddPolicy(ptype, rule...)
 		} else {
 			err = e.watcher.Update()
 		}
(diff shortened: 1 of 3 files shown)


closetool:
@gonsuke plz fix this
image

internal_api.go
```diff
@@ -58,7 +58,7 @@ func (e *Enforcer) addPolicy(sec string, ptype string, rule []string) (bool, err
 	if e.watcher != nil && e.autoNotifyWatcher {
 		var err error
 		if watcher, ok := e.watcher.(persist.WatcherEx); ok {
-			err = watcher.UpdateForAddPolicy(rule...)
```
closetool, line 61: @gonsuke I suggest adding `sec` param either.

closetool (changes requested):
plz tell me your opinions

closetool:
@hsluoyz plz review

internal_api.go
gonsuke: @closetool Thanks for the review.
Sure thing, added `sec` param in https://github.com/casbin/casbin/pull/776/commits/3452597d96873b48e5a9073732602b4103b72221

gonsuke:
> @gonsuke plz fix this
> image

Ah, sorry for that. Just rebased.

gonsuke:
Hello, any update on this? Without this, I don't think we can update only changed policy. Therefore we have to reload all policies every single time.
Any advice, comments, feedback are much appreciated.

internal_api.go
hsluoyz: @closetool why adding sec? Does a rule have sec? I think only model file has sec.

closetool: @hsluoyz for other interface has `sec`, like https://github.com/casbin/casbin/blob/05b22efb450e2e58dd21d31f502261cf46d80f51/persist/adapter.go#L55,
should we keep consistent?

hsluoyz: OK.

hsluoyz:
lgtm

PR #806, opened 2021-05-30
fix: parallel multiple go SyncedEnforcer.StartAutoLoadPolicy(...

sgoroshko:
Two or more goroutines at some time, can start across `e.IsAutoLoadingRunning`

### DIFF
--- a/enforcer_synced.go
+++ b/enforcer_synced.go
@@ -53,10 +53,10 @@ func (e *SyncedEnforcer) IsAutoLoadingRunning() bool {
 // StartAutoLoadPolicy starts a go routine that will every specified duration call LoadPolicy
 func (e *SyncedEnforcer) StartAutoLoadPolicy(d time.Duration) {
 	// Don't start another goroutine if there is already one running
-	if e.IsAutoLoadingRunning() {
+	if !atomic.CompareAndSwapInt32(&e.autoLoadRunning, 0, 1) {
 		return
 	}
-	atomic.StoreInt32(&(e.autoLoadRunning), int32(1))
+
 	ticker := time.NewTicker(d)
 	go func() {
 		defer func() {


hsluoyz:
@closetool @Sagilio plz review.

enforcer_synced.go
```diff
@@ -53,10 +53,10 @@ func (e *SyncedEnforcer) IsAutoLoadingRunning() bool {
 // StartAutoLoadPolicy starts a go routine that will every specified duration call LoadPolicy
 func (e *SyncedEnforcer) StartAutoLoadPolicy(d time.Duration) {
 	// Don't start another goroutine if there is already one running
-	if e.IsAutoLoadingRunning() {
+	if !atomic.CompareAndSwapInt32(&e.autoLoadRunning, 0, 1) {
```
closetool, line 56: If the e.IsAutoLoadingRunning method is not safe, we should fix it

closetool (changes requested):
plz consider this

enforcer_synced.go
sgoroshko: method is safe.
but mechanism that started goroutine, consist of two parts:
 - check if goroutine already started
 - set flag `autoLoadRunning`

two or more coroutines can execute this steps at some time for each step

closetool: @sgoroshko Ok

hsluoyz: @sgoroshko

Why not directly change the function body of `IsAutoLoadingRunning()`?

nodece: @hsluoyz `IsAutoLoadingRunning()` just check value, `atomic.CompareAndSwapInt32` can update the `e. autoLoadRunning` value.

nodece (approved):
LGTM

enforcer_synced.go
hsluoyz: OK

PR #815, opened 2021-06-11
fix: remove index in DeleteDomains method

### LINKED ISSUE #814: The `index` parameter is not used by the `DeleteDomains(...)` method
It seems that the `DeleteDomains(index int, domains ...string)` method does not use the `index` parameter anymore. Could it be removed?

https://github.com/casbin/casbin/blob/bdee23aafb01ac406e1efc9265f94528a840699d/rbac_api_with_domains.go#L121

tangyang9464:
Signed-off-by: tangyang9464 tangyang9464@163.com
fix #814


Sagilio (commented):
Thanks! Please fix: 
image
You need add `fix:` in commit message and with `-s` option.

tangyang9464:
@Sagilio fixed

Issue #825, opened 2021-06-29
[Feature Request] filter in FilteredAdapter can not work with multipie role manager 


Zxilly:
https://github.com/casbin/casbin/blob/77c6eb2379abd3e68fee339c62ffca532c65b8c2/persist/file-adapter/adapter_filtered.go#L36-L39

It cannot be used to filter `g2,g3...` policies.

hsluoyz:
@closetool @tangyang9464

PR #826, opened 2021-06-30
fix: filter not support mutiple ptype

closetool:

Fix: #825

### DIFF
--- a/filter_test.go
+++ b/filter_test.go
@@ -18,6 +18,7 @@ import (
 	"testing"
 
 	fileadapter "github.com/casbin/casbin/v2/persist/file-adapter"
+	"github.com/casbin/casbin/v2/util"
 )
 
 func TestInitFilteredAdapter(t *testing.T) {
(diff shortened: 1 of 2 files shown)


closetool:
@hsluoyz @Zxilly

Zxilly (approved):
lgtm

persist/file-adapter/adapter_filtered.go
```diff
@@ -34,8 +34,7 @@ type FilteredAdapter struct {
 // Filter defines the filtering rules for a FilteredAdapter's policy. Empty values
 // are ignored, but all others must match the filter.
 type Filter struct {
-	P []string
-	G []string
+	PG map[string][][]string
```
hsluoyz, line 37: This is a breaking change. At least we should use "feature:" instead of "fix:"

filter_test.go
```diff
@@ -15,6 +15,7 @@
 package casbin
 
 import (
+	"github.com/casbin/casbin/v2/util"
```
hsluoyz, line 18: sort

filter_test.go
```diff
@@ -44,8 +45,10 @@ func TestLoadFilteredPolicy(t *testing.T) {
 	testHasPolicy(t, e, []string{"admin", "domain2", "data2", "read"}, true)
 
 	if err := e.LoadFilteredPolicy(&fileadapter.Filter{
-		P: []string{"", "domain1"},
-		G: []string{"", "", "domain1"},
+		PG: map[string][][]string{
```
hsluoyz, line 48: A way to keep compatibility:

```go
	if err := e.LoadFilteredPolicy(&fileadapter.Filter{
		P: []string{"", "domain1"},
		G: []string{"", "", "domain1"},
		G2: []string{"", "", "domain1"},
		G3: []string{"", "", "domain1"},
	}); err != nil {
		t.Errorf("unexpected error in LoadFilteredPolicy: %v", err)
	}
```

I think we can hard-code it to G4 or G5. It's enough to use.

hsluoyz (changes requested):

persist/file-adapter/adapter_filtered.go
hsluoyz: For: https://github.com/casbin/casbin/pull/826#discussion_r661350076, we can use fix:

hsluoyz (approved):

Issue #831, opened 2021-07-03
[Question] Prioritizing based on role and user hierarchy level


merajsahebdar:
**What's your scenario? What do you want to achieve?**
As I read in documents, by default Casbin designed to prioritize rules by given definition order (1); and also there is a way to add priority in rules definitions (2). But what I need in my app is to be able to prioritize by roles/users levels; and as levels could be infinite, I can't use the (2) approach, cause in every change I need to change all priority levels.
Is there a way to prioritze based on hierarchy level currently? If no, is it possible to implement a feature for this?

**Your model:**

```ini
[request_definition]
r = sub, obj, act

[policy_definition]
p = sub, obj, act, eft

[role_definition]
g = _, _

[policy_effect]
e = priority(p.eft) || deny

[matchers]
m = g(r.sub, p.sub) && r.obj == p.obj && r.act == p.act
```

**Your policy:**

```
p, admin, posts, create, allow
p, admin, posts, remove, allow

p, editor, posts, create, allow
p, editor, posts, remove, deny

p, subscriber, posts, create, deny
p, subscriber, posts, remove, deny

p, jane, posts, remove, allow

g, jane, editor
g, john, subscriber
```

**Your request(s):**

```
jane, posts, create ---> true
jane, posts, remove ---> false (expected: true)

john, posts, create ---> false
john, posts, remove ---> false
```

hsluoyz:
@merajsahebdar can you give a concrete example for this?

> But what I need in my app is to be able to prioritize by roles/users levels

hsluoyz:
@closetool @tangyang9464

tangyang9464:
@merajsahebdar In your example, just change the priority can solve it. Do you have any other specific examples？

closetool:
@merajsahebdar 
`p, editor, posts, remove, deny` and `g, jane, editor` denied your request `jane, posts, remove`

merajsahebdar:
@hsluoyz @tangyang9464 @closetool 

As I've described in my question, this is the structure we have:
```
├─ role: editor
│  ├─ user: ...
│  ├─ user: ...
│  └─ user: jane
│
└─ role: subscriber
   ├─ user: ...
   ├─ user: ...
   └─ user: john
```

However, I know we can easily pass the priority in policy definitions like this:
```
p, 20, editor, posts, remove, deny
p, 10, jane, posts, remove, allow
```
But in cases that we have more than one level like this:
```
role: root
 └─ role: admin
    ├─ role editor
[...]

PR #833, opened 2021-07-06
feat: add prioritizing based on role and user hierarchy

tangyang9464:
Fix: #831
@hsluoyz @closetool plz review
@merajsahebdar plz just change ``priority(p.eft) || deny`` -> ``subjectPriority(p.eft) || deny``  and test your case.

### DIFF
--- a/examples/subject_priority_policy.csv
+++ b/examples/subject_priority_policy.csv
@@ -1,0 +1,16 @@
+p, root, data1, read, deny
+p, admin, data1, read, deny
+
+p, editor, data1, read, deny
+p, subscriber, data1, deny
+
+p, jane, data1, read, allow
+p, alice, data1, read, allow
+
+g, admin, root
+
+g, editor, admin
+g, subscriber, admin
+
+g, jane, editor
+g, alice, subscriber
--- a/model/model.go
+++ b/model/model.go
@@ -15,6 +15,8 @@
 package model
 
 import (
+	"container/list"
+	"errors"
 	"fmt"
 	"regexp"
 	"sort"
@@ -31,6 +33,9 @@ type Model map[string]AssertionMap
 
 // AssertionMap is the collection of assertions, can be "r", "p", "g", "e", "m".
 type AssertionMap map[string]*Assertion
+
+const defaultDomain string = ""
+const defaultSeparator = "::"
 
 var sectionNameMap = map[string]string{
 	"r": "request_definition",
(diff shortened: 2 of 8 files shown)


hsluoyz:
@closetool plz review.

tangyang9464:
For this feature, the inherited structure of roles and users can only be multiple trees, not graphs. One user cannot correspond to multiple roles.such as ``g, jane,editor`` ``g, jane,subscriberr`` ,the ``jane`` have two roles.This situation does not guarantee the correct result

hsluoyz:
@merajsahebdar plz review.

merajsahebdar:
@tangyang9464 
I tried a lot of scenarios, It's working great.
Thanks for your time and effort! 🙏🏻 Now Casbin becomes more lovely to me!

Just to make it a little more clear about what you mentioned, it works even  in a scenario like this:
```
p, root, data1, write, allow
p, admin, data1, write, deny
p, editor, data1, write, deny

p, user, data1, read, allow
p, subscriber, data1, read, deny

g, admin, root
g, editor, admin
g, subscriber, user

g, jane, root
g, jane, user
```
but does not work in:
```
p, root, data1, write, allow
p, admin, data1, write, deny
p, editor, data1, write, deny

p, user, data1, read, allow
p, user, data1, write, allow # trying to rewrite policies in tree: root
p, subscriber, data1, read, deny

g, admin, root
g, editor, admin
g, subscriber, user

g, jane, root
g, jane, user
``` 
so, users can have multiple roles, but:
1. those roles could not be in the same tree. (logically it's weird to add a user in two different branches of a tree, so might not happen!)
2. you could not define the same policy for both of those different types of roles.

However, I don't think these happen in a well-designed logic, but it's necessary  to make it clear in docs.

Thanks again!

tangyang9464:
@merajsahebdar
Thanks for your advice！But I want to explain something.
1. In your example,``jane`` has the same hierarchy(level) in both trees,so the plolicy ranking will not be affected
2. But Consider the following example

model
```
p, user, data1, read, allow
p, subscriber, data1, read, deny

g, admin1, root1
g, admin2, root2

g, jane,admin1
g, jane,root2 # jane has two roles and the levels in two tree are different

g, alice,admin2
```
the corresponding tree structure
image
the level:

- ``jane``  in `root-1` tree is 3
- But ``jane`` in  ``root-2`` tree is 2

so sometimes ``admin-1`` is 2,but the ``jane`` is 2 also.it's wrong(each role or user can onle has a level.If has two,will overwrite each other)

if one user has multiple roles,You have to make sure the user has the same level in different trees and in your second point
``you could not define the same policy for both of those different types of roles.``
In this case,the order for ``root`` and ``user`` is the default（from top to bottom）because they level are same

I will modify the code so that can report errors when using incorrectly(such as one user have two level)

examples/subject_priority_policy.csv
```diff
@@ -0,0 +1,16 @@
...
+p, jane, data1, read, allow
+p, alice, data1, read, allow
+
+g, admin, root
+
+g, editor,admin
+g, subscriber,admin
+
+g, jane,editor
+g, alice,subscriber
```
closetool, line 16: Add blank after comma.

model/model.go
```diff
@@ -206,6 +208,89 @@ func (model Model) PrintModel() {
...
+	for ptype, assertion := range model["p"] {
+		for index, token := range assertion.Tokens {
+			if token == fmt.Sprintf("%s_sub", ptype) {
+				subIndex = index
+				break
+			}
+		}
+		if subIndex == -1 {
+			continue
+		}
```
closetool, line 225: Can we assume `sub` always to be first, like `SortPoliciesByPriority`

model/model.go
```diff
@@ -206,6 +208,89 @@ func (model Model) PrintModel() {
...
+		policies := assertion.Policy
+		subjectHierarchyMap, err := getSubjectHierarchyMap(model)
```
closetool, line 227: What if
```
alice -> bob
bob -> tom
tom -> alice
```

model/model.go
```diff
@@ -206,6 +208,89 @@ func (model Model) PrintModel() {
...
+			return p1 > p2
+		})
+		for i, policy := range assertion.Policy {
+			assertion.PolicyMap[strings.Join(policy, ",")] = i
+		}
+	}
+	return nil
+}
+
+func getSubjectHierarchyMap(model Model) (map[string]int, error) {
```
closetool, line 243: directly pass in policy instead of model

model/model.go
```diff
@@ -206,6 +208,89 @@ func (model Model) PrintModel() {
...
+	policies := model["g"]["g"].Policy
+	subjectHierarchyMap := make(map[string]int)
+	// Tree structure of role
+	policyMap := make(map[string][]string)
+	for _, policy := range policies {
+		if len(policy) < 2 {
+			return nil, errors.New("policy g expect 2 more params")
+		}
```
closetool, line 251: No need, cause casbin initialization will check this, plz check it

model/model.go
```diff
@@ -206,6 +208,89 @@ func (model Model) PrintModel() {
...
+		child := policy[0]
+		parent := policy[1]
+		policyMap[parent] = append(policyMap[parent], child)
+		if _, ok := subjectHierarchyMap[child]; !ok {
+			subjectHierarchyMap[child] = 0
+		}
+		if _, ok := subjectHierarchyMap[parent]; !ok {
+			subjectHierarchyMap[parent] = 0
+		}
+		subjectHierarchyMap[child] = 1
```
closetool, line 261: What if there is a domain?

model/model.go
```diff
@@ -206,6 +208,89 @@ func (model Model) PrintModel() {
...
+				if _, ok := policyMap[nodeValue]; ok {
+					for _, child := range policyMap[nodeValue] {
+						queue.PushBack(child)
+					}
+				}
+			}
+			lv++
+		}
+	}
+	return subjectHierarchyMap, nil
```
closetool, line 291: I don't get it, what you want to achieve for?

closetool (changes requested):
Plz change these

model/model.go
tangyang9464: I noticed that casbin itself does not support circular structure so I did't consider this unreasonable situation

model/model.go
tangyang9464: This requires users to provide a reasonable tree structure.I have explained it above

model/model.go
tangyang9464: https://github.com/casbin/casbin/blob/ecc6d90bd00c77235384013a5fdfe2e7aed82e66/enforcer.go#L289-L304
I found the check in ``BuildRoleLinks() ``（line 299）,but it executes after.

model/model.go
tangyang9464: In this issue,the method can't support domain.I'm not sure if there is such a demand

closetool: I would recommend it

model/model.go
closetool: 1. `rootArr` is not required, you could use `subjectHierarchyMap` directly.
2. `queue` could be reused, you don't need to create it in every loop.

model/model.go
closetool: OK, I'll check it later.

model/model.go
tangyang9464: In ``SortPoliciesByPriority``,I think it does’t assume that ``sub`` always to be first.
``sub`` and ``domain`` ‘s position are not fixed.Do we need to fix the position of ``sub``and ``dom`` in the model like ``g``?It seems simpler

tangyang9464:
@closetool add domain support and "reuse".plz review again

closetool (approved):
LGTM

hsluoyz:
@merajsahebdar plz review.

merajsahebdar:
Thanks for your patience!
I played a lot around it, and find no issues. It's a great feature. 🤩🙏🏻

Just an important question, since this model, does not support multiple roles for each user (or multiple parent subjects for each child subject), can we implement something to raise an error in case of the client requests to do it? or we should just leave it up to the client?

hsluoyz:
> Just an important question, since this model, does not support multiple roles for each user (or multiple parent subjects for each child subject), can we implement something to raise an error in case of the client requests to do it? or we should just leave it up to the client?

@tangyang9464 @closetool any comments?

tangyang9464:
> @tangyang9464 plz add this new `subjectPriority(p.eft) || deny` feature to the docs as a new section of this page: https://casbin.org/docs/en/priority-model

I'll do it later

tangyang9464:
> > Just an important question, since this model, does not support multiple roles for each user (or multiple parent subjects for each child subject), can we implement something to raise an error in case of the client requests to do it? or we should just leave it up to the client?
> 
> @tangyang9464 @closetool any comments?

I think it is enough to indicate in the docs

hsluoyz:
@tangyang9464 tracked in this issue: https://github.com/casbin/casbin-website/issues/252

hsluoyz:
@TraMiu create issue at jcasbin repo

Issue #835, opened 2021-07-09
[Feature] Multiple policy definitions 


wgarunap:
I am referring https://casbin.org/docs/en/syntax-for-models#policy-definition
similar question thread https://github.com/casbin/casbin/issues/720

Currently I am using all in one policy line for RBAC and ABAC which is lengthy and little messy. 
```
p  = type, role,domain,resource,action,requesterOb,dataOb,eft
```

I think having multiple policy definitions can make the implementation more clear and efficient. As a example

**My Expected Model :**
```ini
[request_definition]
r = role,domain,resource,action,requesterOb, dataOb

[policy_definition]
p = role, domain, resource, action,eft
p2= role,domain,requesterOb,dataOb,action,eft
```

**My Expected Policy file:**
```
p,admin, dom1, /data1, edit, allow
p2,user, dom1, r.requesterOb.id!=r.dataOb.id, view, deny
```

```
req1 
admin,dom1,/data1,edit ---> allow

req2
user,dom1,{"id":123},{"id":12444444},view ---> deny
user,dom1,{"id":123},{"id":123},view ---> allow
```

hsluoyz:
@closetool @tangyang9464 plz implement this feature.

tangyang9464:
@wgarunap your request params must corresponding request_define.
But in your example,you have two kind of request

@hsluoyz @closetool this means we need r2 if the example request don't change？

wgarunap:
@tangyang9464 
Ideally there should be multiple request types and multiple policy types.

tangyang9464:
@wgarunap I think so.I'll try to make it more generic

PR #838, opened 2021-07-12
feat: multiple request, policy, effect, matcher type support

tangyang9464:

fix: #835

### DIFF
--- a/enforcer.go
+++ b/enforcer.go
@@ -50,6 +50,14 @@ type Enforcer struct {
 	logger log.Logger
 }
 
+// EnforceContext is used as the first element of the parameter "rvals" in method "enforce"
+type EnforceContext struct {
+	rType string
+	pType string
+	eType string
+	mType string
+}
+
 // NewEnforcer creates an enforcer via file or DB.
 //
 // File:
@@ -440,6 +448,16 @@ func (e *Enforcer) BuildIncrementalRoleLinks(op model.PolicyOp, ptype string, ru
 	return e.model.BuildIncrementalRoleLinks(e.rmMap, op, "g", ptype, rules)
 }
 
+// NewEnforceContext Create a default structure based on the suffix
+func NewEnforceContext(suffix string) EnforceContext {
+	return EnforceContext{
+		rType: "r" + suffix,
+		pType: "p" + suffix,
+		eType: "e" + suffix,
+		mType: "m" + suffix,
+	}
+}
+
 // enforce use a custom matcher to decides whether a "subject" can access a "object" with the operation "action", input parameters are usually: (matcher, sub, obj, act), use model matcher by default when matcher is "".
 func (e *Enforcer) enforce(matcher string, explains *[]string, rvals ...interface{}) (ok bool, err error) {
 	defer func() {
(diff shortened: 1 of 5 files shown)


tangyang9464:
@closetool  @hsluoyz plz review

sagilio (approved):

sagilio:
@tangyang9464 We may need to resolve conflicts first.
image

tangyang9464:
@sagilio 
I did not update my fork repository, which caused the conflict. Now the conflict is resolved after merging, but there is an additional commit and some file changes. I don’t know if there are other ways to solve this

hsluoyz:
@wgarunap plz review

hsluoyz:
@tangyang9464 don't show:

image

enforcer.go
```diff
@@ -459,9 +477,30 @@ func (e *Enforcer) enforce(matcher string, explains *[]string, rvals ...interfac
 			functions[key] = util.GenerateGFunction(rm)
 		}
 	}
+
+	var (
+		rType = "r"
```
wgarunap, line 482: This variable declaration will allocate heap memory everytime we call `enforce ` method. I think It's better if we can define these values as constants so it will be created in the stack.

sagilio:
https://github.com/casbin/casbin/blob/1fa27006ed7fcfe2d42475cdcdbeae36ac044eb6/enforcer.go#L564-L579

@tangyang9464  The `p_eft` here, maybe `p2_eft` and so on? I think we need to consider other cases.

tangyang9464:
@sagilio fixed and add some test

enforcer.go
```diff
@@ -50,6 +50,14 @@ type Enforcer struct {
 	logger log.Logger
 }
 
+// TypeContext is used as the first element of the parameter "rvals" in method "enforce"
+type TypeContext struct {
```
nodece, line 54: It doesn't seem very easy to understand here, use `EnforceContext` looks great.

enforcer.go
```diff
@@ -50,6 +50,14 @@ type Enforcer struct {
...
+	rType string
```
nodece, line 55: It should be replaced with the full name, so like `rType` -> `requestType`.

enforcer.go
```diff
@@ -440,6 +448,16 @@ func (e *Enforcer) BuildIncrementalRoleLinks(op model.PolicyOp, ptype string, ru
 	return e.model.BuildIncrementalRoleLinks(e.rmMap, op, "g", ptype, rules)
 }
 
+// NewTypeContext Create a default structure based on the suffix
+func NewTypeContext(suffix string) TypeContext {
```
nodece, line 452: I think the suffix is unnecessary.

nodece (changes requested):
There is no problem on the whole, I just feel that the variable name needs to be changed, how do you feel about it?

enforcer.go
tangyang9464: I think ``r,p,e,m Type`` understands well, do you think they are ambiguous？Because the previous code is named like this ``ptype``,I follow them.

enforcer.go
tangyang9464: If someone use this method,it means he needs more type such as ``r2,p2``,so I think the ``suffix`` is more convenient.So as not to need to assign one by one

closetool (approved):
I don't like this implementation, but it is ok.

nodece (approved):
LGTM. Hi @hsluoyz, please review this.

hsluoyz:
@wgarunap plz review.

wgarunap:
I apologize for the late reply. Implementation is straightforward for me. 

This is exactly the feature I was expecting. Amazing Job @tangyang9464  
Tested multiple requests and policies along with multiple matchers and I didn't find any issues either.

Thanks for quick response.

hsluoyz:
@tangyang9464 plz add this feature to docs.

Issue #855, opened 2021-08-04
[Question] Result is not matched to the editor in golang program when using RBAC with domain and pattern


ErikXu:
**Want to prioritize this issue? Try:**

[![issuehunt-to-marktext](https://github.com/BoostIO/issuehunt-materials/raw/master/v1/issuehunt-button-v1.svg)](https://issuehunt.io/r/casbin/casbin)

------

**What's your scenario? What do you want to achieve?**
My golang code should be the same with the online editor, but they show different results.

**Your model:**

```ini
[request_definition]
r = sub, dom, obj, act

[policy_definition]
p = sub, dom, obj, act

[role_definition]
g = _, _, _

[policy_effect]
e = some(where (p.eft == allow))

[matchers]
m = g(r.sub, p.sub, r.dom) && keyMatch(r.dom, p.dom) && r.obj == p.obj && keyMatch(r.act, p.act)
```

**Your policy:**

```
p, role||1, *, menu||1, *
p, role||2, *, menu||2, *
p, role||3, *, menu||3, *
p, role||4, *, menu||4, *
g, role||5, role||1, *
g, role||5, role||2, *
g, role||5, role||3, *
g, role||5, role||4, *

p, role||6, *, menu||5, *
p, role||7, *, menu||6, *
p, role||8, *, menu||7, *
p, role||9, *, menu||8, *
g, role||10, role||6, *
g, role||10, role||7, *
g, role||10, role||8, *
g, role||10, role||9, *

g, user||1, role||5, tenant||1
g, user||2, role||10, tenant||2
g, user||3, role||5, tenant||1
g, user||4, role||10, tenant||2
```

**Your request(s):**

```
user||1, tenant||1, menu||2, * ---> false (expected: true)
```

**Golang code:**

``` golang
package main

import (
	"fmt"

	"github.com/casbin/casbin/v2"
	defaultrolemanager "github.com/casbin/casbin/v2/rbac/default-role-manager"
	"github.com/casbin/casbin/v2/util"
)

func main() {
	e, err := casbin.NewEnforcer("./model.conf", "./policy.csv")
	if err != nil {
		fmt.Println(err)
	}

	//e.EnableLog(true)
	e.GetRoleManager().(*defaultrolemanager.RoleManager).AddDomainMatchingFunc("keyMatch", util.KeyMatch)

	ok, _ := e.Enforce("user||1", "tenant||1", "menu||2", "*")
	
        // output false here
        fmt.Println(ok)
}
```

[...]

PR #858, opened 2021-08-07
fix: roles with domain pattern build error

tangyang9464:
Signed-off-by: tangyang9464 <tangyang9464@163.com>

fix: #855

Issue #701, opened 2021-02-03
RoleManager with MatchingFunc depends on order of link creation.


abichinger:
**Describe the bug**
The behavior of the `RoleManager` should be independent of the order in which the links between roles are created.

**To Reproduce**
1. Add the following function to `role_manager_test.go`:
```go
func TestMatchingFuncOrder(t *testing.T) {
	rm := NewRoleManager(10)
	rm.(*RoleManager).AddMatchingFunc("regexMatch", util.RegexMatch)

	rm.AddLink("g\\d+", "root")
	rm.AddLink("u1", "g1")
	testRole(t, rm, "u1", "root", true)

	rm.AddLink("u1", "g1")
	rm.AddLink("g\\d+", "root")
	testRole(t, rm, "u1", "root", true)
}
```
2. Run `go test ./...`

**Expected behavior**
The test shouldn't fail.

**Additional context**
Issue is already resolved in https://github.com/casbin/pycasbin/issues/107

hsluoyz:
@RobotHuang plz take a look.

RobotHuang:
@hsluoyz 
```
	rm.AddLink("g\\d+", "root")
	rm.AddLink("u1", "g1")
	testRole(t, rm, "u1", "root", true)
```
This code also outputs false possibly because `syn.Map` is disordered.

PR #859, opened 2021-08-08
fix: role inheritance fault

closetool:

Fix: #855

### DIFF
--- a/model/assertion.go
+++ b/model/assertion.go
@@ -64,6 +64,15 @@ func (ast *Assertion) buildIncrementalRoleLinks(rm rbac.RoleManager, op PolicyOp
 		}
 	}
 
+	if op == PolicyAdd {
+		for _, rule := range rules {
+			err := rm.BuildRelationship(rule[0],rule[1],rule[2:]...)
+			if err != nil {
+				return err
+			}
+		}
+	}
+
 	return nil
 }
 
@@ -81,6 +90,13 @@ func (ast *Assertion) buildRoleLinks(rm rbac.RoleManager) error {
 			rule = rule[:count]
 		}
 		err := ast.RM.AddLink(rule[0], rule[1], rule[2:]...)
+		if err != nil {
+			return err
+		}
+	}
+
+	for _, rule := range ast.Policy {
+		err := ast.RM.BuildRelationship(rule[0],rule[1],rule[2:]...)
 		if err != nil {
 			return err
 		}
--- a/model/policy.go
+++ b/model/policy.go
@@ -302,7 +302,7 @@ func (model Model) RemoveFilteredPolicy(sec string, ptype string, fieldIndex int
 			effects = append(effects, rule)
 		} else {
 			tmp = append(tmp, rule)
-			model[sec][ptype].PolicyMap[strings.Join(rule, DefaultSep)] = len(tmp)-1
+			model[sec][ptype].PolicyMap[strings.Join(rule, DefaultSep)] = len(tmp) - 1
 		}
 	}
 
--- a/model_test.go
+++ b/model_test.go
@@ -329,6 +329,9 @@ func (rm *testCustomRoleManager) Clear() error { return nil }
 func (rm *testCustomRoleManager) AddLink(name1 string, name2 string, domain ...string) error {
 	return nil
 }
+func (rm *testCustomRoleManager) BuildRelationship(name1 string,name2 string, domain ...string) error {
+	return nil
+}
 func (rm *testCustomRoleManager) DeleteLink(name1 string, name2 string, domain ...string) error {
 	return nil
 }
--- a/rbac/default-role-manager/role_manager.go
+++ b/rbac/default-role-manager/role_manager.go
@@ -15,7 +15,6 @@
 package defaultrolemanager
 
 import (
-	"fmt"
 	"strings"
 	"sync"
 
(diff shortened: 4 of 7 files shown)


closetool:
@hsluoyz @tangyang9464 Changed a lot for next pr to support cross domain

tangyang9464:
@closetool 
I still use #855 sample to test and found some errors which are same as #858
```
p, role||2, dom2, menu||2, *

g, role||5, role||2, tenant||1

g, user||1, role||5, *
g, user||2, role||5, dom2

r, user||2, dom2, menu||2, *  -->true, but should be false ?
```
More interestingly，I just exchange their order and the result is false
```
p, role||2, dom2, menu||2, *

g, user||2, role||5, dom2  // exchange order

g, role||5, role||2, tenant||1

g, user||1, role||5, *

r, user||2, dom2, menu||2, *  -->false
```
I think the reason is
```
g, role||5, role||2, tenant||1
g, user||1, role||5, *

The mapping relationship is
*
user||1  ->  role||5
role||5  ->  role||2   // This shouldn't exist

tenant||1
user||1  ->  role||5
role||5  ->  role||2
```

closetool:
@tangyang9464 I test your case, but the result is `false`

hsluoyz:
@ErikXu plz review

tangyang9464:
@closetool
Different order will lead to different results，this error is the same as #701
```
g, user||1, role||5, *
g, role||5, role||2, tenant||1

g, role||5, role||2, tenant||1
g, user||1, role||5, *
```
It is caused by ``getPatternDomain`` method

closetool:
@tangyang9464 OK, I have an idea, we build the whole relationship after adding links.
Then perhaps everything will be OK.

sagilio:
In fact, Casbin.NET uses the same implementation as you now. but, It can not fix all errors.

This struct can only check whether has links on one domain (included pattern domain like `*`), at the single loop of `match domains`. Then independently handle the result of each loop (now the final result is determined by the first `true` like `allow override`).

So now we may return an error result when two roles/users have a link in more than one domain at the same time.

ErikXu:
@hsluoyz I've changed my local codes and tried the new version of [role_manager.go](https://github.com/casbin/casbin/blob/9bbcbaf37c555cc59d4ff6433c1f9aabbce84ec5/rbac/default-role-manager/role_manager.go) in this pull request. But the result still seems to be incorrect. Did I miss any steps?
> @ErikXu plz review

tangyang9464:
> @hsluoyz I've changed my local codes and tried the new version of [role_manager.go](https://github.com/casbin/casbin/blob/9bbcbaf37c555cc59d4ff6433c1f9aabbce84ec5/rbac/default-role-manager/role_manager.go) in this pull request. But the result still seems to be incorrect. Did I miss any steps?
> 
> > @ErikXu plz review

@Erikxu
you need call ``e.BuildRoleLinks()`` like this
```
e.AddNamedDomainMatchingFunc("g","",util.KeyMatch)
e.BuildRoleLinks()
```

tangyang9464:
> @tangyang9464 OK, I have an idea, we build the whole relationship after adding links.
> Then perhaps everything will be OK.

@closetool maybe we can just call ``e.model.BuildRoleLinks(e.rmMap)`` instead of ``e.BuildRoleLinks()`` because ``e.BuildRoleLinks()`` will call ``rm.Clear()``.Then the error can be resloved

closetool:
@sagilio @tangyang9464 I think @sagilio is right, perhaps I could not fix this bug. We need to clarify which is wild card and which is not.

tangyang9464:
> @sagilio @tangyang9464 I think @sagilio is right, perhaps I could not fix this bug. We need to clarify which is wild card and which is not.

@closetool @sagilio I understand a little bit, but don't know the specific example.Can you show me a simple example?

ErikXu:
I've tried again as you mentioned, and the result turns out to be correct now. Could you please help to estimate and approve this pull request ASAP. So I can use the new version directly. Thanks very much!

> @ErikXu
> you need call `e.BuildRoleLinks()` like this
> 
> ```
> e.AddNamedDomainMatchingFunc("g","",util.KeyMatch)
> e.BuildRoleLinks()
> ```

tangyang9464:
> I've tried again as you mentioned, and the result turns out to be correct now. Could you please help to estimate and approve this pull request ASAP. So I can use the new version directly. Thanks very much!
> 
> > @ErikXu
> > you need call `e.BuildRoleLinks()` like this
> > ```
> > e.AddNamedDomainMatchingFunc("g","",util.KeyMatch)
> > e.BuildRoleLinks()
> > ```

@ErikXu now this PR still has some other bugs.

ErikXu:
OK, I understand. Please keep going with your own schedule. I will continue watching this issue. Thanks!

> > I've tried again as you mentioned, and the result turns out to be correct now. Could you please help to estimate and approve this pull request ASAP. So I can use the new version directly. Thanks very much!
> > > @ErikXu
> > > you need call `e.BuildRoleLinks()` like this
> > > ```
> > > e.AddNamedDomainMatchingFunc("g","",util.KeyMatch)
> > > e.BuildRoleLinks()
> > > ```
> 
> @ErikXu now this PR still has some other bugs.

closetool:
@tangyang9464 @sagilio This pr should fixed all known bugs, plz review

hsluoyz:
@ErikXu

sagilio (approved):

tangyang9464:
@ErikXu now you can just use it normally.don't need to call ``e.BuildRoleLinks()``

ErikXu:
@hsluoyz @tangyang9464 I've checked the latest version. It works fine so far. Please help to approve the pull request or arrange a version release. Thanks very much!


> @ErikXu now you can just use it normally.don't need to call `e.BuildRoleLinks()`

Issue #860, opened 2021-08-11
[Question] GetUsersForRole seems incorrect when using RBAC with domain and pattern


ErikXu:
**Want to prioritize this issue? Try:**

[![issuehunt-to-marktext](https://github.com/BoostIO/issuehunt-materials/raw/master/v1/issuehunt-button-v1.svg)](https://issuehunt.io/r/casbin/casbin)

------

**What's your scenario? What do you want to achieve?**
Your answer here

**Your model:**

```ini
[request_definition]
r = sub, dom, obj, act

[policy_definition]
p = sub, dom, obj, act

[role_definition]
g = _, _, _

[policy_effect]
e = some(where (p.eft == allow))

[matchers]
m = g(r.sub, p.sub, r.dom) && r.obj == p.obj && keyMatch(r.act, p.act)
```

**Your policy:**

```
"g", "role||2", "role||1", "*"
"g", "user||1", "organization||1", "tenant||1"
"g", "user||2", "organization||1", "tenant||1"
```

**Your request(s):**

```
users, err := e.GetUsersForRole("organization||1", "tenant||1")
for _, user := range users {
    fmt.Println(user)
}

// should print 
// user||1
// user||2
```

**golang code:**

``` golang
package main

import (
	"fmt"

	"github.com/casbin/casbin/v2"
	defaultrolemanager "github.com/casbin/casbin/v2/rbac/default-role-manager"
	"github.com/casbin/casbin/v2/util"
)

func main() {
	e, err := casbin.NewEnforcer("./model.conf", "./policy.csv")
	if err != nil {
		fmt.Println(err)
	}

        // Commenting this line or "g", "role||2", "role||1", "*" in policy.csv will get the correct result
	e.GetRoleManager().(*defaultrolemanager.RoleManager).AddDomainMatchingFunc("keyMatch", util.KeyMatch)

	users, err := e.GetUsersForRole("organization||1", "tenant||1")
	for _, user := range users {
		fmt.Println(user)
	}
}
```

ErikXu:
I've submitted a pull request https://github.com/casbin/casbin/pull/864 might fix this issue . Please help to evaluate it. Thanks.

hsluoyz:
@closetool @tangyang9464

PR #864, opened 2021-08-12
fix: get users of role issue

ErikXu:
Fix: https://github.com/casbin/casbin/issues/860

Should continue instead of returning an error in `GetUsers` function refering the `GetRoles` function:

https://github.com/casbin/casbin/blob/master/rbac/default-role-manager/role_manager.go#L264

image


closetool (approved):
LGTM

PR #887, opened 2021-09-07
feat: add LoadPolicyArray() to load policy from array

techoner:


persist/adapter.go
```diff
@@ -37,10 +37,15 @@ func LoadPolicyLine(line string, m model.Model) {
 		return
 	}
 
-	key := tokens[0]
+	LoadPolicy(tokens, m)
+}
+
+// LoadPolicy loads a policy rule to model.
+func LoadPolicy(rule []string, m model.Model) {
```
tangyang9464, line 44: Modify the method name to avoid confusion with the following interface

tangyang9464 (changes requested):

persist/adapter.go
techoner: The scope of the two is different, I think it does not conflict.

tangyang9464: Problems with code readability. Don't be ambiguous

nodece (approved):
LGTM

tangyang9464 (approved):
LGTM

Issue #889, opened 2021-09-17
[Bug] explains variable in LogEnforce function is always an empty array 


linzworld:
# Describe the bug

Some screenshot of the problem

This is the result of the TestEnforceExLog function in enforce_test.go
![image](https://user-images.githubusercontent.com/48641482/133748353-54b47ebf-6f7f-4f16-ad8e-fd171a132161.png)

This is a variable in the code. (this variable is explains. When you call it, it's called logExplains.)
![image](https://user-images.githubusercontent.com/48641482/133748280-809b9810-f807-4b3d-b5b3-c00c3dc423d2.png)

**The explains variable in LogEnforce function is always an empty array, not the hit policy row.
When the result is true, the variable should be the named policy!** 

# How to repetition error

run TestEnforceExLog function in enforce_test.go

# Some important functions
1. func (e *Enforcer) enforce(matcher string, explains *[]string, rvals ...interface{}) (ok bool, err error) 
2. func (l *DefaultLogger) LogEnforce(matcher string, request []interface{}, result bool, explains [][]string)

PR #890, opened 2021-09-17
fix: Set the correct value for logExplains

linzworld:
Signed-off-by: 2448974318@qq.com
fix: #889


linzworld:
This is the result of the TestEnforceExLog function in enforce_test.go, mainly to fix the comparison between before and after errors.

The following is the result of the original error: 
image

After fixing the bug, here is the correct output:
image

hsluoyz:
@tangyang9464 plz review

tangyang9464 (approved):
LGTM

sagilio (approved):
LGTM

Issue #888, opened 2021-09-15
[Bug]Concurrent map read and map write in SyncedEnforcer


phoenixsqf:
**Describe the bug**

here is the stack

```
fatal error: concurrent map read and map write

goroutine 618 [running]:
runtime.throw(0x1a47826, 0x21)
        /usr/local/go/src/runtime/panic.go:1116 +0x72 fp=0xc000b1c6b0 sp=0xc000b1c680 pc=0x436672
runtime.mapaccess2_faststr(0x18199c0, 0xc00006eab0, 0xc000c44dc0, 0x1f, 0x2, 0xc000c44dc0)
        /usr/local/go/src/runtime/map_faststr.go:116 +0x4a5 fp=0xc000b1c720 sp=0xc000b1c6b0 pc=0x4141a5
github.com/casbin/casbin/v2/rbac/default-role-manager.(*RoleManager).domainMatch(0xc0000c6140, 0xc000c44a20, 0x17, 0xc00084c4a8, 0x6, 0x70000c000b08300)
        /root/go/pkg/mod/github.com/casbin/casbin/v2@v2.37.0/rbac/default-role-manager/role_manager.go:376 +0xd6 fp=0xc000b1c7a0 sp=0xc000b1c720 pc=0xe25216
github.com/casbin/casbin/v2/rbac/default-role-manager.(*RoleManager).getPatternDomain.func1(0x17ba520, 0xc000a4d7b0, 0x1927080, 0xc000a20bd0, 0x17c4d01)
        /root/go/pkg/mod/github.com/casbin/casbin/v2@v2.37.0/rbac/default-role-manager/role_manager.go:195 +0x8d fp=0xc000b1c820 sp=0xc000b1c7a0 pc=0xe26bcd
sync.(*Map).Range(0xc00006ea20, 0xc000b1c920)
        /usr/local/go/src/sync/map.go:345 +0x1d9 fp=0xc000b1c8f8 sp=0xc000b1c820 pc=0x484659
github.com/casbin/casbin/v2/rbac/default-role-manager.(*RoleManager).getPatternDomain(0xc0000c6140, 0xc000c44a20, 0x17, 0x10, 0x10, 0x17c4d40)
        /root/go/pkg/mod/github.com/casbin/casbin/v2@v2.37.0/rbac/default-role-manager/role_manager.go:192 +0xf5 fp=0xc000b1c958 sp=0xc000b1c8f8 pc=0xe23f15
[...]

PR #891, opened 2021-09-18
fix: RoleManager match cache not concurrent

tangyang9464:
fix: #888


phoenixsqf:
@tangyang9464 Hello. Should the cache be a concurrent read write map but not a syncMap?

PR #894, opened 2021-09-28
fix: model AddPolicy method can't set PolicyMap correctly

tangyang9464:
Signed-off-by: tangyang9464 tangyang9464@163.com
fix: #895

### DIFF
--- a/model/policy.go
+++ b/model/policy.go
@@ -146,6 +146,7 @@ func (model Model) HasPolicies(sec string, ptype string, rules [][]string) bool 
 func (model Model) AddPolicy(sec string, ptype string, rule []string) {
 	assertion := model[sec][ptype]
 	assertion.Policy = append(assertion.Policy, rule)
+	assertion.PolicyMap[strings.Join(rule, DefaultSep)] = len(model[sec][ptype].Policy) - 1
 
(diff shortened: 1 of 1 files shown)


model/policy.go
```diff
@@ -157,15 +158,19 @@ func (model Model) AddPolicy(sec string, ptype string, rule []string) {
...
 				if idx > idxInsert {
 					assertion.Policy[i] = assertion.Policy[i-1]
+					assertion.PolicyMap[strings.Join(assertion.Policy[i-1], DefaultSep)]++
 				} else {
 					break
 				}
 			}
 			assertion.Policy[i] = rule
 			assertion.PolicyMap[strings.Join(rule, DefaultSep)] = i
+			isSet = true
```
sagilio, line 168: `isSet` only be set true at line 168, so why not return directly here?

sagilio (commented):
LGTM and here is a little suggestion

model/policy.go
tangyang9464: It'ok. Just move the line ``assertion.PolicyMap[strings.Join(rule, DefaultSep)] = len(model[sec][ptype].Policy) - 1`` up

sagilio (approved):

PR #914, opened 2021-11-02
Update example with stricter matching

kilakewe:
In the current example, regex matching will match both POST and the word POSTER because of the structure of the regex. By having them within the first capture group, using the, start of line and end of line token, it will prevent matching words like VIEW and REVIEW from matching


hsluoyz:
@kilakewe plz fix:

image

kilakewe:
> @kilakewe plz fix:
> 
> image

How do I do this in github?

hsluoyz:
@kilakewe for "Semantic Pull Request", you should use a git client to modify your commit message, PR title can be modified on GitHub.

Issue #917, opened 2021-11-16
[Bug] Casbin ABAC rules fail if policy includes any "deny" rows


silverspace:
**Describe the bug**
When an **ABAC** `Super Admin` matcher is used in conjunction with a **policy** that includes `deny` effects, the matcher erroneously fails if **any** policy in `policy.csv` contains a `deny` effect. 

**To Reproduce**
See this shared example I created using the casbin editor: 
https://casbin.org/casbin-editor/#E3LXSU3ZS

Observe that the two requests succeed if the `"foobar"` rule in the policy is changed to `"allow"`, and both fail if the `"foobar"` rule is `"deny"`. 

Note that I pared down my actual model and policy files to be the minimal reproducible use case, which is why this does not use any groups, actions, users, etc. This fails even if I have legitimate users and roles that do not match to the "foobar" user. 

Short summary without the above online editor:

**model.conf**:
```
[policy_effect]
e = some(where (p.eft == allow)) && !some(where (p.eft == deny))

[matchers]
m = r.sub.Superuser || r.sub.User == r.obj.Owner
```

**policy.csv**:
```
p, foobar, /blah, read, deny
```

**Request**:
```
{ User: 'alice', Superuser: true }, { Owner: 'bob'}, read
```

**Expected behavior**
If we have an ABAC rule to allow access for `Super Users`, that access should not be contingent upon unrelated subjects having `deny` effects in the policy file. 

**Screenshots**

**Successful run without `deny` rule in policy file:**
![image](https://user-images.githubusercontent.com/15318933/141872427-a710dace-ed85-487d-a1cd-a457e15a66cf.png)

**Unsuccessful run with unrelated `deny` rule in policy file:**
![image](https://user-images.githubusercontent.com/15318933/141872473-5d69afa3-44f1-4475-afdf-fbfb1cc56060.png)

hsluoyz:
@abichinger @Abingcbc

tlyyxjz:
Really nice minimal repro. The `!some(where(p.eft == deny))` clause makes any unrelated deny row poison the entire evaluation — easy to forget when you mix ABAC (subject attributes) with policy effects. That ABAC × deny interaction is a classic "silently wrong" trap: it looks fine in tests, fails deep in prod. Exactly the kind of matcher/effect combination an automated diagnostic (over policy_effect + matcher) should catch before it ships.

PR #918, opened 2021-11-16
fix: abac not using policy

Abingcbc:
Signed-off-by: abingcbc <abingcbc626@gmail.com>

Fix: https://github.com/casbin/casbin/issues/917

## How it works

Ignore all policies when matcher doesn't contain any match related to policy.

### DIFF
--- a/enforcer.go
+++ b/enforcer.go
@@ -17,6 +17,7 @@ package casbin
 import (
 	"errors"
 	"fmt"
+	"strings"
 
 	"github.com/Knetic/govaluate"
 	"github.com/casbin/casbin/v2/effector"
@@ -550,7 +551,7 @@ func (e *Enforcer) enforce(matcher string, explains *[]string, rvals ...interfac
 	var effect effector.Effect
 	var explainIndex int
 
-	if policyLen := len(e.model["p"][pType].Policy); policyLen != 0 {
+	if policyLen := len(e.model["p"][pType].Policy); policyLen != 0 && strings.Contains(expString, pType + "_"){
 		policyEffects = make([]effector.Effect, policyLen)
 		matcherResults = make([]float64, policyLen)
 
--- a/examples/abac_not_using_policy_model.conf
+++ b/examples/abac_not_using_policy_model.conf
@@ -1,0 +1,11 @@
+[request_definition]
+r = sub, obj, act
+
+[policy_definition]
+p = sub, obj, act, eft
+
+[policy_effect]
+e = some(where (p.eft == allow)) && !some(where (p.eft == deny))
+
+[matchers]
+m = r.sub == r.obj.Owner
--- a/examples/abac_rule_effect_policy.csv
+++ b/examples/abac_rule_effect_policy.csv
@@ -1,0 +1,4 @@
+p, alice, /data1, read, deny
+p, alice, /data1, write, allow
+p, bob, /data2, write, deny
+p, bob, /data2, read, allow
--- a/model_test.go
+++ b/model_test.go
@@ -549,6 +549,17 @@ func newTestSubject(name string, age int) testSub {
 	return s
 }
 
+func TestABACNotUsingPolicy(t *testing.T) {
+	e, _ := NewEnforcer("examples/abac_not_using_policy_model.conf", "examples/abac_rule_effect_policy.csv")
+	data1 := newTestResource("data1", "alice")
+	data2 := newTestResource("data2", "bob")
+
+	testEnforce(t, e, "alice", data1, "read", true)
+	testEnforce(t, e, "alice", data1, "write", true)
+	testEnforce(t, e, "alice", data2, "read", false)
+	testEnforce(t, e, "alice", data2, "write", false)
+}
+
 func TestABACPolicy(t *testing.T) {
 	e, _ := NewEnforcer("examples/abac_rule_model.conf", "examples/abac_rule_policy.csv")
 	m := e.GetModel()


hsluoyz:
@nodece plz review

hsluoyz:
@silverspace plz review

nodece (approved):
LGTM

examples/abac_rule_deny_model.conf
```diff
@@ -0,0 +1,11 @@
+[request_definition]
+r = sub, obj, act
+
+[policy_definition]
+p = sub_rule, obj, act, eft
+
+[policy_effect]
+e = some(where (p.eft == allow))
```
silverspace, line 8: This is missing the critical deny policy effect. Could you change this to:
```
e = some(where (p.eft == allow)) && !some(where (p.eft == deny))
```

silverspace: (I think this issue is specific to the `allow-and-deny` policy effect.)

enforcer.go
```diff
@@ -550,7 +551,7 @@ func (e *Enforcer) enforce(matcher string, explains *[]string, rvals ...interfac
 	var effect effector.Effect
 	var explainIndex int
 
-	if policyLen := len(e.model["p"][pType].Policy); policyLen != 0 {
+	if policyLen := len(e.model["p"][pType].Policy); policyLen != 0 && strings.Contains(expString, pType + "_"){
```
silverspace, line 554: Thanks for this super fast PR!

examples/abac_rule_deny_model.conf
hsluoyz: @Abingcbc

Abingcbc: Done❤️

sagilio (changes requested):
Can you provide a more reasonable test case? I try to remove the matcher check and run the test, it returns another error:
image
image

Then I try to add `eft` to the abac_model.conf, the new test is passed:
image

I think your changes have nothing to do with the error in the issue.

Abingcbc:
> Can you provide a more reasonable test case? I try to remove the matcher check and run the test, it returns another error: image

@sagilio sa I don't understand what you mean, can you describe in more detail？

Abingcbc:
@sagilio Yep. I realized this error when  I wrote the test. But if the matcher doesn't use policy, no policy should be matched for the request. So no matter what policy definitions look like,  policy should have no effect on the request?

sagilio:
> @sagilio Yep. I realized this error when I wrote the test. But if the matcher doesn't use policy, no policy should be matched for the request. So no matter what policy definitions look like, policy should have no effect on the request?

Yeah, you changes is right，we need not scan.all policies when matcher is not include rtype. But this just a enhance about performance. I do not know why the wrong result returned in #917.

Abingcbc:
> > @sagilio Yep. I realized this error when I wrote the test. But if the matcher doesn't use policy, no policy should be matched for the request. So no matter what policy definitions look like, policy should have no effect on the request?
> 
> Yeah, you changes is right，we need not scan.all policies when matcher is not include rtype. But this just a enhance about performance. I do not know why the wrong result returned in #917.

I think the root cause is that `policy` is for other model like `rbac` and the request's model is `abac`.

When the matcher doesn't use any policy and the request meets the judgment of the matcher, the request will match all policies because the matcher always returns true.
Therefore, if there is `deny` policy, the result of merged effects will be `deny`.

Take https://github.com/casbin/casbin/issues/917 as an example, request:
```
{ User: 'alice', Superuser: true }, { Owner: 'bob'}, read
```

will match policy

```
p, foobar, /blah, read, deny
```

and the merged effect is `deny`, although it is allowed in respect to `abac`.

sagilio:
@Abingcbc I see, you are right. We may need a more effective test case to descript the situation.

Abingcbc:
> @Abingcbc I see, you are right. We may need a more effective test case to descript the situation.

@sagilio I have updated the test case. PTAL ❤️

hsluoyz:
@sagilio plz review

sagilio (approved):
LGTM

harshalmittal4:
Hi @Abingcbc , @hsluoyz ,
I was seeking to understand the abac policy and read the discussion above and at https://github.com/casbin/casbin/issues/917, and get that the request must not depend on policy file for a model like below, i.e. it must be allowed everytime the condition `r.sub == r.obj.Owner` holds. Is this correct?

model - 

```
[request_definition]
r = sub, obj, act

[policy_definition]
p = sub, obj, act, eft

[policy_effect]
e = some(where (p.eft == allow)) && !some(where (p.eft == deny))

[matchers]
m = r.sub == r.obj.Owner
```

For a request as `alice, { Owner: 'alice'}, write`, I used the above model and the below policy to run in online editor - 
```
p, alice, /data1, read, deny
p, alice, /data1, write, allow
p, bob, /data2, write, allow
p, bob, /data2, read, allow
```
 and got the result as false. 

On replacing `p, alice, /data1, read, deny` in policy file with `p, alice, /data1, read, allow` (so that all eft in policy become allow), the result changes to true. Is this the correct behavior (i.e request depending on policy file)?

hsluoyz:
@harshalmittal4 the online editor is pure frontend (https://github.com/casbin/casbin-editor), which uses Node-Casbin: https://github.com/casbin/node-casbin for Casbin enforcement. Usually we have latest features in Golang version Casbin and Node-Casbin maybe still doesn't have the feature of this PR: https://github.com/casbin/casbin/pull/918 yet. 

Issue proposed here: https://github.com/casbin/node-casbin/issues/433

harshalmittal4:
Thanks @hsluoyz , got it. I'm currently getting more familiar with Casbin and would like to try contributing to some issues once I get comfortable

Issue #927, opened 2021-12-09
[Question] Parameters in Cache interface for CachedEnforcer


alikhanz:
**What's your scenario? What do you want to achieve?**

I'm want to use Redis cache in CachedEnforcer

I'm trying to implement `Cache` interface for using Redis

https://github.com/casbin/casbin/blob/3d0481d5d8357c466c4b9abe04b54d137a7c8cf1/persist/cache/cache.go#L21-L39 

And i'm not understand why 3rd parameter in `Set` method is `extra ...interface{}` and not like `expiration time.Duration`? In `CachedEnforcer` 3rd parameter is expriation time:

https://github.com/casbin/casbin/blob/master/enforcer_cached.go#L86

https://github.com/casbin/casbin/blob/3d0481d5d8357c466c4b9abe04b54d137a7c8cf1/enforcer_cached.go#L82-L88

My point of view is that not strictly specifying the parameter can lead to errors in the future, may be change this signature?

hsluoyz:
@closetool any reason for using `extra` and `interface{}`? Actually I didn't see anywhere the arg is used or type casted?

alikhanz:
Also im not understand why `expireTime` property defined as `uint`, and not `time.Duration`? 
I can't find comments what is `expireTime`: milliseconds/seconds/minutes/hours?

hsluoyz:
@alikhanz I think this function and its args may be not very well-designed. Can you make a PR with a more precise declaration?

alikhanz:
> @alikhanz I think this function and its args may be not very well-designed. Can you make a PR with a more precise declaration?

Yes i can, but this changes can break backward compatibility with others `Cache` implementation, will this PR be released in v3?

hsluoyz:
@alikhanz I think there are no other Cache implementations in any other repos (you can also search on GitHub to confirm it). Only this repo uses it. So changing the function args will not cause visible breaking changes to the users.

alikhanz:
> @alikhanz I think there are no other Cache implementations in any other repos (you can also search on GitHub to confirm it). Only this repo uses it. So changing the function args will not cause visible breaking changes to the users.

Done: https://github.com/casbin/casbin/pull/934

hsluoyz:
Closed as: https://github.com/casbin/casbin/pull/934

PR #934, opened 2021-12-14
feat: improve CachedEnforcer

alikhanz:
Fix: https://github.com/casbin/casbin/issues/927

@hsluoyz also i'm added redis cache implementation, but i dont know how to create new repositories :) Can you review, and prompt how i can add this implementation to your project?
https://github.com/alikhanz/casbin-cache-redis

### DIFF
--- a/enforcer_cached.go
+++ b/enforcer_cached.go
@@ -76,7 +76,7 @@ func (e *CachedEnforcer) Enforce(rvals ...interface{}) (bool, error) {
 		return e.Enforcer.Enforce(rvals...)
 	}
 
-	if res, err := e.getEnforcerCachedResult(key); err == nil {
+	if res, err := e.getEnforceCachedResult(key); err == nil {
 		return res, nil
 	} else if err != cache.ErrNoSuchKey {
 		return res, err
@@ -130,7 +130,7 @@ func (e *CachedEnforcer) RemovePolicies(rules [][]string) (bool, error) {
 	return e.Enforcer.RemovePolicies(rules)
 }
 
-func (e *CachedEnforcer) getEnforcerCachedResult(key string) (res bool, err error) {
+func (e *CachedEnforcer) getEnforceCachedResult(key string) (res bool, err error) {
 	e.locker.RLock()
 	defer e.locker.RUnlock()
 	cacheResult, err := e.cache.Get(key)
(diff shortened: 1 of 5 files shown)


sagilio (changes requested):
Why change bool to bytes?

alikhanz:
> Why change bool to bytes?

Hello.
Because `Cache` interface is more general, than the `enforce` response result. In future you can store in this cache any data you need. 
We can declare other interface, specific for `enforce` response, but i think this is unnecessary

alikhanz:
Are we waiting for something?

sagilio:
> > Why change bool to bytes?
> 
> Hello, @sagilio. Because `Cache` interface is more general, than the `enforce` response result. In future you can store in this cache any data you need. We can declare other interface, specific for `enforce` response, but i think this is unnecessary

@hsluoyz @nodece Any comment about this? I think different caches should have different interfaces, instead of using a common interface to store all cache structs.
For example, Casbin.NET has four different caches：
https://github.com/sagilio/Casbin.NET/tree/feat%23229/NetCasbin/Abstractions/Caching
image

alikhanz:
> > > Why change bool to bytes?
> > 
> > 
> > Hello, @sagilio. Because `Cache` interface is more general, than the `enforce` response result. In future you can store in this cache any data you need. We can declare other interface, specific for `enforce` response, but i think this is unnecessary
> 
> @hsluoyz @nodece Any comment about this? I think different caches should have different interfaces, instead of using a common interface to store all cache structs. For example, Casbin.NET has four different caches： https://github.com/sagilio/Casbin.NET/tree/feat%23229/NetCasbin/Abstractions/Caching image

Yeah, it's good point, but see how Casbin.NET interfaces was defined:
`public interface IEnforceCache` - its specific interface, for specific method, and usecase.
As i said, we can add more interfaces for specific cache operations, if it really matters, but basic we need `cache.Cache` interface for using it in other implementations.

alikhanz:
And more: 
I think we must more improvements in cache invalidation logic. 
Because current implementation works badly in RBAC requests. after changing the user's role, the cache will not be invalidated.
Now I have no idea how to fix it correctly. 
As an idea: we can add cache tags for subjects, and drop cache by this tags after changing roles.

alikhanz:
Hi, @sagilio 
Unfortunately i ran into a lot of other problems with cache enforcer, as a result we can't continue to use casbin in our project due to perfomance and issues with cache enforcer, and i will not continue to support this PR, we can close it, i think.
P.S. Im signed CLA, but checks still not updated. After "recheck" too.

PR #940, opened 2021-12-17
fix: Not set hit rule when matched allow rule for rbac_with_deny_policy

techoner:
Signed-off-by: Jon Lee <techlee@qq.com>


sagilio (approved):
I think it is better to add some test cases for this case.

Issue #434, opened 2020-04-24
Suggestion: Make ClearPolicy optional during LoadFilteredPolicy


ashwinsadeep:
I'm designing a system where we have a rather small number of access control policies - say less than 1000. These policies will be defined on roles. However, the system has a somewhat large number of users and corresponding group membership policies - say 10M users & 20M group memberships considering an average of 2 groups per user.

I was thinking of an approach where I can load all access control policies - `p`- on initialization and then load the grouping policies - `g` - on demand while processing requests. 

However, every `LoadFilteredPolicy` is also clearing the existing policies which are loaded which is not ideal. I'm proposing that we add an `autoClear` field which can control this behaviour.

Another approach we can take is to make Model pluggable the same way you have done for Adapters & Watchers. This will allow more control over `ClearPolicy` behaviour.

Please note that the group memberships itself will be cached and probably have the enforcers sharded based on a user_id_hash maybe. These details are probably out of scope for this discussion, but mentioning anyway to add some context.

GopherJ:
@ashwinsadeep  The problem of this method is **we can have duplicated policies in memory**, casbin is an auth library but has no ability to do things like Full-text searching to find the part of policies related to request.

we didn't find a good method for implementing `dynamic filtered adapter` yet.

ashwinsadeep:
IMO, this will be a requirement for any medium to large sized system where storing the entire policy information in memory is not practical. In my case at least, the expected access patterns will be such that only a relatively small subset of subjects accessing a small subset of resources will be active at any given point of time.

Further to that, I don't think a `dynamic filtered adapter` in itself is going to be enough. We'll need significant changes in the model implementation to allow for pruning policies by a `filter` as well. For instance, I'll want to build the adapter such that it'll have an in-memory LRU cache on top of the persistent datastore. So whenever a a policy is getting evicted from memory, I'll want to remove it from model as well while removin

GopherJ:
@ashwinsadeep while `dynamic filtered adapter` loads filtered policies, casbin will clean the policies in memory. Actually I don't think we will really need to prune policies from model because we just need to overwrite with filtered policies and then rebuild the role links, the process here is easy to be done in casbin today. It can be improved: like not overwrite all but do incremental changes.

However, I didn't see a clean and efficient method for loading the part of policies related to request. Like a request:

```
alice, domain1, /data1, read
```

we will need to find the list of policies related to it so that we can load them dynamically. The speed here is important because:

```
[...]

PR #942, opened 2021-12-18
refactor: reimplement default role manager

abichinger:
@techoner, I re-implemented the default role manager as requested https://github.com/casbin/pycasbin/pull/232#issuecomment-995623085

The implementation is almost identical to [pycasbin:role_manager.py](https://github.com/casbin/pycasbin/blob/17be9e13f24144aca5fe5e1c681653e04b2fe24d/casbin/rbac/default_role_manager/role_manager.py)

Changes:

- split `RoleManager` into `DomainManager` and `RoleManager`
- remove `RoleManager.BuildRelationship`

Bugfixes:

- avoid mixing RBAC systems in `Enforcer.GetImplicitRolesForUser` and `Enforcer.GetImplicitUsersForRole`

This is my first contribution using golang, so please review carefully :)


sagilio (changes requested):
This PR seems to improve the code quality, I think this is very good. I need more experiments to confirm that these changes are necessary because here has some breaking changes.

But, from the benchmark result, I think the idea may not be good.
**It seems that the calculation amount of AddLink function which always single execution is transferred to the more frequently executed one hasLink.**

**this PR**
image

**now main branch**
image

abichinger:
@sagilio, the role manager map `rmMap` of `DomainManager` will be populated on the first call to `HasLink`.
 e.g. 
```go
dm.HasLink("alice", "admin", "domain1")
// will create an RoleManager for "domain1" and store it in rmMap
```
That's why there are such huge performance differences.

But if I build all role links with `AddLink`, the performance is still not great. (see below)
I will try to further improve performance.

Code change:
```go
// AddLink adds the inheritance link between role: name1 and role: name2.
// aka role: name1 inherits role: name2.
func (dm *DomainManager) AddLink(name1 string, name2 string, domains ...string) error {
	domain, err := dm.getDomain(domains...)
	if err != nil {
		return err
	}
	_ = dm.getRoleManager(domain) //<=== create role manager if it does not exist

	links := dm.getLinks(domain)
	links.Add(Link{name1, name2})
	for _, rm := range dm.affectedRoleManagers(domain) {
		_ = rm.AddLink(name1, name2, domains...)
	}
	return nil
}
```

modified code
bench-modified

master
bench-master

sagilio:
Cool! These changes make the role match relationship static, and separate different concepts into themselves zone (different `sync.Map`).
I will review it carefully later.

sagilio (changes requested):
I have read all changes carefully. because different concepts of the same "g policy token" can hardly be distinguished accurately in the old logic. The idea of the new design is generally correct.

Here are some problems:
1. For casbin, because there is no requirement on the order of the same g type policies, the final structure of the same set of policies should be exactly the same. When using matching func, we cannot determine the final state in the build process, it is the reason we need to call `BuildRelationship` again after `AddLink` before.  (https://github.com/casbin/casbin/pull/859).
2. The role object cannot be created unlimitedly based on the `HasLink` parameters, which will make the number of role objects uncontrollable when the program is running for a long time.

The rest are just some suggestions:
1. It is unnecessary to change `RoleManager` to `DomainManager` and make `RoleManagerImpl` public. We can make the refactor have no impact on users.
2. `iterXXX` does not suit the naming style of go.  
3. We can try to reduce some temp maps creation in `HasLink`.

abichinger:
@sagilio thanks for your review.

Regarding the problems found:

1. I disagree. I believe the new implementation is able to build the correct relationships for each set of policies (regardless of their order). Have you found an example were this statement does not hold?
Furthermore, on a call to `AddMatchingFunc` the role manager will rebuild itself. Hence, a call to `Enforcer.BuildRoleLinks()` is not required anymore.
2. You are right. I'm going to limit the number of created roles by `HasLink` with a cache.

Regarding the suggestions:

1. good point :+1: 
2. I will rename `iterXXX` to `rangeXXX`
3. I will try

util/util.go
```diff
@@ -217,3 +235,106 @@ func RemoveDuplicateElement(s []string) []string {
...
+
+func (cache *LRUCache) remove(n *node) {
+	delete(cache.m, n.key)
+	n.prev.next = n.next
+	n.next.prev = n.prev
+}
+
+func (cache *LRUCache) add(n *node) {
+	cache.m[n.key] = n
+	head_next := cache.head.next
```
sagilio, line 278: `headNext`

rbac/default-role-manager/role_manager.go
```diff
@@ -25,525 +26,635 @@ import (
...
-					if result {
-						return true, nil
-					} else {
-						for _, r := range role1.roles {
-							roleQueue = append(roleQueue, r.name)
-						}
-					}
-				}
-			}
+	next_roles := map[string]*Role{}
```
sagilio, line 320: `nextRoles`

sagilio (commented):
@abichinger 
1. I tried some situations, and I found that the new implementation will also try to add a match relationship in HasLink. If the input is correct, the results are always correct. Depending on the order of `AddLink`, there will be a different number of `AddMatch` calls, but it doesn't seem to affect the `HasLink` result.
https://github.com/abichinger/casbin/blob/3df5dcb7ed413184a3312bfd8124516d5371fabf/rbac/default-role-manager/role_manager.go#L240-L250
2. New changes LGTM.

sagilio (approved):

sagilio:
@abichinger 
1. A simple example:
```
func TestMatchingFuncOrder2(t *testing.T) {
	rm := NewRoleManager(10)
	rm.AddMatchingFunc("regexMatch", util.RegexMatch)

	_ = rm.AddLink("g\\d+", "root")
	_ = rm.AddLink("u1", "g1")
	testRole(t, rm, "u1", "root", true)

	_ = rm.Clear()

	_ = rm.AddLink("u1", "g1")
	_ = rm.AddLink("g\\d+", "root")
	testRole(t, rm, "u1", "root", true)
}
```
I have added output in `GetUser`, and here is the result:
image
image

2. I think it is not very dangerous to keep the matched role in the memory. We need to limit the role that is completely unrelated to the whole role map, not even caching such the role at all, this way we can also avoid writing the LRU cache. In fact, we have another cache of the g function to cache the outrageous error cases.

3. I would like to sync your new implement to `Casbin.NET`.

abichinger:
@sagilio 

1. Thank you for your example! `add match g1 g\d+` is incorrect. This bug is fixed with a17864db61bcc14acfa1e7d4ce6bbf9fc21c02c8
2. I removed the `roleCache`. `RoleManagerImpl` will now remove a role and all its references, if the role was created by `HasLink`, `GetRoles`, ... . 
Regarding the cache, I think we should keep the LRUCache to limit the size of other caches.
3. Great! :)

tangyang9464:
It looks like this is a very logically clean implementation and helps with previous cross-domain issues. 
After this PR is merged, maybe I'll try to solve the cross domain issue based on this and sync the new implement to ``jCasbin``.

tangyang9464 (approved):
In addition, ``DomainManager`` may be unnecessary, and ``RoleManager`` can also be used directly.
LGTM

hsluoyz:
@abichinger plz resolve:

image

abichinger:
@hsluoyz done
I rebased the branch onto master and squashed all commits.

util/util_test.go
```diff
@@ -181,3 +181,42 @@ func TestReplaceEvalWithMap(t *testing.T) {
...
+		cacheValues = append(cacheValues, v.value.(int))
+	}
+
+	if SetEqualsInt(values, cacheValues) == false {
+		t.Errorf("cache values: %d supposed to be %d", cacheValues, values)
+	}
+}
+
+func TestLRUCache(t *testing.T) {
+
```
hsluoyz, line 212: go fmt all the code

hsluoyz (changes requested):

hsluoyz:
@tangyang9464 plz review again

util/util_test.go
abichinger: Running `gofmt -s -w -l .` does not change anything for me.

hsluoyz: remove empty lines like 212

tangyang9464:
LGTM

hsluoyz (approved):

hsluoyz:
@tangyang9464 @abichinger does this PR contain breaking change? We don't want to release v3 for now. We should avoid any unnecessary breaking changes to bump the major version. It's unplanned.

image

tangyang9464:
> @tangyang9464 @abichinger does this PR contain breaking change? We don't want to release v3 for now. We should avoid any unnecessary breaking changes to bump the major version. It's unplanned.
> 
> image
> 

@hsluoyz Yes, it is a major change, refactoring the RoleManager. Maybe can create a beta branch.

sagilio (approved):

hsluoyz:
@tangyang9464 I don't think refactoring some code is a good reason to make a major release. Remember every time we bump a major version, we cause big influence to the users and some of them may leave us. We must have good reasons (like a lot of very fatanstic and next-generation features) to do this.

abichinger:
@hsluoyz We could keep `RoleManager.BuildRelationship` for now and mark it as deprecated, then it would be a patch release.

hsluoyz:
@abichinger why BuildRelationship is no longer required?

abichinger:
@hsluoyz The new implementation builds the role hierarchy on the fly. In other words the relationship between roles is already established after a call to `AddLink`.

hsluoyz:
@abichinger OK, got it. Can you restore `RoleManager.BuildRelationship()` back and remove the breaking change wording?

abichinger:
@hsluoyz `RoleManager.BuildRelationship()` is now again part of the `RoleManager` interface, but no longer has a functionality.

I think this changes should only increment the patch version, because there is no new functionality and the interface of the RoleManager didn't change.

hsluoyz:
@abichinger plz fix: https://github.com/casbin/casbin/runs/5787773431?check_suite_focus=true

image

hsluoyz:
@abichinger re-run for 3 times, but still failed:

image

abichinger:
The linting job did fail again. At the moment I can't think of a solution.
Maybe someone else has an idea. @tangyang9464 @sagilio

hsluoyz:
@abichinger new release is not out because of this failure, plz fix it: https://github.com/casbin/casbin/actions

image

tangyang9464:
> The linting job did fail again. At the moment I can't think of a solution.
> Maybe someone else has an idea. @tangyang9464 @sagilio 

@abichinger It seems that others have the same problem [#434](https://github.com/golangci/golangci-lint-action/issues/434), but it is not clear which part of the code caused the problem. Maybe you can reopen a PR and restore some code to see if ``Action`` fails.

hsluoyz:
@abichinger can you fork this repo and test which part of code in the PR causes this issue? We'd better get rid of it soon because it blocks our release process.

PR #966, opened 2022-03-03
perf: Put function calls last in example configurations.

thetorpedodog:
String comparisons are a *lot* cheaper than govaluate function calls,
so putting them last in the matcher function can save a shocking amount
of CPU time and allocations, while being functionally identical.

This change reorders the example configurations to put function calls
last where applicable, to provide a good example for users to base
their own configuration off of.

The results of a benchmark where I compared the two options:

    g_first:   g(r.sub, p.sub) && r.obj == p.obj && r.act == p.act
    eq_first:  r.obj == p.obj && r.act == p.act && g(r.sub, p.sub)

    Running tool: /usr/local/bin/go test -benchmem -run=^$ -bench ^BenchmarkRBACModelSizes$ github.com/casbin/casbin/v2

    goos: linux
    goarch: amd64
    pkg: github.com/casbin/casbin/v2
    cpu: Intel(R) Core(TM) i7-10750H CPU @ 2.60GHz
    BenchmarkRBACModelSizes/small/g_first-4      14827      79860 ns/op    38767 B/op      874 allocs/op
    BenchmarkRBACModelSizes/small/eq_first-4     28083      42782 ns/op    16938 B/op      395 allocs/op
    BenchmarkRBACModelSizes/medium/g_first-4      1544     774049 ns/op   372522 B/op     8141 allocs/op
    BenchmarkRBACModelSizes/medium/eq_first-4     4285     284674 ns/op   101990 B/op     2572 allocs/op
    BenchmarkRBACModelSizes/large/g_first-4        159    7562438 ns/op  3533126 B/op    80803 allocs/op
    BenchmarkRBACModelSizes/large/eq_first-4       462    2707900 ns/op   943430 B/op    24257 allocs/op
    PASS
    ok   github.com/casbin/casbin/v2 10.320s

This benchmark is currently only local, but I can commit it to the test
suite if desired, so that the difference can be obvious.

### DIFF
--- a/examples/abac_rule_model.conf
+++ b/examples/abac_rule_model.conf
@@ -8,4 +8,4 @@ p = sub_rule, obj, act
 e = some(where (p.eft == allow))
 
 [matchers]
-m = eval(p.sub_rule) && r.obj == p.obj && r.act == p.act
+m = r.obj == p.obj && r.act == p.act && eval(p.sub_rule)
--- a/examples/eval_operator_model.conf
+++ b/examples/eval_operator_model.conf
@@ -8,4 +8,4 @@ p = sub_rule, obj_rule, act
 e = some(where (p.eft == allow))
 
 [matchers]
-m = eval(p.sub_rule) && eval(p.obj_rule) && (p.act == '*' || r.act == p.act)
+m = (p.act == '*' || r.act == p.act) && eval(p.sub_rule) && eval(p.obj_rule)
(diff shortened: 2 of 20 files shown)


thetorpedodog:
After a few days of this change being in production for us, we’ve found that it’s reduced Casbin latency by about 60%!

sagilio:
According to the principle of expression, this is a valid operation to improve performance. I think we can add this to the docs.
It sounds good If we can auto optimize the matcher.

hsluoyz:
@thetorpedodog can we do this shifting automatically in the code? I prefer not changing the order in the model text for better readability.

hsluoyz:
The user can do this by himself. For docs clarity, we prefer not to do this

PR #985, opened 2022-04-12
feat: Log panic stack

erichulburd:
I've run into a couple of nil pointer exceptions that would have benefited from a stack reference. This simple adds the stack trace to the recovery error.


hsluoyz:
@tangyang9464 plz review

tangyang9464 (approved):
lgtm

Issue #969, opened 2022-03-15
[Question] Get implicit permissions with pattern-matching in RBAC don't work


sasakiyori:
**Want to prioritize this issue? Try:**

[![issuehunt-to-marktext](https://github.com/BoostIO/issuehunt-materials/raw/master/v1/issuehunt-button-v1.svg)](https://issuehunt.io/r/casbin/casbin)

------

**What's your scenario? What do you want to achieve?**

Hello. From the [Issue:968](https://github.com/casbin/casbin/issues/968), I use this function to achieve what I expected:
```go
e.AddNamedDomainMatchingFunc("g", "KeyMatch", util.KeyMatch)
```
It uses the ```KeyMatch``` func, trickily make it access all domains. But I found something wrong when I tried to get user's permission list.

As the [tutorial](https://casbin.org/docs/en/rbac-api#getimplicitpermissionsforuser) said, I tried ```GetImplicitRolesForUser``` and ```GetImplicitPermissionsForUser``` func to get the roles and permissions for user.

I found the execution result of  ```GetImplicitRolesForUser``` is what I need, but the result of ```GetImplicitPermissionsForUser``` is empty!

---
**Your model:**

```ini
[request_definition]
r = sub, dom, obj, act

[policy_definition]
p = sub, dom, obj, act

[role_definition]
g = _, _, _

[policy_effect]
e = some(where (p.eft == allow))

[matchers]
m = g(r.sub, p.sub, r.dom) && r.obj == p.obj && r.act == p.act
```

**Your policy:**

```
p, visitor, *, /list, GET
p, visitor, *, /menu, GET
p, visitor, *, /insight, GET
p, user, *, /update, POST
p, manager, *, /manage, POST
p, sysadmin, *, /account, GET


g, user, visitor, *
g, manager, user, *
g, sysadmin, manager, *


g, bob, manager, domain1
```

**Your request(s):**
```go
// result: [manager user visitor]  --> (expected: [manager user visitor])
e.GetImplicitRolesForUser("bob", "domain1")

// result: [] --> (expected: [[manager * /manage POST] [user * /update POST] [visitor * /list GET] [visitor * /menu GET] [visitor * /insight GET]])
// or [[manager domain1 /manage POST] [user domain1 /update POST] [visitor domain1 /list GET] [visitor domain1 /menu GET] [visitor domain1 /insight GET]] be better!
e.GetImplicitPermissionsForUser("bob", "domain1")
```

---
[...]

PR #992, opened 2022-04-20
feat: support domain matching when getting permissions

Abingcbc:
fix: https://github.com/casbin/casbin/issues/969

## How it works
`GetImplicitPermissionsForUser` depends on `GetFilteredPolicies` to query permissions. But when `enforcer` gets filtered policies, it doesn't support domain matching.
So there are main changes in this PR:
1. Make `match` of role manager public and reuse this function for domain matching when getting policies.
2. Add a new function `GetFilteredPolicyWithDomainMatching` for the user requirements mentioned in
https://github.com/casbin/casbin/issues/969.
4. Unit tests.

### DIFF
--- a/rbac_api.go
+++ b/rbac_api.go
@@ -17,6 +17,7 @@ package casbin
 import (
 	"github.com/casbin/casbin/v2/constant"
 	"github.com/casbin/casbin/v2/errors"
+	"github.com/casbin/casbin/v2/rbac/default-role-manager"
 	"github.com/casbin/casbin/v2/util"
 )
 
@@ -294,22 +295,32 @@ func (e *Enforcer) GetImplicitPermissionsForUser(user string, domain ...string) 
 // GetImplicitPermissionsForUser("alice") can only get: [["admin", "data1", "read"]], whose policy is default policy "p"
 // But you can specify the named policy "p2" to get: [["admin", "create"]] by    GetNamedImplicitPermissionsForUser("p2","alice")
 func (e *Enforcer) GetNamedImplicitPermissionsForUser(ptype string, user string, domain ...string) ([][]string, error) {
-	roles, err := e.GetImplicitRolesForUser(user, domain...)
-	if err != nil {
-		return nil, err
+	permission := make([][]string, 0)
+	rm := e.GetRoleManager()
+	domainIndex, _ := e.GetFieldIndex(ptype, constant.DomainIndex)
+	for _, rule := range e.model["p"][ptype].Policy {
+		if len(domain) == 0 {
+			matched, _ := rm.HasLink(user, rule[0])
+			if matched {
+				permission = append(permission, deepCopyPolicy(rule))
+			}
+		} else {
+			for _, d := range domain {
+				matched := rm.(*defaultrolemanager.RoleManager).Match(d, rule[domainIndex])
+				if !matched {
+					continue
+				}
+				matched, _ = rm.HasLink(user, rule[0], d)
+				if matched {
+					newRule := deepCopyPolicy(rule)
+					newRule[domainIndex] = d
+					permission = append(permission, newRule)
+					break
+				}
+			}
+		}
 	}
-
-	roles = append([]string{user}, roles...)
-
-	var res [][]string
-	var permissions [][]string
-	for _, role := range roles {
-		permissions = e.GetNamedPermissionsForUser(ptype, role, domain...)
-
-		res = append(res, permissions...)
-	}
-
-	return res, nil
+	return permission, nil
 }
 
 // GetImplicitUsersForPermission gets implicit users for a permission.
@@ -396,4 +407,11 @@ func (e *Enforcer) GetImplicitResourcesForUser(user string, domain ...string) ([
 		res = append(res, resLocal...)
 	}
 	return res, nil
+}
+
+// deepCopyPolicy returns a deepcopy version of the policy to prevent changing policies through returned slice
+func deepCopyPolicy(src []string) []string {
+	newRule := make([]string, len(src))
+	copy(newRule, src)
+	return newRule
 }
--- a/rbac_api_test.go
+++ b/rbac_api_test.go
@@ -314,9 +314,9 @@ func TestImplicitRoleAPI(t *testing.T) {
 	testGetRoles(t, e, []string{"/book/1/2/3/4/5", "pen_admin"}, "cathy")
 }
 
-func testGetImplicitPermissions(t *testing.T, e *Enforcer, name string, res [][]string) {
+func testGetImplicitPermissions(t *testing.T, e *Enforcer, name string, res [][]string, domain ...string) {
 	t.Helper()
-	myRes, _ := e.GetImplicitPermissionsForUser(name)
+	myRes, _ := e.GetImplicitPermissionsForUser(name, domain...)
 	t.Log("Implicit permissions for ", name, ": ", myRes)
 
 	if !util.Set2DEquals(res, myRes) {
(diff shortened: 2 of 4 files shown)


tangyang9464:
@Abingcbc I have an idea, why not just use ``rm.HasLink`` to get permissions directly, because ``GetFilteredPolicies`` only works without pattern. This PR solves Domain Match, how to solve RBAC pattern? If use ``rm.HasLink``, then ``GetImplicitPermissionsForUser`` and ``GetPermissionsForUser`` can be combined into the same method. Code draft
```
func (e *Enforcer) GetPermissionsForUser(user string, domain ...string) [][]string {
	permission := make([][]string, 0)
	rm := e.GetRoleManager()
	//
		add domain match here
	//
	for _, rule := range e.model["p"]["p"].Policy {
		matched, _ := rm.HasLink(user, rule[0], domain...)
		if matched {
			permission = append(permission, rule)
		}
	}
	return permission
}
```

Abingcbc:
@tangyang9464 Thx for your idea! I will refactor this PR.

BTW what do you mean by

> GetImplicitPermissionsForUser and GetPermissionsForUser can be combined into the same method.

tangyang9464:
> @tangyang9464 Thx for your idea! I will refactor this PR.
> 
> BTW what do you mean by
> 
> > GetImplicitPermissionsForUser and GetPermissionsForUser can be combined into the same method.

@Abingcbc I wanted to combine the two methods before, but now I think it's inappropriate, please ignore.

Abingcbc:
@tangyang9464 PTAL

hsluoyz:
@sasakiyori @tangyang9464 plz review

rbac_api.go
```diff
@@ -276,22 +277,33 @@ func (e *Enforcer) GetImplicitPermissionsForUser(user string, domain ...string)
...
+		for _, d := range domain {
+			matched = rm.(*defaultrolemanager.RoleManager).Match(d, rule[domainIndex])
+			if matched {
+				break
+			}
+		}
+		if !matched {
+			continue
+		}
+		matched, _ = rm.HasLink(user, rule[0], domain...)
```
tangyang9464, line 294: ``rm.HasLink`` only support one domain. So maybe you should put it in the above loop
```
for _, d := range domain {
	matched = rm.(*defaultrolemanager.RoleManager).Match(d, rule[domainIndex])
	if matched {
		matched, _ = rm.HasLink(user, rule[0], domain...)
	}
}
```

rbac_api_test.go
```diff
@@ -353,10 +353,16 @@ func TestImplicitPermissionAPI(t *testing.T) {
 	testGetImplicitPermissions(t, e, "alice", [][]string{{"alice", "data1", "read"}, {"data1_admin", "data1", "read"}, {"data1_admin", "data1", "write"}, {"data2_admin", "data2", "read"}, {"data2_admin", "data2", "write"}})
 	testGetImplicitPermissions(t, e, "bob", [][]string{{"bob", "data2", "write"}})
 
+	e, _ = NewEnforcer("examples/rbac_with_domain_pattern_model.conf", "examples/rbac_with_domain_pattern_policy.csv")
+	e.AddNamedDomainMatchingFunc("g", "KeyMatch", util.KeyMatch)
+
+	testGetImplicitPermissions(t, e, "admin", [][]string{{"admin", "domain1", "data1", "read"}, {"admin", "domain1", "data1", "write"}, {"admin", "domain1", "data3", "read"}}, "domain1")
```
tangyang9464, line 359: Add tests for multiple domains

tangyang9464 (changes requested):

rbac_api.go
Abingcbc: done

rbac_api_test.go
Abingcbc: done

Abingcbc:
@tangyang9464 plz review

tangyang9464 (changes requested):

rbac_api.go
```diff
@@ -294,22 +295,34 @@ func (e *Enforcer) GetImplicitPermissionsForUser(user string, domain ...string)
...
+	permission := make([][]string, 0)
+	rm := e.GetRoleManager()
+	domainIndex, _ := e.GetFieldIndex(ptype, constant.DomainIndex)
+	for _, rule := range e.model["p"][ptype].Policy {
+		if len(domain) == 0 {
+			matched, _ := rm.HasLink(user, rule[0])
+			if matched {
+				// Copy rule to prevent changing policies through returned slice
+				newRule := make([]string, len(rule))
+				copy(newRule, rule)
```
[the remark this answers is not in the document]
Abingcbc: Yep, the `rule` is a slice. If users update the value in returned slice, the policy will also be changed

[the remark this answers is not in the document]
Abingcbc: Maybe we should also check if this bug happens anywhere else?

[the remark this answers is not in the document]
Abingcbc: ok :)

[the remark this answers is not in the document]
Abingcbc: I will fix this issue in a new PR

Abingcbc:
@tangyang9464 @hsluoyz  plz review

tangyang9464 (approved):
LGTM

JalinWang:
@Abingcbc @tangyang9464  Hi, I don't think it has solved this problem totally. 

policy
```
g, alice, admin, *
g, bob, admin, domain2
```

test code
```go
	e, _ = NewEnforcer("examples/rbac_with_domain_pattern_model.conf", "examples/rbac_with_domain_pattern_policy.csv")
	e.AddNamedDomainMatchingFunc("g", "KeyMatch", util.KeyMatch)

	res, _ := e.GetRolesForUser("bob", "*")
	t.Log(res)

	res, _ = e.GetRolesForUser("bob", "domain1")
	t.Log(res)

	res, _ = e.GetRolesForUser("bob", "domain2")
	t.Log(res)
```

result:
image

expected result:
`res, _ := e.GetRolesForUser("bob", "*")` returns `[admin]` (for `*` matches `domain2` and `bob` is `admin` in `domain2`)

rbac_api.go
```diff
@@ -294,22 +295,32 @@ func (e *Enforcer) GetImplicitPermissionsForUser(user string, domain ...string)
...
+	domainIndex, _ := e.GetFieldIndex(ptype, constant.DomainIndex)
+	for _, rule := range e.model["p"][ptype].Policy {
+		if len(domain) == 0 {
+			matched, _ := rm.HasLink(user, rule[0])
+			if matched {
+				permission = append(permission, deepCopyPolicy(rule))
+			}
+		} else {
+			for _, d := range domain {
+				matched := rm.(*defaultrolemanager.RoleManager).Match(d, rule[domainIndex])
```
JalinWang, line 309: What's more, we can't assume the `RoleManager` is the default one.

JalinWang:
I think we should modify the code of `DomainManager`.
One elegant way is to let `getRoleManager` can return all the role managers that match the domain, e.g., "*" return "domain" and "domain2", while now it only returns one.

```go
// load or create a RoleManager instance of domain
func (dm *DomainManager) getRoleManager(domain string, store bool) *RoleManagerImpl {
	var rm *RoleManagerImpl
	var ok bool

	if rm, ok = dm.load(domain); !ok {
		rm = newRoleManagerWithMatchingFunc(dm.maxHierarchyLevel, dm.matchingFunc)
		if store {
			dm.rmMap.Store(domain, rm)
		}
		if dm.domainMatchingFunc != nil {
			dm.rmMap.Range(func(key, value interface{}) bool {
				domain2 := key.(string)
				rm2 := value.(*RoleManagerImpl)
				if domain != domain2 && dm.Match(domain, domain2) {
					rm.copyFrom(rm2)
				}
				return true
			})
		}
	}
	return rm
}
```

Abingcbc:
> policy
> 
> ```
> g, alice, admin, *
> g, bob, admin, domain2
> ```
> 
> test code
> 
> ```go
> 	e, _ = NewEnforcer("examples/rbac_with_domain_pattern_model.conf", "examples/rbac_with_domain_pattern_policy.csv")
> 	e.AddNamedDomainMatchingFunc("g", "KeyMatch", util.KeyMatch)
> 
> 	res, _ := e.GetRolesForUser("bob", "*")
> 	t.Log(res)
> 
> 	res, _ = e.GetRolesForUser("bob", "domain1")
> 	t.Log(res)
> 
> 	res, _ = e.GetRolesForUser("bob", "domain2")
> 	t.Log(res)
> ```
> 
> result: image
> 
> expected result: `res, _ := e.GetRolesForUser("bob", "*")` returns `[admin]` (for `*` matches `domain2` and `bob` is `admin` in `domain2`)

Sure, this PR only adds domain matching when getting permissions. And `GetImplicitPermission` doesn't use `GetRolesForUser`.

rbac_api.go
Abingcbc: Yep, we may refactor from the first imported dependency
https://github.com/casbin/casbin/blob/9b434269935411da709f6c85a6b59e9ce2401e4c/enforcer.go#L730-L745

An interface may be better than struct.

Issue #993, opened 2022-04-21
Why semantic-pr check is gone in every PR?


hsluoyz:
See: https://github.com/casbin/casbin/pull/992 and https://github.com/casbin/casnode/pull/497

But previously like this PR: https://github.com/casbin/casbin/pull/941 , we have it:

![image](https://user-images.githubusercontent.com/3787410/164467514-d6490b6b-6af7-4703-a356-6401e05251bb.png)

Saw this issue: https://github.com/zeke/semantic-pull-requests/issues/183

Can anyone fix this? We need to fix it in all Casbin orgs.

hsluoyz:
@tangyang9464

hsluoyz:
@JalinWang this issue has been here for more than 1 month. Do we have a good fix now?

/cc @tangyang9464

JalinWang:
@hsluoyz 
I find an alternative https://github.com/apps/semantic-prs
FYI: https://github.com/zeke/semantic-pull-requests/issues/183#issuecomment-1120327873

Screenshot on my test repo:
![image](https://user-images.githubusercontent.com/20621784/171627761-679f50d4-3b2e-407c-b741-83ee3dd2c0c6.png)

hsluoyz:
Fixed by solution: https://github.com/zeke/semantic-pull-requests/issues/183#issuecomment-1120327873

PR #994, opened 2022-04-21
fix: use semantic-pull-request by GitHub Action

tangyang9464:

fix: #993


tangyang9464:
> @tangyang9464 so fast PR! Do we need to update all repos for this?

@hsluoyz According to https://github.com/zeke/semantic-pull-requests/issues/183#issuecomment-1104408204, if the author can get Heroku support, maybe it can continue to use without fix. Looking forward to the author's follow-up work

nodece:
Use `chore` instead of `fix`.

PR #1008, opened 2022-05-11
fix: UpdateFilteredPolicies returns the oldPolicy containing ptype

tangyang9464:

fix: [xorm-adapter#49](https://github.com/casbin/xorm-adapter/issues/49)

### DIFF
--- a/internal_api.go
+++ b/internal_api.go
@@ -330,6 +330,12 @@ func (e *Enforcer) updateFilteredPolicies(sec string, ptype string, newRules [][
 				return false, err
 			}
 		}
+		// For compatibility, because some adapters return oldRules containing ptype, see https://github.com/casbin/xorm-adapter/issues/49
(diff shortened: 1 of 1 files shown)


tangyang9464:
> @tangyang9464 can we add a test for this?

@hsluoyz This requires the support of third-party adapter, because ``casbin`` built-in file-adapter doesn't support ``UpdateFilteredPolicies``. If adding tests, then we need to import additional third-party dependencies such as ``xorm-adapter`` in ``casbin`` and only use it in test. Is this acceptable?

tangyang9464:
@hsluoyz In fact, the adapter instead of casbin should be modified for this problem, but because there are many adapters, I choose to modify casbin to be compatible

hsluoyz:
@tangyang9464 ok, let's keep it this way first

Issue #1011, opened 2022-05-13
[Bug] Duplicate adapter entries


abichinger:
**Want to prioritize this issue? Try:**

[![issuehunt-to-marktext](https://github.com/BoostIO/issuehunt-materials/raw/master/v1/issuehunt-button-v1.svg)](https://issuehunt.io/r/casbin/casbin)

------

**Describe the bug**
Duplicate rules inside the adapter will cause duplicate rules inside the policy.
Although duplicate entries should not occur, when the API of casbin is used, I think we should implement a consistent behavior between `persist.LoadPolicyArray` and `e.AddPolicy`

**To Reproduce**
```go
package casbin

import (
	"testing"

	"github.com/casbin/casbin/v2/model"
	"github.com/casbin/casbin/v2/persist"
)

func testRuleCount(t *testing.T, model model.Model, expected int, sec string, ptype string, tag string) {
	t.Helper()

	ruleCount := len(model[sec][ptype].Policy)
	if ruleCount != expected {
		t.Errorf("[%s] rule count: %d, expected %d", tag, ruleCount, expected)
	}
}

func TestDuplicatePolicyInAdapter(t *testing.T) {
	e, _ := NewEnforcer("examples/basic_model.conf")

	e.AddPolicy("alice", "data1", "read")
	e.AddPolicy("alice", "data1", "read")

	testRuleCount(t, e.GetModel(), 1, "p", "p", "AddPolicy")

	e.ClearPolicy()

	//simulate adapter.LoadPolicy with duplicate rules
	persist.LoadPolicyArray([]string{"p", "alice", "data1", "read"}, e.model)
	persist.LoadPolicyArray([]string{"p", "alice", "data1", "read"}, e.model)

	testRuleCount(t, e.GetModel(), 1, "p", "p", "LoadPolicyArray")
}
```

Output: 
```
--- FAIL: TestDuplicatePolicyInAdapter (0.00s)
   test.go:47: [LoadPolicyArray] rule count: 2, expected 1
FAIL
FAIL	github.com/casbin/casbin/v2	0.003s
FAIL
```

issue origin: https://github.com/casbin/pycasbin/issues/258

PR #1012, opened 2022-05-13
fix: handle duplicate adapter entries

abichinger:
fix: #1011 

Signed-off-by: Andreas Bichinger <andreas.bichinger@gmail.com>

### DIFF
--- a/persist/persist_test.go
+++ b/persist/persist_test.go
@@ -1,9 +1,53 @@
-package persist
+// Copyright 2017 The casbin Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


tangyang9464 (approved):
lgtm

persist/persist_test.go
```diff
@@ -1,9 +1,39 @@
-package persist
+package persist_test
```
hsluoyz, line 1: why change pkg name?

Also no Apache license header

hsluoyz (changes requested):

persist/persist_test.go
abichinger: to avoid an import cycle

hsluoyz (approved):

Issue #1016, opened 2022-05-20
[Feature] LoadPolicyLine should check the policy length whether match the model policy definition


qingwave:
**Want to prioritize this issue? Try:**

[![issuehunt-to-marktext](https://github.com/BoostIO/issuehunt-materials/raw/master/v1/issuehunt-button-v1.svg)](https://issuehunt.io/r/casbin/casbin)

------

**Is your feature request related to a problem? Please describe.**
When call `LoadPolicyLine` mothed to load policy from adapter(file-adapter), if some policies is invaild (length not matched the parameters in policy definition), we should return a error or skip. 
https://github.com/casbin/casbin/blob/290ed05ab3ab406e74d2538830410ca2d826a0eb/persist/adapter.go#L25

If not, a invaild policy would cause unexcepted error from Enforce result

**Describe the solution you'd like**
Skip invaild policy or return a error in function `LoadPolicyLine`

**Describe alternatives you've considered**

**Additional context**

hsluoyz:
@tangyang9464

tangyang9464:
@qingwave The policy length is checked when enforce is called, but it would be better to add the check in LoadPolicyLine.

qingwave:
> @qingwave The policy length is checked when enforce is called, but it would be better to add the check in LoadPolicyLine.

@tangyang9464 Thanks for your information, that's means an invaild policy couldn't affect the enforce result, it's make sense.

PR #1019, opened 2022-05-27
feat: add policy length check in LoadPolicyLine

tangyang9464:
Fix: https://github.com/casbin/casbin/issues/1016

### DIFF
--- a/persist/adapter.go
+++ b/persist/adapter.go
@@ -33,7 +33,17 @@ func LoadPolicyLine(line string, m model.Model) {
 	r.TrimLeadingSpace = true
 
 	tokens, err := r.Read()
-	if err != nil {
+	if err != nil || len(tokens) <= 0 {
+		return
+	}
+
+	ptype := tokens[0]
+	sec := ptype[:1]
+	if _, ok := m[sec][ptype]; !ok {
+		return
+	}
+	expectLen := len(tokens) - 1
+	if (sec == "g" && strings.Count(m[sec][ptype].Value, "_") != expectLen) || (sec == "p" && len(m[sec][ptype].Tokens) != expectLen) {
 		return
 	}
 


hsluoyz:
@tangyang9464 @qingwave what's the benchmark downgrade? We prefer to add checks to compile-time instead of run-time mostly for performance reason.

tangyang9464:
> @tangyang9464 @qingwave what's the benchmark downgrade? We prefer to add checks to compile-time instead of run-time mostly for performance reason.

@hsluoyz Maybe it's not a good idea. The length has been checked elsewhere. If the performance is degraded, then this PR is useless

tangyang9464:
@hsluoyz The current implementation is that invalid policy may exist when loading the model, and the length will be checked only when the policy is used, such as the enforce method. So it looks more efficient than checking from the beginning of the loading policy

hsluoyz:
@tangyang9464 I agree, we don't need to add duplicated checks. Then we don't need this.

sagilio:
@tangyang9464 @hsluoyz  Do we have any existing tests to see the rate of performance degradation because of this PR?
I think if we check the length in any edit policy API, we will need not check it at `Enforce` func. 
And If add a check in the right place, there will have almost no impact on performance. like here:

```go
// HasPolicy determines whether a model has the specified policy rule.
func (model Model) HasPolicy(sec string, ptype string, rule []string) bool {
	assertion := model[sec][ptype]
	if len(assertion.Tokens) != len(rule) {
		return false
	}
	_, ok := assertion.PolicyMap[strings.Join(rule, DefaultSep)]
	return ok
}
```

hsluoyz:
@sagilio see https://casbin.org/docs/en/benchmark and xxx_b_test.go at: https://github.com/casbin/casbin#benchmarks

sagilio:
@hsluoyz I see, but the changes in this PR don't seem to be covered by any of the tests in `xxx_b_test.go`,
and the benchmark result in each CI run may also max deviation of about 20%. 
The result in the docs seems a very old version.

hsluoyz:
@sagilio then we need to add more tests to cover this case. And we can run multiple rounds and use average number to reduce the randomness. @JalinWang actually we want to build a new benchmark system to list all Casbin related benchmark data there.

sagilio:
@hsluoyz The reason for the deviation of each ci test is that the environment(CPU) is not fixed. There is only a little difference between the same environments.

Casbin.NET's result will output the environment information.
image
image

hsluoyz:
@sagilio good point. We will build our own CI anyway.

hsluoyz:
We don't do a lot of input checks to sacrifice the performance. The user is responsible for input validity.

PR #1020, opened 2022-05-29
feat: add HasPolicyEx() API to determine the policy exist with an error 

sagilio:
Fixes: #1016 

Next, we may replace the existing `HasPolicy` call with this and remove check at `Enforce` func.

### DIFF
--- a/model/model.go
+++ b/model/model.go
@@ -71,6 +71,8 @@ func (model Model) AddDef(sec string, key string, value string) bool {
 		for i := range ast.Tokens {
 			ast.Tokens[i] = key + "_" + strings.TrimSpace(ast.Tokens[i])
 		}
+	} else if sec == "g" {
+		ast.Tokens = strings.Split(ast.Value, ",")
 	} else {
 		ast.Value = util.RemoveComments(util.EscapeAssertion(ast.Value))
 	}
--- a/model/policy.go
+++ b/model/policy.go
@@ -15,6 +15,7 @@
 package model
 
 import (
+	"fmt"
 	"strconv"
 	"strings"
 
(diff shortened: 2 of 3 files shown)


model/model.go
```diff
@@ -71,6 +71,8 @@ func (model Model) AddDef(sec string, key string, value string) bool {
 		for i := range ast.Tokens {
 			ast.Tokens[i] = key + "_" + strings.TrimSpace(ast.Tokens[i])
 		}
+	} else if sec == "g" {
+		ast.Tokens = strings.Split(ast.Value, ",")
```
hsluoyz, line 75: What's this? If this is fixing a bug, plz raise another PR

model/policy.go
```diff
@@ -125,6 +125,23 @@ func (model Model) GetFilteredPolicy(sec string, ptype string, fieldIndex int, f
 	return res
 }
 
+// CheckPolicy determines whether the specified policy rule can be added to model.
+func (model Model) CheckPolicy(sec string, ptype string, rule []string) bool {
```
hsluoyz, line 129: Rename to `HasPolicyEx`, don't revert the return value

hsluoyz (changes requested):

model/model.go
sagilio: Now, `ast.Tokens` will be set nil when the section is `g`, so the `len(ast.Tokens)` will always return 0. 
We can not check the length by `len(rule) < len(ast.Tokens)` if not change here.

In another way, We can also add a `tlen`/`tokenLen` field to record the token len when loading the model.

model/policy.go
sagilio: If using `HasPolicyEx`, It is best to return `false, err` when the length is invalid, instead of `true`.

model/model.go
hsluoyz: got it.

hsluoyz (approved):

Issue #1027, opened 2022-06-07
[Bug] - Domain manager rebuild() infinite loop?


kizjig:
[<img alt="Issuehunt badges" src="https://img.shields.io/badge/IssueHunt-%24200%20Funded-%2300A156.svg" />](https://issuehunt.io/r/casbin/casbin/issues/1027)

**Want to prioritise this issue? Try:**

[![issuehunt-to-marktext](https://github.com/BoostIO/issuehunt-materials/raw/master/v1/issuehunt-button-v1.svg)](https://issuehunt.io/r/casbin/casbin)

------

**Describe the bug**
When adding a domain matching function to the default RoleManager, it triggers a rebuild call of the internal DomainManager, this function never seems to finishing executing on a large dataset. Internally it seems to be trying to "AddLinks" forever.

** Model - HRBAC with ABAC**
```
[request_definition]
r = sub, tenant, perm, act, eft

[policy_definition]
p = sub, tenant, perm, act, eft

[role_definition]
g = _, _, _

[policy_effect]
e = some(where (p.eft == allow)) && !some(where (p.eft == deny))

[matchers]
m = (g(r.sub, p.sub, r.tenant) && keyMatch4(r.tenant, p.tenant) && r.perm == p.perm && regexMatch(r.act, p.act))
```

** Policy data **
Here's a small sample of some of the policy data showing permissions with actions grouped under roles. And wildcarding on the `role_superuser1` grouping so it can work across X tenants. **Typically the policy dataset is ~45000 database rows.**

For example user id 7073 is a super user and can do anything across any tenant. Where as user 65716 is just a creator for tenant 753.
```
p,role_superuser1,tenant/{id},tenant,(^view$)|(^manage$),allow,
p,role_superuser1,tenant/{id},dashboard,login,allow,
p,role_superuser1,tenant/{id},model_upload_max_size,large,allow,
p,role_superuser1,tenant/{id},presentation,(^local$)|(^remote$),allow,
p,role_superuser1,tenant/{id},invite,(^create$)|(^delete$)|(^edit$)|(^view$),allow,
p,role_superuser1,tenant/{id},library,(^create$)|(^edit$)|(^view$)|(^delete$),allow,
p,role_superuser1,tenant/{id},desktop,login,allow,
p,role_presenter1,tenant/{id},dashboard,login,deny,
p,role_presenter1,tenant/{id},tenant,view,allow,
p,role_creator1,tenant/{id},dashboard,login,allow,
[...]

PR #1029, opened 2022-06-07
feat: add IPatternMatcher

abichinger:
fix: #1027 

## Changelog
- add interface IPatternMatcher
- add function SetNamedRoleMatcher
- add function SetNamedDomainMatcher
- mark AddNamedMatchingFunc and AddNamedDomainMatchingFunc as deprecated (these function should be removed with the next major release)

@kizjig please review this changes.
`roleManager.AddDomainMatchingFunc("KeyMatch", util.KeyMatch)` needs to be replaced with: 
```go
domainMatcher := defaultrolemanager.NewPatternMatcher(util.IsKeyMatch4Pattern, util.KeyMatch4)
roleManager.SetDomainMatcher(domainMatcher)
```

## Benchmarks
```
before:
BenchmarkBuildRoleLinksWithPatternLarge-12                             1        11420789697 ns/op       5610426576 B/op 63632133 allocs/op
BenchmarkBuildRoleLinksWithDomainPatternLarge-12                      70          16872961 ns/op         4251498 B/op      86735 allocs/op
BenchmarkBuildRoleLinksWithPatternAndDomainPatternLarge-12             1        11617835457 ns/op       5619315344 B/op 63660158 allocs/op

after:
BenchmarkBuildRoleLinksWithPatternLarge-12                           135           8714727 ns/op         2660236 B/op      62784 allocs/op
BenchmarkBuildRoleLinksWithDomainPatternLarge-12                     142           8885176 ns/op         2848862 B/op      62765 allocs/op
BenchmarkBuildRoleLinksWithPatternAndDomainPatternLarge-12           130           9349494 ns/op         2896310 B/op      65735 allocs/op
```

Signed-off-by: Andreas Bichinger <andreas.bichinger@gmail.com>


sagilio (approved):
LGTM, the new matcher uses `IsPattern` to pre-check the tokens whether needed to call the Match function to avoid calling duplicate and useless match functions and storing useless match results.

rbac/default-role-manager/role_manager.go
```diff
@@ -185,9 +225,9 @@ func NewRoleManagerImpl(maxHierarchyLevel int) *RoleManagerImpl {
 }
 
 // use this constructor to avoid rebuild of AddMatchingFunc
```
tangyang9464, line 227: AddMatchingFunc should be change

role_manager_b_test.go
```diff
@@ -140,8 +143,10 @@ func BenchmarkBuildRoleLinksWithDomainPatternLarge(b *testing.B) {
 
 func BenchmarkBuildRoleLinksWithPatternAndDomainPatternLarge(b *testing.B) {
 	e, _ := NewEnforcer("examples/performance/rbac_with_pattern_large_scale_model.conf", "examples/performance/rbac_with_pattern_large_scale_policy.csv")
-	e.AddNamedMatchingFunc("g", "", util.KeyMatch4)
-	e.AddNamedDomainMatchingFunc("g", "", util.KeyMatch4)
+	roleMatcher := defaultrolemanager.NewPatternMatcher(util.IsKeyMatch4Pattern, util.KeyMatch4)
+	domainMatcher := defaultrolemanager.NewPatternMatcher(util.IsKeyMatch4Pattern, util.KeyMatch4)
+	e.SetNamedRoleMatcher("g", roleMatcher)
```
tangyang9464, line 148: Can we merge these two steps into one API? Like ``e.AddNamedRole/DomainMatcher(util.IsKeyMatch4Pattern, util.KeyMatch4)``？

sagilio: For the accuracy of the function name, I think the two steps are clearer.

tangyang9464: I just think a simple API like AddMatchingFunc should be provided, and users might not have to explicitly call NewPatternMatcher. But these two steps are ok. We should provide the default PatternMatcher when creating RoleManger. Because I have seen too many users use pattern directly without adding Match Func. Even if it is stated in the document

tangyang9464 (approved):
lgtm

kizjig:
Thank you for these changes.  Looks good to me too.

I have a related question, forgive me if I'm misunderstanding something here btw... If we create a role/domain manager with a max hierarchy of less than the default 10, for example 4. How is that value preserved when the enforcer has hardcoded calls to recreate the role manager with functions like this?

```
func (e *Enforcer) initRmMap() {
	for ptype := range e.model["g"] {
		if rm, ok := e.rmMap[ptype]; ok {
			_ = rm.Clear()
		} else {
			e.rmMap[ptype] = defaultrolemanager.NewRoleManager(10)
		}
	}
}```

**Comment by kizjig** [comment]
And as a follow up clarification question...

What's the best flow or best time to customise/recreate the role manager in terms of setting domain matching functions and max hierarchy level?

Should it be once we've created the enforcer with our model but before we've done the first initial load of the policy data?

**Comment by tangyang9464** [comment]
@kizjig In fact, you must call ``e.SetRoleManager`` (need to customize RoleManger, including setting max hierarchy level and domain matching functions) after ``NewEnforcer``,  then reload the policy (``e.LoadPolicy``). Code like
```
e, _ := NewEnforcer("","")

roleManager := NewRoleManager(4)

domainMatcher := defaultrolemanager.NewPatternMatcher(util.IsKeyMatch4Pattern, util.KeyMatch4)
roleManager.SetDomainMatcher(domainMatcher)

e.SetRoleManager(roleManager)
e.LoadPolicy()
```

**Comment by kizjig** [comment]
> @kizjig In fact, you must call `e.SetRoleManager` (need to customize RoleManger, including setting max hierarchy level and domain matching functions) after `NewEnforcer`, then rebuild role links(`e.BuildRoleLinks`). Code like
> 
> ```
> e, _ := NewEnforcer("","")
> 
> roleManager := NewRoleManager(4)
> 
> domainMatcher := defaultrolemanager.NewPatternMatcher(util.IsKeyMatch4Pattern, util.KeyMatch4)
> roleManager.SetDomainMatcher(domainMatcher)
> 
> e.SetRoleManager(roleManager)
> e.BuildRoleLinks()
> ```

Thanks. So if I take that approach in your code above, and then afterwards load policy data from disk/file, it will automatically build the role links for the incoming data  right (assuming `e.EnableAutoBuildRoleLinks(true)` is already set)?

I'm trying to avoid loading/processing the data twice. So I see the flow working as something like...
1. create enforcer
2. customise new role manager and domain matching.
3. set new role manager on enforcer
4. Set `e.EnableAutoBuildRoleLinks(true)`
5. load policy data from disk into enforcer internal memory model.

Does that sound correct?

**Comment by tangyang9464** [comment]
@kizjig do like this
```
// 1. don't set adapter first to avoid calling LoadPolicy automatically.
e, _ := NewEnforcer(model)

// 2. customise new role manager and domain matching.
roleManager := NewRoleManager(4)
domainMatcher := defaultrolemanager.NewPatternMatcher(util.IsKeyMatch4Pattern, util.KeyMatch4)
roleManager.SetDomainMatcher(domainMatcher)

// 3. set new role manager and adapter on enforcer
e.SetRoleManager(roleManager)
e.SetAdapter(adapter)

// 4. load policy
e.LoadPolicy()
```
Unfortunately, autoBuildRoleLinks is set to true by default when the enforcer is created, so e.EnableAutoBuildRoleLinks(true) has no effect. Maybe we will add a NewEnforcerWithOutBuild API to set autoBuildRoleLinks when creating an enforcer. So for now, we can only do it like the above

**Comment by kizjig** [comment]
Yes. thanks. That's what I had in mind. I am glad I was on the same page.

File: role_manager_b_test.go (hunk 2 above)
Comment (abichinger) [reply role_manager_b_test.go:148]: > Can we merge these two steps into one API?

We could provide some general instances of `PatternMatcher`.
Then the user can use one of the predefined `PatternMatcher` instances or create one himself.

> We should provide the default PatternMatcher when creating RoleManger. Because I have seen too many users use pattern directly without adding Match Func. Even if it is stated in the document

If no `PatternMatcher` is set the performance should be slightly better, because `isPattern` never gets called.
Besides, I think it is not clear, what we should use as default `PatternMatcher`.

Comment (tangyang9464) [reply role_manager_b_test.go:148]: > If no `PatternMatcher` is set the performance should be slightly better, because `isPattern` never gets called. Besides, I think it is not clear, what we should use as default `PatternMatcher`.

I think default `PatternMatcher` only needs to satisfy the most basic  pattern functions. Such as ``KeyMatch4`` in the above example. But that doesn't matter, what matters is that it really shouldn't provide a default if it's more efficient without `PatternMatcher`.
So just follow your PR exactly without any changes!

**Comment by kizjig** [comment]
Can we get this change progressed and merged in soon please?

**Comment by hsluoyz** [comment]
@abichinger `SetNamedRoleMatcher` should be renamed to `SetNamedRoleMatchingFunc`. `Matcher` is a reserved keyword to specify the matcher (m = xxxx part) in PERM metamodel. It should not be pointed to anything else in Casbin. Actually in Casbin, anything called `XXXer` should be a complete and non-trivial component, like adapter, watcher, etc. A matching function is quite small and should not be named after `XXXer`.

**Comment by abichinger** [comment]
@hsluoyz renaming `SetNamedRoleMatcher` to `SetNamedRoleMatchingFunc` and removing the reserved word `Matcher` would probably lead to the following function definition: (first solution)
```go
//drm is defaultrolemanager
func (e *Enforcer) SetNamedRoleMatchingFunc(ptype string, isMatch drm.MatchingFunc, isPattern drm.IsPatternFunc) bool
```
`IPatternMatcher` would be split into `MatchingFunc` and `IsPatternFunc`
This would lead to minor disadvantages over the following solution:
- we could remove redundant code if we move the cache from `DomainManager` and `RoleManagerImpl` to `IPatternMatcher`
- no default implementations of `IPatternMatcher` like `PrefixMatcher`

How about I do the following renames: (second solution)
- rename `SetNamedRoleMatcher` to `SetPatternFuncsForRoles`
- rename `SetNamedDomainMatcher` to `SetPatternFuncsForDomains`
- rename `IPatternMatcher` to `IPatternFuncs`
- rename `PrefixMatcher` to `PatternFuncsWithPrefix`

I would prefer the second solution.
Please let me know which solution you would like me to implement.

timabilov:
why this one is not merged yet?

Issue #1028, opened 2022-06-07
[Feature] Matcher expressions are always re-evaluated, even when they don't change


timconinx:
(I'm using the feature template because I'm not sure if this is a bug or an unfortunate feature)

We've started an internal discussion on whether to replace our home-cooked access framework with casbin. First experiments point to casbin being far inferior in terms of performance.

**Is your feature request related to a problem? Please describe.**
Investigation showed that a lot of performance was lost in re-evaluating the Matcher expressions. Each call to `enforce` will trigger `govaluate.NewEvaluableExpressionWithFunctions(expString, functions)` in line 544 of `/enforcer.go`

**Describe the solution you'd like**
As experiment, we tried to cache the matcher expression, by first adding two field to the `Enforcer` struct
```
expression *govaluate.EvaluableExpression
expressionBasedOn string
```
And then making the evalution of the matcher string dependent on the existance and equality of these fields
```
if e.expression == nil || e.expressionBasedOn != expString {
    expression, err = govaluate.NewEvaluableExpressionWithFunctions(expString, functions)
    if err != nil {
        return false, err
    }
    e.expression = expression
    e.expressionBasedOn = expString
} else {
    expression = e.expression
}
```

**Describe alternatives you've considered**
There are probably other or better ways to archieve this, but the question remains: did we find a bug, or is there a reason why the expression should not be cached?

**Additional context**
These are our benchmark numbers

![image](https://user-images.githubusercontent.com/1834193/172376610-d0554c06-2056-4e62-95ea-5b0d55e11b92.png)

The `CasbinR` numbers are those with the described code-change: roughly one fourth of the normal execution time...

rvdurm:
The Knetic/govaluate engine which is used specifies in its readme ...

Expressions are parsed once, and can be re-used multiple times. Parsing is the compute-intensive phase of the process, so if you intend to use the same expression with different parameters, just parse it once.

hsluoyz:
@rvdurm can you make a PR for it?

timconinx:
@hsluoyz You meant a PR for the code I provided, right?

hsluoyz:
@timconinx your solution is not beautiful. Why not use a map[string]Evaluation to store all evaluations? If the map is not hit, we will create a new Evaluation and store into the map.

abichinger:
I thought of an edge case, which is very unlikely, but I thought I should mention it.

A cached expression should be invalidated, if a function inside the `FunctionMap` is modified.
e.g.:
[...]

PR #1031, opened 2022-06-13
feat: caching the govaluate expressions

timconinx:
Trying to provide a solution for issue https://github.com/casbin/casbin/issues/1028 

All tests pass, but I'm not quite sure if every edge case is covered

Fix: https://github.com/casbin/casbin/issues/1028

### DIFF
--- a/enforcer.go
+++ b/enforcer.go
@@ -542,27 +542,14 @@ func (e *Enforcer) enforce(matcher string, explains *[]string, rvals ...interfac
 		pTokens: pTokens,
 	}
 
-	var expression *govaluate.EvaluableExpression
 	hasEval := util.HasEval(expString)
-
-	if !hasEval {
-		var cachedExpression, isPresent = e.matcherMap.Load(expString)
-		if !isPresent {
-			expression, err = govaluate.NewEvaluableExpressionWithFunctions(expString, functions)
-			if err != nil {
-				return false, err
-			}
-			e.matcherMap.Store(expString, expression)
-		} else {
-			expression = cachedExpression.(*govaluate.EvaluableExpression)
-		}
-	} else {
+	if hasEval {
 		functions["eval"] = generateEvalFunction(functions, &parameters)
-		expression, err = govaluate.NewEvaluableExpressionWithFunctions(expString, functions)
-		if err != nil {
-			return false, err
-		}
-		e.matcherMap.Store(expString, expression)
+	}
+	var expression *govaluate.EvaluableExpression
+	expression, err = e.getAndStoreMatcherExpression(hasEval, expString, functions)
+	if err != nil {
+		return false, err
 	}
 
 	if len(e.model["r"][rType].Tokens) != len(rvals) {
@@ -693,6 +680,23 @@ func (e *Enforcer) enforce(matcher string, explains *[]string, rvals ...interfac
 	e.logger.LogEnforce(expString, rvals, result, logExplains)
 
 	return result, nil
+}
+
+func (e *Enforcer) getAndStoreMatcherExpression(hasEval bool, expString string, functions map[string]govaluate.ExpressionFunction) (*govaluate.EvaluableExpression, error) {
+	var expression *govaluate.EvaluableExpression
+	var err error
+	var cachedExpression, isPresent = e.matcherMap.Load(expString)
+
+	if !hasEval && isPresent {
+		expression = cachedExpression.(*govaluate.EvaluableExpression)
+	} else {
+		expression, err = govaluate.NewEvaluableExpressionWithFunctions(expString, functions)
+		if err != nil {
+			return nil, err
+		}
+		e.matcherMap.Store(expString, expression)
+	}
+	return expression, nil
 }
 
 // Enforce decides whether a "subject" can access a "object" with the operation "action", input parameters are usually: (sub, obj, act).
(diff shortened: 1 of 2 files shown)


enforcer.go
```diff
@@ -38,10 +39,11 @@ type Enforcer struct {
...
 
-	adapter    persist.Adapter
-	watcher    persist.Watcher
-	dispatcher persist.Dispatcher
-	rmMap      map[string]rbac.RoleManager
+	adapter     persist.Adapter
+	watcher     persist.Watcher
+	dispatcher  persist.Dispatcher
+	rmMap       map[string]rbac.RoleManager
+	expressions sync.Map
```
tangyang9464, line 46: We should declare ``expressions`` as a variable, not a member of the enforcer structure

enforcer.go
```diff
@@ -537,13 +545,24 @@ func (e *Enforcer) enforce(matcher string, explains *[]string, rvals ...interfac
...
+		if !isPresent {
+			expression, err = govaluate.NewEvaluableExpressionWithFunctions(expString, functions)
+			if err != nil {
+				return false, err
+			}
+			e.expressions.Store(expString, expression)
+		} else {
+			expression = cachedExpression.(*govaluate.EvaluableExpression)
+		}
+	} else {
```
tangyang9464, line 559: I think the logic of this part of the code should be
1. If hasEval then generateEval function
2.if !Present then new expression
else use cache. 
If hasEval, it may still use cache.

tangyang9464 (changes requested):

enforcer.go
hsluoyz, line 46: rename to `matcherMap`

hsluoyz: @tangyang9464 you mean global var? Why not an enforcer member?

enforcer.go
```diff
@@ -471,6 +475,10 @@ func NewEnforceContext(suffix string) EnforceContext {
 	}
 }
 
+func (e *Enforcer) invalidateCache() {
```
hsluoyz, line 478: invalidateMatcherMap

enforcer.go
```diff
@@ -537,13 +545,24 @@ func (e *Enforcer) enforce(matcher string, explains *[]string, rvals ...interfac
 	var expression *govaluate.EvaluableExpression
 	hasEval := util.HasEval(expString)
 
-	if hasEval {
+	if !hasEval {
+		var cachedExpression, isPresent = e.expressions.Load(expString)
```
hsluoyz, line 549: This code logic is way too complicated when messed up together, can we extract a wrapper out which will use cache?

hsluoyz (changes requested):

enforcer.go
timconinx: @hsluoyz do you mean wrap the caching code in a separate func?

timconinx: Code cleanup done, and made the logic easier to understand

enforcer.go
timconinx: When using the cache with 'hasEval', a lot of tests fail... For now I'm restricting the cache to matchers without eval

enforcer.go
tangyang9464: I made a mistake. It's ok.

enforcer.go
tangyang9464: That's weird. Did you find out why?

timconinx: At first glance it looks like the cached key is the matcher-string with the eval not filled in, while the EvaluableExpression has the request parameters filled in. This means that a second call, with different request parameters gets a hit on the matcher string, but uses the Expression with the parameters of the previous request.
I'm not sure how to solve this, but I'm inclined to believe that eval matchers are "worth" to be regenerated into new Expressions each time, because of the added eval-complexity.

abichinger:
@timconinx Please make sure, that the PR title and commit messages follow the [Conventional Commits spec](https://www.conventionalcommits.org/en/v1.0.0/#summary)

Otherwise it looks good to me.

util/builtin_operators.go
```diff
@@ -331,7 +332,7 @@ func GlobMatchFunc(args ...interface{}) (interface{}, error) {
 
 // GenerateGFunction is the factory method of the g(_, _[, _]) function.
 func GenerateGFunction(rm rbac.RoleManager) govaluate.ExpressionFunction {
-	memorized := map[string]bool{}
+	memorized := sync.Map{}
```
thetorpedodog, line 335: This was something that got me when I worked in this function earlier—at the time, the `g` function was called by govaluate purely serially, within the context of exactly one enforcement call. Is this something that is changed by the new expression caching?

thetorpedodog (commented):
Hello! I recently did some work in this corner of the codebase and noticed something this change. I’m not a maintainer or anything, so please don’t read this as a binding request.

util/builtin_operators.go
hsluoyz: @timconinx

timconinx: @thetorpedodog 
I did change it, because the caching lead to concurrency problems in the `g` function. I discussed it in the related issue ([direct link](https://github.com/casbin/casbin/issues/1028#issuecomment-1153980413))

I'm not a real fan of replacing string bool maps by syncMaps, so if you see a better solution, then by all means

hsluoyz (approved):

Issue #1032, opened 2022-06-14
How can I save Role Manager Links in DB


shipovalovyuriy:
------

**What's your scenario? What do you want to achieve?**

I want to build multi-domain hierarchical tree with build-in Role manager via Links, but can't save result in DB and my Links doesn't work

```ini
[request_definition]
r = sub, dom, obj, act

[policy_definition]
p = sub, dom, obj, act

[role_definition]
g = _, _, _

[policy_effect]
e = some(where (p.eft == allow))

[matchers]
m = g(r.sub, p.sub, r.dom) && r.dom == p.dom && r.obj == p.obj && r.act == p.act

**Your policy:**

"p"	"admin"	"domain1"	"employee_list"	"read"
"p"	"superadmin"	"domain1"	"employee_list"	"write"

```

**Your request(s):**

```

	sub := "alice"         // the user that wants to access a resource.
	obj := "employee_list" // the resource that is going to be accessed.
	act := "read"          // the operation that the user performs on the resource.
	dom := "domain1"
        
        got false (role manager didnt work)

```
My code:
```
package main

import (
	"fmt"

	"github.com/casbin/casbin/v2"
	defaultrolemanager "github.com/casbin/casbin/v2/rbac/default-role-manager"
	_ "github.com/lib/pq"
	sqlxadapter "github.com/memwey/casbin-sqlx-adapter"
)

func main() {

	opts := &sqlxadapter.AdapterOptions{
		DriverName:     "postgres",
		DataSourceName: "host=localhost user=urban dbname=urban password=test sslmode=disable",
		TableName:      "casbin_rule",
	}
	r := defaultrolemanager.NewRoleManager(10)

	a := sqlxadapter.NewAdapterFromOptions(opts)

	e, err := casbin.NewEnforcer("./rbac.conf", a)

	e.SetRoleManager(r)
	e.LoadPolicy()

	if err != nil {
		fmt.Println(err)
	}

	e.SetRoleManager(r)

	e.LoadPolicy()

	r.AddLink("admin", "superadmin")
	r.AddLink("alice", "admin")

	e.BuildRoleLinks()

	if err != nil {
		fmt.Println("err")
	}

	sub := "alice"         // the user that wants to access a resource.
	obj := "employee_list" // the resource that is going to be accessed.
	act := "read"          // the operation that the user performs on the resource.
	dom := "domain1"

	if res, err := e.Enforce(sub, dom, obj, act); res {
		// permit alice to read data1
[...]

Issue #872, opened 2021-08-27
[Question] Hierarchical domains with (hierarchical) roles


Sefriol:
**What's your scenario? What do you want to achieve?**

Pretty much a follow-up from https://github.com/casbin/casbin/issues/718#issuecomment-889785940

> So, do you think this problem is solvable by Casbin? I think this is pretty typical use case of hierarchical RBAC where you do not want to define global role hierarchy into each domain.
> 
> General idea would be pretty similar to the original question:
> 
> - global domain
> - - subdomain1
> - - - lowersubdomain1
> - - subdomain2
> 
> 1. So if `alice` is an `admin` in `global domain`, she should be an `admin` in all domains below it (`subdomain1`, `lowersubdomain1`, `subdomain2`).
>     1. This should happen without any additional policy lines. `g, alice, admin, global domain` should be sufficient
>     2. Defining what permissions `admin` has in each domain is still required.
> 2. If `bob` is an `admin` in `subdomain1`, he should have `admin` access to `lowersubdomain1`
>     1. But no access to any other domain

**Your model:**

```ini
[request_definition]
r = sub, dom, obj, act

[policy_definition]
p = sub, dom, obj, act

[role_definition]
g = _, _, _
g2 = _, _

[policy_effect]
e = some(where (p.eft == allow))

[matchers]
m = g(r.sub, p.sub, r.dom) && g2(r.dom, p.dom) && regexMatch(r.obj, p.obj) && regexMatch(r.act, p.act)
```

**Your policy:**

```
p, admin, global_domain, data1, (read|write)
p, admin, sub_domain, data2, (read|write)

g, alice, admin, global_domain
g, bob, admin, sub_domain
g2, sub_domain, global_domain
```

**Your request(s):**

```
alice, global_domain, data1, read --> true, correct
alice, sub_domain, data2, read --> false, incorrect
bob, sub_domain, data2, read --> true, correct
```

tangyang9464:
@Sefriol 
I think your needs are very similar to this one #493 , isn't it ?

Sefriol:
Ah yes, this looks very similar indeed! Hopefully this gets implemented soon, since I think it's pretty typical RBAC usage situation!

Sefriol:
Refencing this comment as an update to this question: https://github.com/casbin/casbin/issues/493#issuecomment-907785066

[...]

Issue #718, opened 2021-02-26
What model would be efficient for hierarchial objects? [Question]


emgageadmin:
[<img alt="Issuehunt badges" src="https://img.shields.io/badge/IssueHunt-%2440%20Funded-%2300A156.svg" />](https://issuehunt.io/r/casbin/casbin/issues/718)

**Want to prioritize this issue? Try:**

[![issuehunt-to-marktext](https://github.com/BoostIO/issuehunt-materials/raw/master/v1/issuehunt-button-v1.svg)](https://issuehunt.io/r/casbin/casbin)

------

We have scenario similar to a file system, where the location in the hierarchy determines if a subject is authorized for particular object.  

This is for a multi-tenant environement 
There could be millions of objects and 96% of the objects will not have their own policy but will inherit a policy from their parent or an ancestor.
the hierarchy will be relativly static and small (maybe 12 deep with 20 to 300 nodes)
subjects will be individual users or roles
roles wil be hierarchical and it will be relativly static

# for example #

- h1  (p, alice, data1, read)
- - h2
- - - h3  (p, alice, data2, read)
- - - - h4 (p, aplice, data1, delete)
- - - - h5  (p, bob, data1, write)
- - - h6
- - - - h7 (p, admin_role, data1, write)
- - - h8  (p, alice, data1, write)

**This hierarchy should produce these authorization results:**
A) alice is granted read data1 at h1
B) alice is granted read data1 at h2 (because it deos not specify a policy for data1 and thus it inherits from h1)
C) alice is granted read for data2, data1 at h3 (there is data2 read policy at h3 and data1 read policy is inherited from h1)
D) alice is granted delete data1 and read data1 at h4 (there is data1 delete policy at h4 and data1 read policy inherited from h1) 
E) bob is granted write data1 at h5
F) bob is NOT granted any access at h1, h2, h3, h4, h6, h7, h8 
G) alice is granted read data2 at h6 and h7 
H) admin_role members are is granted  write data1 at h7
I) alice is granted  write data1 at h8

sorry if i belabered my question with too many example. just want to make sure there is no confusion. :)

# the question #

The question is, what would be the optimal model for this type of structures? 
We would like to model that would scale and be relativly performant.

any ideas would be greatly appreciate. Love casbin! and i have funded this question.
keep up the awesome work

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $40.00)

- $40.00 have been anonymously funded.

### Tips

[...]

PR #1040, opened 2022-06-25
feat: support domain hierarchical and GetNamedRoleManager

tangyang9464:

for: https://github.com/casbin/casbin/issues/1032#issuecomment-1159985149

### DIFF
--- a/examples/rbac_with_domains_hierarchical_policy.csv
+++ b/examples/rbac_with_domains_hierarchical_policy.csv
@@ -1,0 +1,11 @@
+p, admin, domain1, data1, (read|write)
+p, admin, domain2, data2, (read|write)
+p, admin, domain3, data3, (write)
+p, admin, sibling2, sdata2, (read|write)
+
+g, alice, admin, domain1
+g, bob, admin, domain2
+
+g2, domain2, domain1
+g2, sibling2, domain1
+g2, domain3, domain2
--- a/rbac/default-role-manager/role_manager.go
+++ b/rbac/default-role-manager/role_manager.go
@@ -437,12 +437,13 @@ func (rm *RoleManagerImpl) BuildRelationship(name1 string, name2 string, domain 
 }
 
 type DomainManager struct {
-	rmMap              *sync.Map
-	maxHierarchyLevel  int
-	matchingFunc       MatchingFunc
-	domainMatchingFunc MatchingFunc
-	logger             log.Logger
-	matchingFuncCache  *util.SyncLRUCache
+	rmMap                  *sync.Map
+	maxHierarchyLevel      int
+	matchingFunc           MatchingFunc
+	domainMatchingFunc     MatchingFunc
+	logger                 log.Logger
+	matchingFuncCache      *util.SyncLRUCache
+	domainHierarchyManager *RoleManager // Used to manage hierarchical relationships between domains
 }
 
 // NewDomainManager is the constructor for creating an instance of the
@@ -452,6 +453,11 @@ func NewDomainManager(maxHierarchyLevel int) *DomainManager {
 	_ = dm.Clear() // init rmMap and rmCache
 	dm.maxHierarchyLevel = maxHierarchyLevel
 	return dm
+}
+
+func (dm *DomainManager) SetDomainHierarchyManager(dhm *RoleManager) {
+	dm.domainHierarchyManager = dhm
+	dm.rebuild()
 }
 
 // SetLogger sets role manager's logger.
@@ -529,10 +535,18 @@ func (dm *DomainManager) Match(str string, pattern string) bool {
 }
 
 func (dm *DomainManager) rangeAffectedRoleManagers(domain string, fn func(rm *RoleManagerImpl)) {
-	if dm.domainMatchingFunc != nil {
+	if dm.domainMatchingFunc != nil || dm.domainHierarchyManager != nil {
 		dm.rmMap.Range(func(key, value interface{}) bool {
 			domain2 := key.(string)
-			if domain != domain2 && dm.Match(domain2, domain) {
+			matched := false
+			if dm.domainMatchingFunc != nil {
+				matched = dm.Match(domain2, domain)
+			}
+			hasLink := false
+			if dm.domainHierarchyManager != nil {
+				hasLink, _ = dm.domainHierarchyManager.HasLink(domain2, domain)
+			}
+			if domain != domain2 && (matched || hasLink) {
 				fn(value.(*RoleManagerImpl))
 			}
 			return true
@@ -557,11 +571,19 @@ func (dm *DomainManager) getRoleManager(domain string, store bool) *RoleManagerI
 		if store {
 			dm.rmMap.Store(domain, rm)
 		}
-		if dm.domainMatchingFunc != nil {
+		if dm.domainMatchingFunc != nil || dm.domainHierarchyManager != nil {
 			dm.rmMap.Range(func(key, value interface{}) bool {
 				domain2 := key.(string)
 				rm2 := value.(*RoleManagerImpl)
-				if domain != domain2 && dm.Match(domain, domain2) {
+				matched := false
+				if dm.domainMatchingFunc != nil {
+					matched = dm.Match(domain, domain2)
+				}
+				hasLink := false
+				if dm.domainHierarchyManager != nil {
+					hasLink, _ = dm.domainHierarchyManager.HasLink(domain, domain2)
+				}
+				if domain != domain2 && (matched || hasLink) {
 					rm.copyFrom(rm2)
 				}
 				return true
--- a/rbac_api_with_domains_test.go
+++ b/rbac_api_with_domains_test.go
@@ -18,6 +18,7 @@ import (
 	"sort"
 	"testing"
 
+	defaultrolemanager "github.com/casbin/casbin/v2/rbac/default-role-manager"
 	"github.com/casbin/casbin/v2/util"
 )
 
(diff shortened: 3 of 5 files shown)


tangyang9464:
@shipovalovyuriy  plz review. You can ues it like the example in the test code
```
func TestDomainHierarchical(t *testing.T) {
	e, _ := NewEnforcer("examples/rbac_with_domains_hierarchical_model.conf", "examples/rbac_with_domains_hierarchical_policy.csv")
	origin := e.GetNamedRoleManager("g").(*defaultrolemanager.RoleManager)
	dest := e.GetNamedRoleManager("g2").(*defaultrolemanager.RoleManager)
	origin.SetTenantManager(dest)

	testDomainEnforce(t, e, "alice", "dom2", "data", "read", true)
}
```

zsh1995:
Great job! Could you add `SetNamedRoleManager` method for semantic consistency?

tangyang9464:
> Great job! Could you add `SetNamedRoleManager` method for semantic consistency?

done

hsluoyz:
@zsh1995

rbac/default-role-manager/role_manager.go
```diff
@@ -557,6 +574,17 @@ func (dm *DomainManager) getRoleManager(domain string, store bool) *RoleManagerI
 				return true
 			})
 		}
+		if dm.tenantManager != nil {
```
Sefriol, line 577: Even though I do not know much about `golang`, this feels little repetitive, don't you think?

I think some comments would also be warranted. These specific conditions should be explained, otherwise somebody would just scratch their head in the future when they try to figure out why these are done as they are.

examples/rbac_with_domains_hierarchical_policy.csv
```diff
@@ -0,0 +1,6 @@
+p, admin, dom3, data, read
```
Sefriol, line 1: Also little more complexity to the domain is warranted. At least one extra user to test out that they do not have access to domain1 if they only have access to domain2

rbac_api_with_domains_test.go
```diff
@@ -267,3 +268,12 @@ func TestGetAllDomains(t *testing.T) {
 
 	testGetAllDomains(t, e, []string{"domain1", "domain2"})
 }
+
```
Sefriol, line 271: I think singular test case is not sufficient for such a complex feature. At least one which supports deep hierarchy should also be done. Like here:

https://github.com/casbin/casbin/issues/718#issuecomment-889785940

rbac_api_with_domains_test.go
```diff
@@ -267,3 +268,12 @@ func TestGetAllDomains(t *testing.T) {
...
+func TestDomainHierarchical(t *testing.T) {
+	e, _ := NewEnforcer("examples/rbac_with_domains_hierarchical_model.conf", "examples/rbac_with_domains_hierarchical_policy.csv")
+	origin := e.GetNamedRoleManager("g").(*defaultrolemanager.RoleManager)
+	dest := e.GetNamedRoleManager("g2").(*defaultrolemanager.RoleManager)
+	origin.SetTenantManager(dest)
+
+	testDomainEnforce(t, e, "alice", "dom2", "data", "read", true)
```
Sefriol, line 278: I think in terms of the nature of this test case, alice should be tested from domains 1, 2 and 3 since she has access to all of them.

Sefriol (changes requested):
Nice that people are finally taking this into consideration. I am not a `golang` expert, but I gave my two cents about this pull request.

Reference issues:
https://github.com/casbin/casbin/issues/718
https://github.com/casbin/casbin/issues/872
https://github.com/casbin/casbin/issues/1032

rbac_api_with_domains_test.go
Sefriol: More specifically: Multiple subjects, multiple objects, multiple domains

1. Test cases for accessing data in an upper domain from a lower domain (Failure)
2. Test cases for accessing data in a lower domain from a middle domain (Success)
3. Test cases for accessing data in a sibling domain from another sibling (Failure)
4. Test cases for accessing data in a sibling domain from parent (success)

rbac/default-role-manager/role_manager.go
```diff
@@ -449,6 +451,11 @@ func NewDomainManager(maxHierarchyLevel int) *DomainManager {
 	return dm
 }
 
+func (dm *DomainManager) SetTenantManager(tm rbac.RoleManager) {
```
Sefriol, line 454: Needs documentation on how to use this. Though at the same time, shouldn't the hierarchy be enabled by default?

What is base definition for Named Grouping Policies? If you set up a grouping policy for domains shouldn't they be hierarchical by default? @hsluoyz

rbac/default-role-manager/role_manager.go
tangyang9464: I don't quite understand what repetition means, what is your suggestion

tangyang9464: did some refactoring

examples/rbac_with_domains_hierarchical_policy.csv
tangyang9464: Added more tests

rbac/default-role-manager/role_manager.go
tangyang9464: I don't quite understand what you are saying. This PR is to use the RoleManager to manage the hierarchy of the domain

rbac/default-role-manager/role_manager.go
```diff
@@ -438,6 +439,7 @@ type DomainManager struct {
 	domainMatchingFunc MatchingFunc
 	logger             log.Logger
 	matchingFuncCache  *util.SyncLRUCache
+	tenantManager      rbac.RoleManager // Used to manage hierarchical relationships between domains, see
```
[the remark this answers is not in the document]
tangyang9464: rename

rbac_api_with_domains_test.go
tangyang9464: Added more tests

rbac_api_with_domains_test.go
tangyang9464: Added more tests

tangyang9464:
@Sefriol Some changes have been made in response to your suggestion, please review again.

rbac_api_with_domains_test.go
Sefriol: I tried to implement this in https://github.com/casbin/node-casbin/pull/380

I think it's not possible to define `g2` in the format of `child, parent`. At least in nodeJS, it provided incorrect results, but maybe `golang` has a different RoleManager definition.

You can try to replicate the test case I did for nodeJS. I think it provides quite good coverage on all the possible use cases. Especially the 4 domain structure is essential.

I have tried to comment on why certain decisions were made in the pull request and a little bit about the thought behind the hierarchy model.

rbac/default-role-manager/role_manager.go
Sefriol: I was thinking out loud and more broadly how hierarchy should be enabled for the DomainManager. Not something that is necessary to solve in this pull request.

rbac/default-role-manager/role_manager.go
```diff
@@ -529,10 +535,18 @@ func (dm *DomainManager) Match(str string, pattern string) bool {
 }
 
 func (dm *DomainManager) rangeAffectedRoleManagers(domain string, fn func(rm *RoleManagerImpl)) {
-	if dm.domainMatchingFunc != nil {
+	if dm.domainMatchingFunc != nil || dm.domainHierarchyManager != nil {
 		dm.rmMap.Range(func(key, value interface{}) bool {
 			domain2 := key.(string)
-			if domain != domain2 && dm.Match(domain2, domain) {
+			matched := false
```
Sefriol, line 541: This can probably be packed similarly to [this](https://github.com/casbin/node-casbin/pull/380/files#diff-d6c8111bf43f43510834bf57fa2402634aeec52b3988dd17230dfe3cdb7baba5R202).

Otherwise lgtm.

rbac/default-role-manager/role_manager.go
```diff
@@ -454,6 +455,11 @@ func NewDomainManager(maxHierarchyLevel int) *DomainManager {
 	return dm
 }
 
+func (dm *DomainManager) SetDomainHierarchyManager(dhm *RoleManager) {
```
Sefriol, line 458: I am not sure if this is possible through the code, but I think this could have a default value like `e.rmMap['g2']`, so that users can define it, but do not necessarily need to.

rbac_api_with_domains_test.go
```diff
@@ -267,3 +268,22 @@ func TestGetAllDomains(t *testing.T) {
 
 	testGetAllDomains(t, e, []string{"domain1", "domain2"})
 }
+
+func TestDomainHierarchical(t *testing.T) {
+	e, _ := NewEnforcer("examples/rbac_with_domains_hierarchical_model.conf", "examples/rbac_with_domains_hierarchical_policy.csv")
+	origin := e.GetNamedRoleManager("g").(*defaultrolemanager.RoleManager)
+	dest := e.GetNamedRoleManager("g2").(*defaultrolemanager.RoleManager)
+	origin.SetDomainHierarchyManager(dest)
```
Sefriol, line 276: Should we add a helper function, so that you can enable the hierarchy without digging up the RoleManager? Like `EnableDomainHierarchy` which would set DomainHierachy to `g` with `g2` by default.

Sefriol (changes requested):
I think we need somebody else to review this than just me 😀.

rbac/default-role-manager/role_manager.go
tangyang9464: Unless a ``[domain_definition]`` is defined in the model like ``[role_definition]``, but redundant coding will greatly increase

rbac_api_with_domains_test.go
tangyang9464: @Sefriol 
> I think it's not possible to define g2 in the format of child, parent. At least in nodeJS, it provided incorrect results, but maybe golang has a different RoleManager definition.
> You can try to replicate the test case I did for nodeJS. I think it provides quite good coverage on all the possible use cases. Especially the 4 domain structure is essential.
> I have tried to comment on why certain decisions were made in the pull request and a little bit about the thought behind the hierarchy model.

Maybe you should use ``conf`` and ``csv`` files so that we can replicate each other's use cases

tangyang9464:
@Sefriol 
> I think it's not possible to define g2 in the format of child, parent. At least in nodeJS, it provided incorrect results, but maybe golang has a different RoleManager definition.
> You can try to replicate the test case I did for nodeJS. I think it provides quite good coverage on all the possible use cases. Especially the 4 domain structure is essential.
> I have tried to comment on why certain decisions were made in the pull request and a little bit about the thought behind the hierarchy model.

Maybe you should use ``conf`` and ``csv`` files so that we can replicate each other's use cases

rbac_api_with_domains_test.go
tangyang9464: It might be good, but I don't have time to do it right now

Sefriol:
@tangyang9464 Sure. I changed the implementation to use csv and conf files.

https://github.com/casbin/node-casbin/pull/380/files

tangyang9464:
> > Maybe you should use `conf` and `csv` files so that we can replicate each other's use cases
> 
> @tangyang9464 Sure. I changed the implementation to use `csv` and `conf` files.
> 
> https://github.com/casbin/node-casbin/pull/380/files

@zsh1995 I tested all your use cases and all passed. Maybe we don't need to change g2 to ``g2(parent, child)``.

tangyang9464:
@zsh1995 At the same time, I want to discuss the difference between ``m = g(r.sub, p.sub, r.dom) && r.dom==p.dom && r.obj == p.obj && regexMatch(r.act, p.act)`` and ``m = g(r.sub, p.sub, r.dom) && g2(p.dom, r.dom) && r.obj == p.obj && regexMatch(r.act, p.act)``. The former means that users can be automatically assigned roles in multiple domains according to the domain hierarchy, but defining what permissions role has in each domain is still required. The latter is a role that automatically assigns multiple domains to the user according to the domain hierarchy, and the permissions of the role will also be automatically assigned according to the domain hierarchy. Please pay attention to the similarities and differences of the results of your test cases ``expect(e.enforceSync('alice', 'domain1', 'data2', 'read')).toBe(true)`` under the two models

Sefriol:
Yeah I am not sure what kind of exact solution we should be looking for. My intent with following policy lines:

```csv
p, admin, domain1, data1, (read|write)
p, admin, domain2, data2, (read|write)
p, admin, domain3, data3, (write)
p, admin, sibling2, sdata2, (read|write)

g, alice, admin, domain1
g, bob, admin, domain2
g2, domain1, domain2
g2, domain1, sibling2
g2, domain2, domain3
```

is to create domain structure like 

```bash
domain1
├── domain2
│   └── domain3
└── sibling2
```

Which would result that Alice is an admin in all domains and Bob only in domains 2 and 3 with as few `g` definitions as possible. 

Key things I am trying to achieve here:
1. Each domain can have their custom permission table and can define their roles as they want to.
2. Easy role addition / removal from an user

Policy testing should include that:
1. Bob has no rights to `domain1`, `sibling2´, but does for `domain2` and `domain3`
2. Alice has access to all domains

Checks like `expect(e.enforceSync('alice', 'domain1', 'data2', 'read')).toBe(true)` do not matter too much. Personally just ` expect(e.enforceSync('alice', 'domain2, 'data2', 'read')).toBe(true)` is good enough.

I guess more important factor in this case is performance. In a case of not knowing which domain object `dataX` belongs to, is it faster to check all domains or just use reversed `g2` model instead? Sometimes we are interested whether subject has access to an object and not if subject has access to an object in specific domain. 

I derived my implementation from a model used in our production, so using hierarchy would not require us to do any conversation work.

I'll add your model too to `node-casbin` test cases to see that it works as well. I might need to add some performance testing as well to see how much it takes time in each model.

Sefriol:
For some reason these tests in `node-casbin` only work when `HasLink` inputs are implemented in reverse order compared to this (`this.domainHierarchyManager.syncedHasLink(domain, domain2)` vs `dm.domainHierarchyManager.HasLink(domain2, domain)`). But maybe I just cannot read the `golang` code correctly.

@tangyang9464 Can you send me the results of `golang` `HasLink` function per domain from your example? Like `domain, domain2, hasLinkResult`. I need to investigate whether the function works correctly.

Sefriol:
@nodece Commented on this in the `node-casbin` pull request. https://github.com/casbin/node-casbin/pull/380#issuecomment-1228714762

In general I am in favor of implementing this directly to the model definition.

tangyang9464:
> @nodece Commented on this in the `node-casbin` pull request. https://github.com/casbin/node-casbin/pull/380#issuecomment-1228714762
> 
> In general I am in favor of implementing this directly to the model definition.

@Sefriol Agree, modify later

Sefriol:
@tangyang9464 Have you heard any updates on the issue? Is there currently a plan on how this could be implemented?

@hsluoyz Pinging you as well if there is anything to update :)

hsluoyz:
@tangyang9464 @imp2002

sujit-baniya:
@tangyang9464 on latest casbin code, following test cases are not working.

```go
	testDomainEnforce(t, e, "alice", "domain1", "data2", "read", true)  // Failed
	testDomainEnforce(t, e, "alice", "domain1", "sdata2", "read", true) // Failed
	testDomainEnforce(t, e, "alice", "domain1", "data3", "write", true) // Failed
	testDomainEnforce(t, e, "alice", "domain2", "data3", "write", true) // Failed
	testDomainEnforce(t, e, "bob", "domain2", "data3", "write", true)   // Failed
```

Sefriol:
@sujit-baniya Current code is unlikely to be accepted anyway since there was more preferred idea to support this directly in the model definition.

sujit-baniya:
@Sefriol Can you please suggest preferred ways to implement via model definition?

nodece:
> @Sefriol Can you please suggest preferred ways to implement via model definition?

I suggest we should do so to avoid diversification.

Sefriol:
> @Sefriol Can you please suggest preferred ways to implement via model definition?

I think Casbin Team should decide what kind of approach we should be going into. They have a better idea than I do on what needs to be taken into account. I can be part of the implementation on the NodeJS side, but

Sefriol:
@hsluoyz @nodece Any progress on the design front? This seems to be quite a popular feature request among users.

Sefriol:
@nodece @tangyang9464 @hsluoyz Have you had a discussion about this? NodeJS version of this was kinda just merged with unsolved questions.

Issue #1015, opened 2022-05-20
[Bug]DeleteRole do not totally work in priority model


phoenixsqf:
**Want to prioritize this issue? Try:**

[![issuehunt-to-marktext](https://github.com/BoostIO/issuehunt-materials/raw/master/v1/issuehunt-button-v1.svg)](https://issuehunt.io/r/casbin/casbin)

------

**Describe the bug**
Enforcer.DeleteRole do not totally work in priority model

**Model used**

```
[request_definition]
r = sub, obj, act

[policy_definition]
p = priority, sub, obj, act, eft

[role_definition]
g = _, _

[policy_effect]
e = priority(p.eft) || deny

[matchers]
m = g(r.sub, p.sub) && r.obj == p.obj && r.act == p.act
```

Function DeleteRole
```
// DeleteRole deletes a role.
// Returns false if the role does not exist (aka not affected).
func (e *Enforcer) DeleteRole(role string) (bool, error) {
	var err error
	res1, err := e.RemoveFilteredGroupingPolicy(1, role)
	if err != nil {
		return res1, err
	}

	res2, err := e.RemoveFilteredPolicy(0, role)
	return res1 || res2, err
}
```

In this model, priority is the first index of policy definition，but the function try to delete policy by the first index.
So the permission will not be removed. It might caused some problems.

**Expected behavior**
Maybe try to call e.RemoveFilteredPolicy(1, role) when the value of priorityIndex is 0？

hsluoyz:
@tangyang9464

hsluoyz:
@JalinWang @tangyang9464

tangyang9464:
@hsluoyz Both ``DeleteRole`` and ``DeleteUser`` default ``sub`` principal in the first column of the ``policy_definition``, which obviously does not apply in all cases. Should we provide parameters to let our user pass in their defined ``sub index`` or just restrict our user to have to define the ``sub`` as the first column?

hsluoyz:
@tangyang9464 here is a simpler method. We can detect what index the `sub` is in within the model. e.g., we can read `p = priority, sub, obj, act, eft` and know `sub` is the 2nd element. Better we can store this information somewhere in the enforcer, so we don't need to retrieve it for every call.

tangyang9464:
[...]

PR #1041, opened 2022-06-26
feat: add FieldIndexMap and constant

tangyang9464:
# ChangLog
Some APIs (mostly RBAC APIs) need to obtain the index of variables such as sub, obj, dom, priority, etc. in policy_define. So this PR did the following work
1. add ``FieldIndexMap`` to store index.
2. add ``GetFieldIndex`` and ``SetFieldIndex``, try to get the index with ``sub, obj, dom, priority`` as the default definition named values. If the user does not use the default value, then ``SetFieldIndex`` should be used to set the index

Some other work is adding constants, putting some constants together instead of using strings directly in the code.

related: https://github.com/casbin/casbin/issues/1015#issuecomment-1166218883 #1015 

### DIFF
--- a/constant/constants.go
+++ b/constant/constants.go
@@ -1,0 +1,30 @@
+// Copyright 2022 The casbin Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package constant
+
+const (
+	DomainIndex   = "dom"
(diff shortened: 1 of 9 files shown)


constant/constants.go
```diff
@@ -0,0 +1,16 @@
+package constant
```
hsluoyz, line 1: Apache header

model/model.go
```diff
@@ -308,22 +309,28 @@ func getNameWithDomain(domain string, name string) string {
 
 func (model Model) SortPoliciesByPriority() error {
 	for ptype, assertion := range model["p"] {
-		for index, token := range assertion.Tokens {
-			if token == fmt.Sprintf("%s_priority", ptype) {
-				assertion.priorityIndex = index
-				break
+		hasPriority := false
+		if _, ok := assertion.FieldIndexMap[constant.PriorityIndex]; !ok {
```
hsluoyz, line 313: Why don't use the newly defined `GetFieldIndex` for all array element fetching?

hsluoyz (changes requested):

constant/constants.go
tangyang9464: done

model/model.go
tangyang9464: done

hsluoyz (approved):

PR #1061, opened 2022-07-26
fix: wrong type of locking

JalinWang:
`SavePolicy()` should require the lock for RW.

### DIFF
--- a/enforcer_synced.go
+++ b/enforcer_synced.go
@@ -131,8 +131,8 @@ func (e *SyncedEnforcer) LoadIncrementalFilteredPolicy(filter interface{}) error
 
 // SavePolicy saves the current policy (usually after changed with Casbin API) back to file/database.
 func (e *SyncedEnforcer) SavePolicy() error {
-	e.m.RLock()
-	defer e.m.RUnlock()
+	e.m.Lock()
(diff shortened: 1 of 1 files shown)


hsluoyz:
@tangyang9464

enforcer_synced.go
```diff
@@ -131,8 +131,8 @@ func (e *SyncedEnforcer) LoadIncrementalFilteredPolicy(filter interface{}) error
 
 // SavePolicy saves the current policy (usually after changed with Casbin API) back to file/database.
 func (e *SyncedEnforcer) SavePolicy() error {
-	e.m.RLock()
-	defer e.m.RUnlock()
+	e.m.Lock()
```
tangyang9464, line 134: Will there be a race in SavePolicy, I think it just reads the resource

JalinWang: Saving policies cause the adapter to write to, i.e., a database.
A `RemovePolicy()` and the `SavePolicy` will emit concurrent database writes.

tangyang9464: The adapter itself should ensure transaction security, but it seems that it does not now, it can only be locked here

tangyang9464 (approved):
lgtm

PR #1068, opened 2022-08-08
fix: correct the pattern name in builtin_operators_test

cs1137195420:
Correct the pattern name in builtin_operators_test.go


cs1137195420:
When I tested the test case in casbin-cpp, I found these patterns that may be wrong. I'm not sure if any of these needs to be changed, is there anyone who can check?

hsluoyz:
@cs1137195420 which patterns are wrong? Did you double-check it in other libs like Go, Java, etc?

hsluoyz:
@cs1137195420 I think these cases are expected to have open bracket to test the corner testes, so nothing wrong here

PR #1073, opened 2022-08-10
fix WatcherEx comment

GerryLon:
fix WatcherEx comment


JalinWang:
I think it's okay. Should we reopen it? @hsluoyz

Issue #1092, opened 2022-09-01
WatcherEx Callback Loop


Shivansh-yadav13:
Im working on watcherEx for node-casbin, and for testing we created a callback function for updating other enforcer instances for 'addPolicy'.
In the callback function we used 'addPolicyInternal'.
The issue is addPolicyInternal calls the watcherEx callback function and so it will become a loop

proposed solution-
we can maybe call watcherEx callback function in 'addPolicy' and not in 'addPolicyInternal'

same for all other methods

JalinWang:
You are right. Sorry for stalling on https://github.com/casbin/casbin/issues/999 for a while, I'm gonna resume.

Shivansh-yadav13:
@nodece should we make these changes in node-casbin & casbin-core?

nodece:
`addPolicyInternal` means an internal method, do we have a better name?

nodece:
> @nodece should we make these changes in node-casbin & casbin-core?

Yes, you can make a PR to node-casbin.

Issue #999, opened 2022-04-27
[Proposal] Working on WatcherEx implementation


JalinWang:
I'm currently working on implementing the WatcherEx interface for #943 #960 #949 .

This issue is open for tracking.

JalinWang:
Completed.

PR #1093, opened 2022-09-01
feat: add management api for WatcherEx

JalinWang:
fix https://github.com/casbin/casbin/issues/1092
part of: https://github.com/casbin/casbin/issues/999

### DIFF
--- a/internal_api.go
+++ b/internal_api.go
@@ -30,8 +30,12 @@ func (e *Enforcer) shouldPersist() bool {
 	return e.adapter != nil && e.autoSave
 }
 
+func (e *Enforcer) shouldNotify() bool {
+	return e.watcher != nil && e.autoNotifyWatcher
+}
+
 // addPolicy adds a rule to the current policy.
-func (e *Enforcer) addPolicy(sec string, ptype string, rule []string) (bool, error) {
+func (e *Enforcer) addPolicyWithoutNotify(sec string, ptype string, rule []string) (bool, error) {
 	if e.dispatcher != nil && e.autoNotifyDispatcher {
 		return true, e.dispatcher.AddPolicies(sec, ptype, [][]string{rule})
 	}
@@ -57,21 +61,11 @@ func (e *Enforcer) addPolicy(sec string, ptype string, rule []string) (bool, err
 		}
 	}
 
-	if e.watcher != nil && e.autoNotifyWatcher {
-		var err error
-		if watcher, ok := e.watcher.(persist.WatcherEx); ok {
-			err = watcher.UpdateForAddPolicy(sec, ptype, rule...)
-		} else {
-			err = e.watcher.Update()
-		}
-		return true, err
-	}
-
 	return true, nil
 }
 
 // addPolicies adds rules to the current policy.
-func (e *Enforcer) addPolicies(sec string, ptype string, rules [][]string) (bool, error) {
+func (e *Enforcer) addPoliciesWithoutNotify(sec string, ptype string, rules [][]string) (bool, error) {
 	if e.dispatcher != nil && e.autoNotifyDispatcher {
 		return true, e.dispatcher.AddPolicies(sec, ptype, rules)
 	}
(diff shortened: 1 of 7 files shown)


management_api.go
```diff
@@ -411,10 +411,38 @@ func (e *Enforcer) UpdateNamedGroupingPolicies(ptype string, oldRules [][]string
...
-	return e.removeFilteredPolicy("g", ptype, fieldIndex, fieldValues...)
+	return e.removeFilteredPolicy("g", ptype, fieldIndex, fieldValues)
 }
 
 // AddFunction adds a customized function.
 func (e *Enforcer) AddFunction(name string, function govaluate.ExpressionFunction) {
 	e.fm.AddFunction(name, function)
 }
+
+func (e *Enforcer) SelfAddPolicy(sec string, ptype string, rule []string) (bool, error) {
```
tangyang9464, line 422: My opinion is to extract the code in ``addPolicy`` and put it into ``SelfAddPolicy``, and then call ``SelfAddPolicy`` in ``addPolicy`` instead of modifying the original parameters of ``addPolicy``.
like this
```
func (e *Enforcer) addPolicy(sec string, ptype string, rule []string) (bool, error) {
	e.SelfAddPolicy(sec,ptype,rule)
	if e.watcher != nil && e.autoNotifyWatcher {

        }
```

tangyang9464 (changes requested):

internal_api.go
```diff
@@ -343,38 +276,187 @@ func (e *Enforcer) updateFilteredPolicies(sec string, ptype string, newRules [][
...
-		return true, e.dispatcher.UpdateFilteredPolicies(sec, ptype, oldRules, newRules)
+		return oldRules, e.dispatcher.UpdateFilteredPolicies(sec, ptype, oldRules, newRules)
 	}
 
 	ruleChanged := e.model.RemovePolicies(sec, ptype, oldRules)
 	e.model.AddPolicies(sec, ptype, newRules)
 	ruleChanged = ruleChanged && len(newRules) != 0
 	if !ruleChanged {
-		return ruleChanged, nil
+		return nil, nil
```
tangyang9464, line 286: I think it is better to return empty array（oldRules） instead of nil

internal_api.go
```diff
@@ -343,38 +276,187 @@ func (e *Enforcer) updateFilteredPolicies(sec string, ptype string, newRules [][
...
+
+	if e.shouldNotify() {
 		var err error
 		if watcher, ok := e.watcher.(persist.WatcherUpdatable); ok {
 			err = watcher.UpdateForUpdatePolicies(oldRules, newRules)
 		} else {
 			err = e.watcher.Update()
 		}
-		return ruleChanged, err
+		return true, err
```
tangyang9464, line 456: can be removed

management_api.go
```diff
@@ -258,7 +258,7 @@ func (e *Enforcer) UpdateFilteredPolicies(newPolicies [][]string, fieldIndex int
 }
 
 func (e *Enforcer) UpdateFilteredNamedPolicies(ptype string, newPolicies [][]string, fieldIndex int, fieldValues ...string) (bool, error) {
-	return e.updateFilteredPolicies("p", ptype, newPolicies, fieldIndex, fieldValues...)
+	return e.updateFilteredPolicies("p", ptype, newPolicies, fieldIndex, fieldValues)
```
tangyang9464, line 261: Why modify here

management_api.go
```diff
@@ -411,10 +411,38 @@ func (e *Enforcer) UpdateNamedGroupingPolicies(ptype string, oldRules [][]string
...
+	return e.removeFilteredPolicy("g", ptype, fieldIndex, fieldValues)
 }
 
 // AddFunction adds a customized function.
 func (e *Enforcer) AddFunction(name string, function govaluate.ExpressionFunction) {
 	e.fm.AddFunction(name, function)
 }
+
+func (e *Enforcer) SelfAddPolicy(sec string, ptype string, rule []string) (bool, error) {
+	return e.addPolicy(sec, ptype, rule)
```
tangyang9464, line 423: This is what should be called ``addPolicyWithoutNotify``

tangyang9464 (changes requested):

management_api.go
JalinWang: I forget to reset this

management_api.go
JalinWang: Uhh? My test fails to check it……

management_api.go
JalinWang: Fixed as well as the test

internal_api.go
JalinWang: It is used to return the error (the scope of `err` is limited in this `if` statement)

management_api.go
JalinWang: Fixed, as well as the test case.

internal_api.go
tangyang9464: ok

tangyang9464 (approved):
lgtm

hsluoyz:
@JalinWang plz also add these new APIs to docs v2

PR #1109, opened 2022-09-18
fix: refactor the LoadPolicy in SyncedEnforcer, reduce the lock range

### LINKED ISSUE #1108: [Feature] Cut down the lock range in SyncedEnforcer
In SyncedEnforcer, LoadPolicy is a heavy time cost function if there are a lots of policies.

https://github.com/casbin/casbin/blob/f58b2c99df610c757ddbda947f985d7f566709e1/enforcer_synced.go#L117-L121

It can use the lock only when access the critical section.

deagon:
fix #1108

### DIFF
--- a/enforcer_synced.go
+++ b/enforcer_synced.go
@@ -22,6 +22,8 @@ import (
 	"github.com/Knetic/govaluate"
 
 	"github.com/casbin/casbin/v2/persist"
+	"github.com/casbin/casbin/v2/rbac"
+	defaultrolemanager "github.com/casbin/casbin/v2/rbac/default-role-manager"
 )
 
 // SyncedEnforcer wraps Enforcer and provides synchronized access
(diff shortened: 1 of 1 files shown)


hsluoyz:
@deagon what does this PR do?

deagon:
> @deagon what does this PR do?

@hsluoyz fix https://github.com/casbin/casbin/issues/1108

hsluoyz:
@JalinWang plz review

/cc @tangyang9464

JalinWang:
The pr is technically ok. However, when `LoadPolicy` is called, it indicates changes in policies. If the function starts without locking, there might be other enforcement requests using out-of-date policies.

deagon:
> The pr is technically ok. However, when `LoadPolicy` is called, it indicates changes in policies. If the function starts without locking, there might be other enforcement requests using out-of-date policies.

That's right. However, `LoadPolicy` costs almost 3 seconds in my production environment. The permission read requests would be blocked 3 seconds If it locks, which is unacceptable for us.
In this case, I think out-of-date policies might be ok.

enforcer_synced.go
```diff
@@ -115,9 +117,42 @@ func (e *SyncedEnforcer) ClearPolicy() {
 
 // LoadPolicy reloads the policy from file/database.
 func (e *SyncedEnforcer) LoadPolicy() error {
```
hsluoyz, line 119: Define a new function named `LoadPolicyFast()` to hold your code

hsluoyz (changes requested):

deagon:
@hsluoyz  Please take a review, thanks.

hsluoyz (approved):

hsluoyz:
@deagon hi, this is impressive PR! Plz join the QQ group: 540163681

pinkyo:
thanks a lot, it help us solved a global performance issue. we got 400k policy data in db, and a reload operation cost more than 10s.

PR #1117, opened 2022-10-11
feat: fix sub hierarchy invalid in `loadFilteredPolicy`

imp2002:
`SortPoliciesBySubjectHierarchy()` should be executed at `loadFilteredPolicy()` just like `LoadPolicy()`.

### DIFF
--- a/enforcer.go
+++ b/enforcer.go
@@ -353,6 +353,10 @@ func (e *Enforcer) loadFilteredPolicy(filter interface{}) error {
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


imp2002:
The CI error seem be due to a temporary network issue.
image

hsluoyz:
@imp2002 which issue does this PR fix?

imp2002:
There is no issue about it, but maybe it should work in `loadFilteredPolicy` too.
[load-policy-with-priority-based-on-role-and-user-hierarchy](https://casbin.io/docs/priority-model#load-policy-with-priority-based-on-role-and-user-hierarchy)

Issue #1176, opened 2023-01-19
Does keyMatch5 does not support '*'?


MkDierz:
**Want to prioritize this issue? Try:**

[![issuehunt-to-marktext](https://github.com/BoostIO/issuehunt-materials/raw/master/v1/issuehunt-button-v1.svg)](https://issuehunt.io/r/casbin/casbin)

------

**What's your scenario? What do you want to achieve?**
so i'm using RESTful model but when im adding '*' to policy file the request got rejected 

**model:**

```ini
  [request_definition]
  r = sub, obj, act

  [policy_definition]
  p = sub, obj, act, eft

  [role_definition]
  g = _, _

  [policy_effect]
  e = some(where (p.eft == allow))

  [matchers]
  m = g(r.sub, p.sub) && keyMatch5(r.obj, p.obj) && (r.act == p.act || p.act == "*")
```

**policy:**

```
  p, 1, /api/v1/sdm, GET
  p, 1, /api/v1/sdm/*, GET

```

**Your request(s):**

```
1,  /api/v1/sdm , GET ---> true (expected: true)
1,  /api/v1/sdm/20 , GET ---> false (expected: true)
```

MkDierz:
Hey im learning to contribute, this is my first.
may i take on this issue?
@hsluoyz

hsluoyz:
@MkDierz OK, also add test to: https://github.com/casbin/casbin/blob/857bd7f49362691392804117ab04eee4d65e8262/util/builtin_operators_test.go#L485

MkDierz:
Thanks, will do.

MkDierz:
um hey @hsluoyz i can't push my branch into master to do a pull request, am i doing it the wrong way?

MkDierz:
sorry my bad i figured it out

MkDierz:
Hey @hsluoyz I've created pull request please review

PR #1177, opened 2023-01-19
fix: add support to use '*' in keymatch5 utils

MkDierz:
Fix: https://github.com/casbin/casbin/issues/1176

### DIFF
--- a/enforcer.go
+++ b/enforcer.go
@@ -66,13 +66,12 @@ type EnforceContext struct {
 //
 // File:
 //
-// 	e := casbin.NewEnforcer("path/to/basic_model.conf", "path/to/basic_policy.csv")
+//	e := casbin.NewEnforcer("path/to/basic_model.conf", "path/to/basic_policy.csv")
 //
 // MySQL DB:
 //
-// 	a := mysqladapter.NewDBAdapter("mysql", "mysql_username:mysql_password@tcp(127.0.0.1:3306)/")
-// 	e := casbin.NewEnforcer("path/to/basic_model.conf", a)
-//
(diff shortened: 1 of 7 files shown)


hsluoyz:
@PokIsemaine plz review

PokIsemaine:
@MkDierz Thanks for your contribution, the basic implementation and testing are fine, it would be even better if you can help improve the comments of KeyMatch5.
``` golang
// KeyMatch determines whether key1 matches the pattern of key2 and ignores the parameters in key2.
// For example, "/foo/bar?status=1&type=2" matches "/foo/bar"
func KeyMatch5(key1 string, key2 string) bool {
}
```
For example：
* fix `KeyMatch` -> `KeyMatch5`
* add more examples
  * You can choose some examples from your test
  * Add an example of matching failure

You can refer to the comments of KeyMatch4
```golang
// KeyMatch4 determines whether key1 matches the pattern of key2 (similar to RESTful path), key2 can contain a *.
// Besides what KeyMatch3 does, KeyMatch4 can also match repeated patterns:
// "/parent/123/child/123" matches "/parent/{id}/child/{id}"
// "/parent/123/child/456" does not match "/parent/{id}/child/{id}"
// But KeyMatch3 will match both.
```

MkDierz:
Hi thanks for the review, ill do it ASAP.

PokIsemaine:
@MkDierz You can try to use `push -f` to remove commits that do not meet the requirements
image

MkDierz:
> @MkDierz You can try to use `push -f` to remove commits that do not meet the requirements image

i'm sorry for the trouble
i'll do it next time thanks

PR #1184, opened 2023-01-31
fix : rebuild() function infinite loop by  role_manager.go

anujcontractor:


hsluoyz:
This PR doesn't make sense and is low quality

Issue #1181, opened 2023-01-27
[Bug] `Enforce()`  does not reflect latest policy after calling `LoadPolicy()`


t3-aoki:
**Describe the bug**
Since v2.60.0, [Enforcer.Enforce()](https://pkg.go.dev/github.com/casbin/casbin/v2#Enforcer.Enforce) results don't reflect the latest policy after calling [LoadPolicy()](https://pkg.go.dev/github.com/casbin/casbin/v2#Enforcer.LoadPolicy) method.

**To Reproduce**

```go
package main_test

import (
	"log"
	"os"
	"path/filepath"
	"testing"

	"github.com/casbin/casbin/v2"
	"github.com/casbin/casbin/v2/model"
	fileadapter "github.com/casbin/casbin/v2/persist/file-adapter"
)

func TestLoadPolicy(t *testing.T) {
	conf := `[request_definition]
r = sub, obj, act

[policy_definition]
p = sub, obj, act

[role_definition]
g = _ , _

[policy_effect]
e = some(where (p.eft == allow))

[matchers]
m = g(r.sub, p.sub)  && r.obj == p.obj && r.act == p.act
`
	m, err := model.NewModelFromString(conf)
	if err != nil {
		log.Fatal(err)
	}

	policyFile := filepath.Join(t.TempDir(), "policy.csv")

	f, err := os.Create(policyFile)
	if err != nil {
		log.Fatal(err)
	}
	defer f.Close()

	policy := `p, admin, data1, read
g, alice, admin
`
	if _, err := f.WriteString(policy); err != nil {
		log.Fatal(err)
	}

	a := fileadapter.NewAdapter(policyFile)

	e, err := casbin.NewEnforcer(m, a)
	if err != nil {
		log.Fatal(err)
	}

	t.Run("first", func(tt *testing.T) {
		ok, err := e.Enforce("bob", "data1", "read")
		if err != nil {
			log.Fatal(err)
		}

		if ok {
			tt.Errorf("%v, expected %v", ok, false)
		}
	})

	newPolicy := "g, bob, admin\n"
	if _, err := f.WriteString(newPolicy); err != nil {
		log.Fatal(err)
	}

	if err := e.LoadPolicy(); err != nil {
		log.Fatal(err)
	}

	t.Run("second", func(tt *testing.T) {
		ok, err := e.Enforce("bob", "data1", "read")
		if err != nil {
			log.Fatal(err)
		}

		if !ok {
[...]

PR #1188, opened 2023-02-03
fix: LoadPolicy() does not reflect the latest results

PokIsemaine:
Fix: https://github.com/casbin/casbin/issues/1181
## Description:
https://github.com/casbin/casbin/issues/1181#issuecomment-1415122920

### DIFF
--- a/enforcer.go
+++ b/enforcer.go
@@ -66,13 +66,12 @@ type EnforceContext struct {
 //
 // File:
 //
-// 	e := casbin.NewEnforcer("path/to/basic_model.conf", "path/to/basic_policy.csv")
+//	e := casbin.NewEnforcer("path/to/basic_model.conf", "path/to/basic_policy.csv")
 //
 // MySQL DB:
 //
-// 	a := mysqladapter.NewDBAdapter("mysql", "mysql_username:mysql_password@tcp(127.0.0.1:3306)/")
-// 	e := casbin.NewEnforcer("path/to/basic_model.conf", a)
-//
+//	a := mysqladapter.NewDBAdapter("mysql", "mysql_username:mysql_password@tcp(127.0.0.1:3306)/")
+//	e := casbin.NewEnforcer("path/to/basic_model.conf", a)
 func NewEnforcer(params ...interface{}) (*Enforcer, error) {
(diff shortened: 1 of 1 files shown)


xingzem:
Hi, @PokIsemaine, @hsluoyz,   one issue I met is the enforcer.getFilteredNamedGroupingPolicy() does not give me the latest results in DB in my first call to this function. And the second call would reflect the correct result. Thinking it might be a cache mismatch with storage. I tried to add enforcer.loadPolicy() before the above function. The same issue still persists. I have tried the latest version here: https://central.sonatype.com/artifact/org.casbin/jcasbin/1.32.2/versions. Could you help with some comments on this? Thanks.

PokIsemaine:
@xingzem Thanks for your feedback! Could you provide a code demo to reproduce this issue? This helps us confirm if they are the same issue and fix the problem. If convenient, you can create a new issue and describe it in detail in the issue.

hsluoyz:
@xingzem which Casbin are you using? Go or Java?

xingzem:
Hi @hsluoyz , we are using java. @PokIsemaine I am not sure whether I can get a demo out as I am not familiar how to set up the environment to run casbin separately. Let me check that though. 
The code we have right now is as belows: 
```
  @Override
  @Synchronized
  public List<User> getUsers(@Nonnull final String role, @Nonnull final String domain) {
    final List<User> users = new ArrayList<>();
    enforcer.loadPolicy();
    final List<List<String>> policies =
        enforcer.getFilteredNamedGroupingPolicy(GROUPING.toString(), 0, "", role, domain);

    for (final List<String> policy : policies) {
      users.add(toUser(policy));
    }

    return users;
  }
```
We have run into the issue of mismatch between cache memory and disk memory. The loadPolicy() was used to solve that issue. However, adding loadPolicy() does not seem to help. We still need two calls to this function to get fresh results from DB regardless whether loadPolicy() is used or not.

hsluoyz:
@xingzem 

1. Create an issue instead of commenting in closed PR
2. Go to jCasbin repo instead of Golang Casbin
3. Loading policy is highly related to the adapter you used. Name your adapter, also try to switch to another adapter to see if issue exists

This PR will be locked to avoid distraction.

PR #1203, opened 2023-03-01
fix: make the default cache expirable

### LINKED ISSUE #1202: CachedEnforcer does not expire keys
**Want to prioritize this issue? Try:**

[issuehunt-to-marktext](https://issuehunt.io/r/casbin/casbin)

------

**Describe the bug**
We are using cached enforcer to perform authorization. Authorization along with decision is getting cached as per the behaviour of cached enforcer. But Problem is key's in cache are not getting expire.

Our code stub for initialising cached enforcer instance and setting expiry TTL for cache keys.

```
enforcer, err := casbin.NewCachedEnforcer(configPath, RBACLenderDashboardSvc.Adapter)
if err != nil {
	return err
}

enforcer.EnableCache(true)
enforcer.SetExpireTime(10)
```

**First question - We have set expire time value as 10, this TTL is considered in seconds or milliseconds or hour or other ?**

So we deep dived in casbin golang library code and here is the code flow of cached enforcer Authoriser(Enforce method)

```
func (e *CachedEnforcer) Enforce(rvals ...interface{}) (bool, error) {
if atomic.LoadInt32(&e.enableCache) == 0 {
	return e.Enforcer.Enforce(rvals...)
}

key, ok := e.getKey(rvals...)
if !ok {
	return e.Enforcer.Enforce(rvals...)
}

if res, err := e.getCachedResult(key); err == nil {
	return res, nil
} else if err != cache.ErrNoSuchKey {
	return res, err
}

res, err := e.Enforcer.Enforce(rvals...)
if err != nil {
	return false, err
}

err = e.setCachedResult(key, res, e.expireTime)
return res, err
}
```

If we look in above **Enforce** method,  we see we are calling **setCachedResult** method to save result to cache, and we are passing the expireTime in it. 

```
func (e *CachedEnforcer) setCachedResult(key string, res bool, extra ...interface{}) error {
e.locker.Lock()
defer e.locker.Unlock()
return e.cache.Set(key, res, extra...)
}
```

**setCachedResult** method in turn calls to **Set** method with all parameter's. 

```
func (c *DefaultCache) Set(key string, value bool, extra ...interface{}) error {
(*c)[key] = value
return nil
}
```

If we take a look at  **Set** method, we can spot that expireTime value was never used and therefore TTL was never applied to cache key's. Thus preventing key's in cache to expire and make the cache stale. 

**To Reproduce**

**Expected behavior**
Cache key's should expire when TTL is hit.

JalinWang:
fix #1202

### DIFF
--- a/enforcer_cached.go
+++ b/enforcer_cached.go
@@ -18,6 +18,7 @@ import (
 	"strings"
 	"sync"
 	"sync/atomic"
+	"time"
 
 	"github.com/casbin/casbin/v2/persist/cache"
 )
@@ -25,7 +26,7 @@ import (
 // CachedEnforcer wraps Enforcer and provides decision cache
 type CachedEnforcer struct {
 	*Enforcer
-	expireTime  uint
+	expireTime  time.Duration
 	cache       cache.Cache
 	enableCache int32
 	locker      *sync.RWMutex
@@ -45,8 +46,7 @@ func NewCachedEnforcer(params ...interface{}) (*CachedEnforcer, error) {
 	}
 
 	e.enableCache = 1
-	cache := cache.DefaultCache(make(map[string]bool))
-	e.cache = &cache
+	e.cache, _ = cache.NewDefaultCache()
 	e.locker = new(sync.RWMutex)
 	return e, nil
 }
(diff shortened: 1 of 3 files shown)


hsluoyz:
@yugant007 plz review

yugant007:
LGTM

yugant007 (approved):

Issue #1204, opened 2023-03-01
[Bug] Implicit role doesn't work in go-lang lib but work in casbin editor


longsunli:
**Want to prioritize this issue? Try:**

[![issuehunt-to-marktext](https://github.com/BoostIO/issuehunt-materials/raw/master/v1/issuehunt-button-v1.svg)](https://issuehunt.io/r/casbin/casbin)

------

**Describe the bug**
We are using RBAC with domain model  to do authorization. And using group policy to extend one role to another. 

In the casbin editor below we can see that `can_manage` extends `can_use` role in all domain, and user with `can_manage` role can also perform `can_use` action like attach. 

https://editor.casbin.org/#DQV237WAL

<img width="1563" alt="image" src="https://user-images.githubusercontent.com/17940602/222183677-539cc0f7-b36f-4de8-bff7-b68cf5f82ded.png">

But same logic doesn't work in casbin go-lang lib even latest  version `v2.64.0`

**To Reproduce**
Code example

```
package main

import (
	"log"

	"github.com/casbin/casbin/v2"
	casbinModel "github.com/casbin/casbin/v2/model"
)

const (
	modelText = `
[request_definition]
r = sub, dom, obj, act

[policy_definition]
p = sub, dom, obj, act

[role_definition]
g = _, _, _

[policy_effect]
e = some(where (p.eft == allow))

[matchers]
#RBAC
m = g(r.sub, p.sub, r.dom) && keyMatch(r.dom, p.dom) && keyMatch(r.obj, p.obj) && regexMatch(r.act, p.act) 
`
)

func main() {

	m, err := casbinModel.NewModelFromString(modelText)
	if err != nil {
		panic(err)
	}
	e, err := casbin.NewSyncedEnforcer(m)
	if err != nil {
		panic(err)
	}

	e.AddPolicy("can_manage", "engines/*", "*", "(pause)|(resume)")
	e.AddPolicy("can_use", "engines/*", "*", "(attach)|(detach)")

	e.AddGroupingPolicy("can_manage", "can_use", "*")
	e.AddGroupingPolicy("Username==test2", "can_manage", "engines/engine1")

	implicitRoles, _ := e.GetImplicitRolesForUser("Username==test2", "engines/engine1")
	log.Println(implicitRoles)
        // can_manage

	ok, _, _ := e.EnforceEx("Username==test2", "engines/engine1", "*", "attach")
	log.Println(ok)
        // false
}

```

**Expected behavior**
[...]

PR #1223, opened 2023-04-04
fix: fix keymatch with rbac in domain

imp2002:
Fix: #1204 

image


hsluoyz:
@longsunli plz review

Issue #1252, opened 2023-05-16
[Bug] Large amount of memory and CPU allocation causing system OOM


kooksee:
**Want to prioritize this issue? Try:**

[![issuehunt-to-marktext](https://github.com/BoostIO/issuehunt-materials/raw/master/v1/issuehunt-button-v1.svg)](https://issuehunt.io/r/casbin/casbin)

------

**Describe the bug**
![image](https://github.com/casbin/casbin/assets/6829652/6ae2a2d0-6eb5-42e0-a59e-aa546f96811a)
<img width="1322" alt="image-20230421-090436" src="https://github.com/casbin/casbin/assets/6829652/82518f15-694d-474b-940b-8536c92e42a0">
![WX20230516-163200@2x](https://github.com/casbin/casbin/assets/6829652/f5e08e02-7be0-4178-9c4c-49a90c90bae2)

The policy in our production environment has approximately 1 million pieces of data, as shown in the above figure. Every instance of force requests a very large slice, which can easily cause system OOM in high concurrency situations. Moreover, monitoring shows that CPU and memory spikes are very obvious

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

PokIsemaine:
Thanks for your feedback, here are some of my thoughts.
1. Use smaller types
2. Don't allocate all at once, pre-allocate a part, and then take advantage of conditional branching to avoid unnecessary memory allocations
I'm not sure if it affects correctness as well as performance, it will take a while to verify

kooksee:
@PokIsemaine In our production environment, I made simple optimizations
- Transforming pre allocated slices into maps
- Filter unnecessary data using domain
- Object Pool Reuse Map [Not the most important thing]
- Because our model is deterministic, we can optimize it accordingly
- Currently, it is running well

hsluoyz:
@kooksee can you make a PR to add your optimizations?

kooksee:
@hsluoyz https://github.com/casbin/casbin/pull/1256
These are some of our adjustments, to emphasize, they have been optimized for our model, and this is not a general optimization

PR #1256, opened 2023-05-23
Some specific optimizations about the enforce function

kooksee:
- https://github.com/casbin/casbin/issues/1252


kooksee:
This PR is only a reference, it is an incomplete solution

hsluoyz:
@PokIsemaine ptal

Issue #1332, opened 2023-10-20
[Bug] LoadPolicyFast   perm not work


rentiansheng:
**Want to prioritize this issue? Try:**

<img width="630" alt="image" src="https://github.com/casbin/casbin/assets/3955934/df2d11de-d081-408a-b222-f2a4dfff46fe">

enforcer_synced.go  新add  	e.invalidateMatcherMap() can work

[![issuehunt-to-marktext](https://github.com/BoostIO/issuehunt-materials/raw/master/v1/issuehunt-button-v1.svg)](https://issuehunt.io/r/casbin/casbin)

------

**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

hsluoyz:
@rentiansheng can you make a PR to fix it?

rentiansheng:
> @rentiansheng can you make a PR to fix it?

yes, is testing. later commit

PR #1333, opened 2023-10-20
fix: LoadPolicyFast perm not work  #1332

rentiansheng:
sense:
Two nodes.  node A add Policy.  node B use LoadPolicyFast() reload data. but not work.
 
reason:

e.matcherMap not clean.  cache data effect verified.


hsluoyz:
@PokIsemaine plz review

hsluoyz:
Replaced by: https://github.com/casbin/casbin/pull/1402

PR #1378, opened 2024-03-27
fix: remove policy panic

caoxianfeng:


model/policy.go
```diff
@@ -247,7 +247,12 @@ func (model Model) RemovePolicy(sec string, ptype string, rule []string) bool {
 		return false
 	}
 
-	model[sec][ptype].Policy = append(model[sec][ptype].Policy[:index], model[sec][ptype].Policy[index+1:]...)
+	if index < len(model[sec][ptype].Policy)-1 {
```
hsluoyz, line 250: What does this code mean?

hsluoyz (changes requested):

hsluoyz:
@caoxianfeng

model/policy.go
caoxianfeng: Ensure that the index exists in the slice, and using copy is more efficient than append and can reduce memory allocation.

caoxianfeng: @hsluoyz

caoxianfeng: image

hsluoyz: Use append instead of copy. Readability is most important

hsluoyz (changes requested):

hsluoyz:
@MuZhou233 plz review

PR #1389, opened 2024-05-03
feat: add error to some of the APIs

### LINKED ISSUE #1369: [Bug] nil pointer panic when calling role related functions with no role definition model
**Describe the bug**
Currently `role_definition` is an optional section in the model definitions and the role assertions (`model["g"]`) won't be initialized if it is omited. However, the other parts of the code won't check if it is initialized before use and result in an nil pointer panic (e.g. some lines directly read `model["g"]["g"]` or even more without check). This affects a wide range of public and private functions.

**To Reproduce**
1. Initialize the model with no `role_definition` section
2. Call a function that related to role
3. Some of them may return, most of them will panic

**Expected behavior**
The panic should not be happen. Involve nil checks to return an error rather than panic.

**Additional context**
According to the documentation, this is a misuse. Following the guidelines would not encounter this issue.

MuZhou233:
Fix #1369 

This issue was paritially fixed by #1371, and this PR contains a fully fix.

- Add nil check on `model` before anywhere `model["g"]["g"]` may be accessed.
- **Public API changed**

### DIFF
--- a/management_api_test.go
+++ b/management_api_test.go
@@ -20,9 +20,12 @@ import (
 	"github.com/casbin/casbin/v2/util"
 )
 
-func testStringList(t *testing.T, title string, f func() []string, res []string) {
+func testStringList(t *testing.T, title string, f func() ([]string, error), res []string) {
 	t.Helper()
-	myRes := f()
+	myRes, err := f()
+	if err != nil {
+		t.Error(err)
+	}
 	t.Log(title+": ", myRes)
 
 	if !util.ArrayEquals(res, myRes) {
@@ -41,7 +44,10 @@ func TestGetList(t *testing.T) {
 
 func testGetPolicy(t *testing.T, e *Enforcer, res [][]string) {
 	t.Helper()
-	myRes := e.GetPolicy()
+	myRes, err := e.GetPolicy()
+	if err != nil {
+		t.Error(err)
+	}
 	t.Log("Policy: ", myRes)
 
 	if !util.Array2DEquals(res, myRes) {
@@ -51,7 +57,10 @@ func testGetPolicy(t *testing.T, e *Enforcer, res [][]string) {
 
 func testGetFilteredPolicy(t *testing.T, e *Enforcer, fieldIndex int, res [][]string, fieldValues ...string) {
 	t.Helper()
-	myRes := e.GetFilteredPolicy(fieldIndex, fieldValues...)
+	myRes, err := e.GetFilteredPolicy(fieldIndex, fieldValues...)
+	if err != nil {
+		t.Error(err)
+	}
 	t.Log("Policy for ", util.ParamsToString(fieldValues...), ": ", myRes)
 
 	if !util.Array2DEquals(res, myRes) {
@@ -75,7 +84,10 @@ func testGetFilteredNamedPolicyWithMatcher(t *testing.T, e *Enforcer, ptype stri
 
 func testGetGroupingPolicy(t *testing.T, e *Enforcer, res [][]string) {
 	t.Helper()
-	myRes := e.GetGroupingPolicy()
+	myRes, err := e.GetGroupingPolicy()
+	if err != nil {
+		t.Error(err)
+	}
 	t.Log("Grouping policy: ", myRes)
 
 	if !util.Array2DEquals(res, myRes) {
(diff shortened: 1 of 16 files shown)


model/policy.go
```diff
@@ -127,11 +127,17 @@ func (model Model) ClearPolicy() {
 
 // GetPolicy gets all rules in a policy.
 func (model Model) GetPolicy(sec string, ptype string) [][]string {
+	if model[sec] == nil || model[sec][ptype] == nil {
```
hsluoyz, line 130: Is this fix good? I think it's more of hiding problem

rbac_api_with_domains.go
```diff
@@ -76,14 +76,17 @@ func (e *Enforcer) DeleteRolesForUserInDomain(user string, domain string) (bool,
 }
 
 // GetAllUsersByDomain would get all users associated with the domain.
-func (e *Enforcer) GetAllUsersByDomain(domain string) []string {
+func (e *Enforcer) GetAllUsersByDomain(domain string) ([]string, error) {
+	if e.model["g"] == nil || e.model["g"]["g"] == nil {
```
hsluoyz, line 80: Put error checking logic together into one function

model/policy.go
```diff
@@ -177,6 +186,9 @@ func (model Model) HasPolicyEx(sec string, ptype string, rule []string) (bool, e
 
 // HasPolicy determines whether a model has the specified policy rule.
 func (model Model) HasPolicy(sec string, ptype string, rule []string) bool {
+	if model[sec] == nil || model[sec][ptype] == nil {
+		return false
```
hsluoyz, line 190: can we stop checking "p", let's focus on "g" first

model/model.go
```diff
@@ -236,6 +236,9 @@ func (model Model) SortPoliciesBySubjectHierarchy() error {
 	if model["e"]["e"].Value != constant.SubjectPriorityEffect {
 		return nil
 	}
+	if model["g"] == nil || model["g"]["g"] == nil {
+		return errors.New("role definition is not initialized")
```
hsluoyz, line 240: these are two different errors

hsluoyz (changes requested):

model/policy.go
```diff
@@ -217,15 +247,21 @@ func (model Model) AddPolicy(sec string, ptype string, rule []string) {
...
+	_, err := model.AddPoliciesWithAffected(sec, ptype, rules)
+	return err
 }
 
 // AddPoliciesWithAffected adds policy rules to the model, and returns affected rules.
-func (model Model) AddPoliciesWithAffected(sec string, ptype string, rules [][]string) [][]string {
+func (model Model) AddPoliciesWithAffected(sec string, ptype string, rules [][]string) ([][]string, error) {
+	_, err := model.GetAssertion(sec, ptype)
+	if err != nil {
+		return [][]string{}, err
```
hsluoyz, line 263: if err is not nil, the first return value should be nil

model/policy.go
```diff
@@ -234,17 +270,24 @@ func (model Model) AddPoliciesWithAffected(sec string, ptype string, rules [][]s
 			continue
 		}
 		affected = append(affected, rule)
-		model.AddPolicy(sec, ptype, rule)
+		err = model.AddPolicy(sec, ptype, rule)
+		if err != nil {
+			break
```
hsluoyz, line 275: why isn't it `return affected, err`?

management_api_test.go
```diff
@@ -41,7 +44,10 @@ func TestGetList(t *testing.T) {
 
 func testGetPolicy(t *testing.T, e *Enforcer, res [][]string) {
 	t.Helper()
-	myRes := e.GetPolicy()
+	myRes, err := e.GetPolicy()
+	if err != nil {
+		t.Error(err)
+	}
```
hsluoyz, line 50: new line

model/policy.go
```diff
@@ -362,31 +421,40 @@ func (model Model) RemoveFilteredPolicy(sec string, ptype string, fieldIndex int
...
-func (model Model) GetValuesForFieldInPolicy(sec string, ptype string, fieldIndex int) []string {
+func (model Model) GetValuesForFieldInPolicy(sec string, ptype string, fieldIndex int) ([]string, error) {
 	values := []string{}
 
+	_, err := model.GetAssertion(sec, ptype)
+	if err != nil {
+		return nil, err
+	}
+
 	for _, rule := range model[sec][ptype].Policy {
```
hsluoyz, line 436: use return value

hsluoyz (changes requested):

hsluoyz (approved):

PR #1410, opened 2024-06-11
fix: fixed incorrect results of methods in the Management API in RBAC With Domains

Dobmod:
Using the `GetFieldIndex` method instead of the fixed FieldIndex, so fixed incorrect results of GetAllSubjects,GetAllObjects,GetAllActions, etc.in the Management API in RBAC With Domains and added a test case.😁


hsluoyz:
Replaced by: https://github.com/casbin/casbin/pull/1411

PR #1413, opened 2024-06-15
feat: CachedEnforcer does not clean the result cached in memory when call ClearPolicy method

cococolanosugar:
CachedEnforcer does not clean the result cached in memory when call ClearPolicy method


hsluoyz:
@cococolanosugar fix the two errors:

image

image

cococolanosugar:
review again please.

PR #1465, opened 2024-12-17
feat: Enforcer config & DefaultLogger customization

28Pollux28:
* Use a config struct for Enforcer creation
* Allow for passing a log.Logger to casbin DefaultLogger

You are now allowed to pass a custom log.Logger to the casbin DefaultLogger. If nil, it'll fallback onto the log.Default() logger (which is the current behavior). Function to init a DefaultLogger is now : casbinlog.NewDefaultLogger(log.Logger)

Refactored NewEnforcer method to a cleaner NewEnforcerWithConfig. This new method takes a EnforcerConfig struct to store the different options for the Enforcer. Old function edited to use the new method, but is still available for backward compatibility.

PR was supposed to be about logs, but I stumbled upon the ugliness of the NewEnforcer function, and went for a small refactor.
Feel free to add or suggest edits.
Pollux

### DIFF
--- a/enforcer.go
+++ b/enforcer.go
@@ -78,64 +78,9 @@ type EnforcerConfig struct {
 	EnableLog  bool
 }
 
-// NewEnforcerWithConfig creates an enforcer via file or DB with a configuration.
-// Specifying a Model will override the ModelPath.
-//
-// File:
-//
-//	e := casbin.NewEnforcerWithConfig(casbin.EnforcerConfig{ModelPath: "path/to/basic_model.conf", PolicyPath: "path/to/basic_policy.csv"})
-//
-// MySQL DB:
-//
-//	a := mysqladapter.NewDBAdapter("mysql", "mysql_username:mysql_password@tcp(127.0.0.1:3306)/")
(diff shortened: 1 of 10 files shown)


hsluoyz:
@28Pollux28 you changed too much code for such a trivial thing. This PR will not pass.

28Pollux28:
@hsluoyz maybe you should go hit the ask ai button about your New enforcer function, see if it applies the best go practices. 
These changes were to make Casbin more straightforward to use, and thus lower the entry knowledge necessary to start using Casbin.

hsluoyz:
@28Pollux28 it's not a good idea that someone new proposes a big PR to refactor the code for the first time. Even some very senior people here don't have this capability to control the code quality

28Pollux28:
@hsluoyz Ok I get that, that's the reason I encouraged you to suggest edits and to discuss around it, but you just simply closed the PR. By no means I believe I have sufficient knowledge of Casbin or even Go to make changes and not have them challenged, edited, refused, but at least I want to understand the reason behind such actions. The refactor was not that huge of a change, kept the old behaviors alive, passed all the tests (I would have been happy to write more tests to support the PR). Code quality was not happy because of the use of the deprecated function in the tests, but that can easily be changed

PR #1475, opened 2025-03-12
feat: fix addPolicyWithoutNotify()'s affected value

SMALL-head:
The `AddPolicy()` function's comment states that if the enforcer already contains the same policy as the one being added, `AddPolicy()` should `return false, nil`. However, the current implementation does not align with this behavior. This PR updates the code to ensure consistency with the function's intended behavior.


hsluoyz:
@SMALL-head you should refer to other addPolicy() APIs, and other Casbin languages

PR #784, opened 2021-05-04
feat: add new rbac api

closetool:
Signed-off-by: closetool <4closetool3@gmail.com>

Fix: #782

PR #1496, opened 2025-07-05
feat: completely remove domain from RoleManager when calling DeleteDomains

### LINKED ISSUE #1492: [Bug] RBAC with domains: `DeleteDomains` does not remove the domain itself
**Describe the bug**
The RBAC with domains API is defined like so:

```go
// DeleteDomains would delete all associated users and roles.
// It would delete all domains if parameter is not provided.
func (e *Enforcer) DeleteDomains(domains ...string) (bool, error) {
	if len(domains) == 0 {
		e.ClearPolicy()
		return true, nil
	}
	for _, domain := range domains {
		if _, err := e.DeleteAllUsersByDomain(domain); err != nil {
			return false, err
		}
	}
	return true, nil
}
```
This deletes all users by domain, but the domain itself is not deleted. This is verifiable by running `DeleteDomains(domain)` followed by `GetAllDomains()` and the original `domain` would still be present.

**To Reproduce**
1. Add policies that mention a domain
2. run `DeleteDomains`
3. run `GetAllDomains`
4. the deleted domain will still be present

**Expected behavior**
The `domain` deleted by `DeleteDomains` should not be in the output of `GetAllDomains`

**Notes**:

This issue can be closed if this is expected behavior. [This is a minimal reproducer go project here](https://tangled.sh/@oppi.li/casbin-repro/blob/main/main.go). It can be run like so:

```
λ git clone https://tangled.sh/@oppi.li/casbin-repro

λ cd casbin-repro

λ go run .
=== casbin issue repro ===

1. adding policies and users...
2. domains before removal:
   [server-foo]
3. users with server:member role before removal:
   [server:owner user-b]
4. Calling DeleteDomains('server-foo')...
5. Domains after removal:
   [server-foo]
6. users with server:member role after removal:
   []

 bug: expected 0 domains, got 1: [server-foo]
   DeleteDomains() only removes users/roles, not the domain itself!
```

D0000M:
close #1492 
The original `DeleteDomains()` only performed the task of `DeleteAllUsersByDomain()`, but now `DeleteDomains() ` will delete all associated policies as well.

### DIFF
--- a/rbac/default-role-manager/role_manager.go
+++ b/rbac/default-role-manager/role_manager.go
@@ -15,6 +15,7 @@
 package defaultrolemanager
 
 import (
+	"errors"
 	"fmt"
 	"strings"
 	"sync"
@@ -717,6 +718,12 @@ func (dm *DomainManager) BuildRelationship(name1 string, name2 string, domain ..
 	return nil
 }
 
+// DeleteDomain deletes the specified domain from DomainManager.
+func (dm *DomainManager) DeleteDomain(domain string) error {
+	dm.rmMap.Delete(domain)
+	return nil
+}
+
 type RoleManager struct {
 	*DomainManager
 }
@@ -725,6 +732,11 @@ func NewRoleManager(maxHierarchyLevel int) *RoleManager {
 	rm := &RoleManager{}
 	rm.DomainManager = NewDomainManager(maxHierarchyLevel)
 	return rm
+}
+
+// DeleteDomain does nothing for RoleManagerImpl (no domain concept).
+func (rm *RoleManagerImpl) DeleteDomain(domain string) error {
+	return errors.New("DeleteDomain is not supported by RoleManagerImpl (no domain concept)")
 }
 
 type ConditionalRoleManager struct {
--- a/rbac_api_with_domains.go
+++ b/rbac_api_with_domains.go
@@ -144,16 +144,25 @@ func (e *Enforcer) DeleteAllUsersByDomain(domain string) (bool, error) {
 	return true, nil
 }
 
-// DeleteDomains would delete all associated users and roles.
+// DeleteDomains would delete all associated policies.
 // It would delete all domains if parameter is not provided.
 func (e *Enforcer) DeleteDomains(domains ...string) (bool, error) {
 	if len(domains) == 0 {
-		e.ClearPolicy()
-		return true, nil
+		var err error
+		domains, err = e.GetAllDomains()
+		if err != nil {
+			return false, err
+		}
 	}
 	for _, domain := range domains {
 		if _, err := e.DeleteAllUsersByDomain(domain); err != nil {
 			return false, err
+		}
+		// remove the domain from the RoleManager.
+		if e.GetRoleManager() != nil {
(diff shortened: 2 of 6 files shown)


hsluoyz:
@oppiliappan plz review

hsluoyz:
@tangyang9464 plz review

tangyang9464:
@D0000M 
I looked through the historical commits ((#784 #790 )) of the `DeleteDomains`-related APIs and  found that their function is to delete both the `g` and `p` policies (the comments and documentation of the related APIs are a bit vague, and they describe it as `would delete all associated users and roles`). 

The key problem of this issue #1492 is that `DomainRoleManager` is a later implementation that does not take into account the `DeleteDomains`-related APIs. 
Therefore, I think this fix should only require adding the deletion of `RoleManager` in the `DeleteDomains` and other APIs, and there is no need to modify the deletion logic of `DeleteDomains` for the policy, because it is correct.

rbac_api_with_domains.go
```diff
@@ -144,17 +144,35 @@ func (e *Enforcer) DeleteAllUsersByDomain(domain string) (bool, error) {
...
 // It would delete all domains if parameter is not provided.
 func (e *Enforcer) DeleteDomains(domains ...string) (bool, error) {
 	if len(domains) == 0 {
-		e.ClearPolicy()
+		allDomains, _ := e.GetAllDomains()
+		for _, domain := range allDomains {
+			if _, err := e.DeleteDomains(domain); err != nil {
+				return false, err
+			}
+		}
```
tangyang9464, line 156: The logic here can be combined with the following, just use `e. GetAllDomains `at the beginning. 
such as
```go
allDomains := domains
if len(domains) == 0 {
    allDomains, _ := e.GetAllDomains()
}
ffor _, domain := range allDomains {
```

rbac_api_with_domains.go
```diff
@@ -144,17 +144,35 @@ func (e *Enforcer) DeleteAllUsersByDomain(domain string) (bool, error) {
...
+		if err != nil {
+			return false, err
+		}
+		// delete all policies containing the domain.
+		if _, err := e.RemoveFilteredPolicy(domainIndex, domain); err != nil {
+			return false, err
+		}
+		// remove the domain from the RoleManager.
+		if e.GetRoleManager() != nil {
+			e.GetRoleManager().DeleteDomain(domain)
```
tangyang9464, line 174: err handle

rbac/default-role-manager/role_manager.go
```diff
@@ -727,6 +733,16 @@ func NewRoleManager(maxHierarchyLevel int) *RoleManager {
...
+// DeleteDomain does nothing for RoleManagerImpl (no domain concept).
+func (rm *RoleManagerImpl) DeleteDomain(domain string) error {
+	return nil
+}
+
+// DeleteDomain proxies to DomainManager.
+func (rm *RoleManager) DeleteDomain(domain string) error {
+	return rm.DomainManager.DeleteDomain(domain)
+}
+
```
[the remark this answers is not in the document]
D0000M: smell good

tangyang9464 (approved):
LGTM

Issue #1404, opened 2024-06-08
[Bug] Role hierarchy level behaviour inconsist with document


MuZhou233:
**Describe the bug**

In [RBAC doc](https://casbin.org/docs/rbac#role-hierarchy), we have following description:

> Casbin's RBAC supports RBAC1's role hierarchy feature
> ...
> For the built-in role manager in Casbin, you can specify the maximum hierarchy level.

But actually `maxHierarchyLevel` only affects `HasLink`. 
Other functions like `GetImplicitRolesForUser` and `GetImplicitUsersForRole` ignore it and seeking infinity hierachy level.

**To Reproduce**

See description

**Expected behavior**

It would be better to have an consist behaviour.

hsluoyz:
OK

PR #1497, opened 2025-07-08
fix:align RoleManager behavior with documentation(#1404)

bugoutianzhen123:

It ensures that maxHierarchyLevel not only limits the depth of the HasLink function, but also applies to other role  functions such as:

GetImplicitRolesForUser
GetImplicitUsersForRole


hsluoyz:
Replaced by: https://github.com/casbin/casbin/pull/1499

PR #1498, opened 2025-07-08
feat:align RoleManager behavior with documentation(#1404)

bugoutianzhen123:
It ensures that maxHierarchyLevel not only limits the depth of the HasLink function, but also applies to other role traversal  such as:
GetImplicitRolesForUser
GetImplicitUsersForRole

Why this change is necessary
Previously, maxHierarchyLevel only affected HasLink, while other functions could traverse indefinitely, which led to inconsistencies between the documented behavior and actual implementation.
This update:
Makes the behavior of RoleManager consistent with the documentation
Keeps the change self-contained within RoleManager, without requiring modifications to external components
Ensures all relevant traversal functions now honor the maxHierarchyLevel setting

Summary of Changes
Added level field to Role struct
Dynamically maintained level values within RoleManager
Applied hierarchy level checks consistently across relevant APIs

### DIFF
--- a/rbac/default-role-manager/role_manager.go
+++ b/rbac/default-role-manager/role_manager.go
@@ -15,6 +15,7 @@
 package defaultrolemanager
 
 import (
+	"errors"
 	"fmt"
 	"strings"
 	"sync"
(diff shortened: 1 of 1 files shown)


hsluoyz:
Replaced by: https://github.com/casbin/casbin/pull/1499

PR #1499, opened 2025-07-08
fix: align RoleManager behavior with documentation (#1404)

bugoutianzhen123:
It ensures that maxHierarchyLevel not only limits the depth of the HasLink function, but also applies to other role traversal  such as:
GetImplicitRolesForUser
GetImplicitUsersForRole

Why this change is necessary
Previously, maxHierarchyLevel only affected HasLink, while other functions could traverse indefinitely, which led to inconsistencies between the documented behavior and actual implementation.
This update:
Makes the behavior of RoleManager consistent with the documentation
Keeps the change self-contained within RoleManager, without requiring modifications to external components
Ensures all relevant traversal functions now honor the maxHierarchyLevel setting

Summary of Changes
Added level field to Role struct
Dynamically maintained level values within RoleManager
Applied hierarchy level checks consistently across relevant APIs

### DIFF
--- a/rbac/default-role-manager/role_manager.go
+++ b/rbac/default-role-manager/role_manager.go
@@ -331,8 +331,35 @@ func (rm *RoleManagerImpl) Clear() error {
 func (rm *RoleManagerImpl) AddLink(name1 string, name2 string, domains ...string) error {
 	user, _ := rm.getRole(name1)
 	role, _ := rm.getRole(name2)
+
+	if rm.hasPath(role, user) {
+		return fmt.Errorf("cannot add link from '%s' to '%s': would create a cycle", name1, name2)
+	}
 	user.addRole(role)
+
 	return nil
+}
+
+func (rm *RoleManagerImpl) hasPath(from, to *Role) bool {
+	visited := make(map[string]bool)
+	var dfs func(r *Role) bool
+	dfs = func(r *Role) bool {
+		if r.name == to.name {
+			return true
+		}
+		visited[r.name] = true
+		found := false
+		r.roles.Range(func(_, value interface{}) bool {
+			next := value.(*Role)
+			if !visited[next.name] && dfs(next) {
+				found = true
+				return false
+			}
+			return true
+		})
+		return found
+	}
+	return dfs(from)
 }
 
 // DeleteLink deletes the inheritance link between role: name1 and role: name2.
@@ -392,6 +419,7 @@ func (rm *RoleManagerImpl) GetRoles(name string, domains ...string) ([]string, e
 	if created {
 		defer rm.removeRole(user.name)
 	}
+
 	return user.getRoles(), nil
 }
 
@@ -402,7 +430,58 @@ func (rm *RoleManagerImpl) GetUsers(name string, domain ...string) ([]string, er
 	if created {
 		defer rm.removeRole(role.name)
 	}
+
 	return role.getUsers(), nil
+}
+
+// GetImplicitRoles gets all roles that a user inherits, including direct and indirect roles.
+func (rm *RoleManagerImpl) GetImplicitRoles(name string, domains ...string) ([]string, error) {
+	return rm.getImplicitElements(name, func(r *Role, fn func(key, value interface{}) bool) {
+		r.rangeRoles(fn)
+	}, domains...)
+}
+
+// GetImplicitUsers gets all users that inherit a role, including direct and indirect users.
+func (rm *RoleManagerImpl) GetImplicitUsers(name string, domains ...string) ([]string, error) {
+	return rm.getImplicitElements(name, func(r *Role, fn func(key, value interface{}) bool) {
+		r.rangeUsers(fn)
+	}, domains...)
+}
+
+func (rm *RoleManagerImpl) getImplicitElements(name string, rf func(*Role, func(key, value interface{}) bool), domains ...string) ([]string, error) {
+	element, created := rm.getRole(name)
+	if created {
+		defer rm.removeRole(element.name)
+	}
+
+	result := make([]string, 0, 10)
+	visited := make(map[string]bool)
+
+	queue := []*Role{element}
+	level := rm.maxHierarchyLevel
+
+	for len(queue) > 0 && level > 0 {
+		size := len(queue)
+		for i := 0; i < size; i++ {
+			currentRole := queue[0]
+			queue = queue[1:]
+
+			rf(currentRole, func(key, value interface{}) bool {
+				elementName := key.(string)
+				elementValue := value.(*Role)
+
+				if !visited[elementName] {
+					visited[elementName] = true
+					result = append(result, elementName)
+					queue = append(queue, elementValue)
+				}
+				return true
+			})
+		}
+		level--
+	}
+
+	return result, nil
 }
 
 func (rm *RoleManagerImpl) toString() []string {
@@ -605,12 +684,17 @@ func (dm *DomainManager) AddLink(name1 string, name2 string, domains ...string) 
 		return err
 	}
 	roleManager := dm.getRoleManager(domain, true) // create role manager if it does not exist
-	_ = roleManager.AddLink(name1, name2, domains...)
+	if linkErr := roleManager.AddLink(name1, name2, domains...); linkErr != nil {
+		return linkErr
+	}
 
+	var managerErr error
 	dm.rangeAffectedRoleManagers(domain, func(rm *RoleManagerImpl) {
-		_ = rm.AddLink(name1, name2, domains...)
+		if managerErr == nil {
+			managerErr = rm.AddLink(name1, name2, domain)
+		}
 	})
-	return nil
+	return managerErr
 }
 
 // DeleteLink deletes the inheritance link between role: name1 and role: name2.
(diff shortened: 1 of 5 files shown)


hsluoyz:
@tangyang9464 plz review

tangyang9464:
@bugoutianzhen123 Tests need to be added

rbac/default-role-manager/role_manager.go
```diff
@@ -46,12 +48,41 @@ func newRole(name string) *Role {
...
+	if newLevel != r.level {
+		r.level = newLevel
+		r.roles.Range(func(_, value interface{}) bool {
+			role := value.(*Role)
+			role.updateLevel()
+			return true
+		})
+	}
+}
+
```
tangyang9464, line 81: Consider streamlining redundant local variables and overall processes
example:
```go
func (r *Role) updateLevel() {
	maxLevel := 0
	r.users.Range(func(_, value interface{}) bool {
		if parentRole := value.(*Role); parentRole.level >= 0 && parentRole.level+1 > maxLevel {
			maxLevel = parentRole.level + 1
		}
		return true
	})
	if r.level == maxLevel {
		return
	}
	r.level = maxLevel
	r.roles.Range(func(_, value interface{}) bool {
		value.(*Role).updateLevel()
		return true
	})
}
```

Also, what if the role map has a ring?

rbac/default-role-manager/role_manager.go
```diff
@@ -331,6 +362,9 @@ func (rm *RoleManagerImpl) Clear() error {
 func (rm *RoleManagerImpl) AddLink(name1 string, name2 string, domains ...string) error {
 	user, _ := rm.getRole(name1)
 	role, _ := rm.getRole(name2)
+	if rm.maxHierarchyLevel > 0 && role.level >= rm.maxHierarchyLevel-1 {
+		return errors.New("exceeded level limit")
```
tangyang9464, line 366: Need more detailed error message

tangyang9464 (changes requested):

bugoutianzhen123:
Thank you for the previous review!
I’ve updated the PR with the following improvements:

Added cycle detection to prevent circular role inheritance;
Added corresponding test cases;
Optimized some internal logic for better readability and flow;
Improved error messages for better debugging;
Also, I noticed that the TestMaxHierarchyLevel test previously assumed that the setting limits the "difference in role levels." However, based on the documentation and common usage, maxHierarchyLevel should limit the length of the inheritance chain. I've adjusted the test and logic accordingly.

Please take another look when convenient

rbac/default-role-manager/role_manager.go
```diff
@@ -46,17 +47,40 @@ func newRole(name string) *Role {
...
+		value.(*Role).updateLevel()
+		return true
+	})
+}
+
 func (r *Role) addRole(role *Role) {
 	r.roles.Store(role.name, role)
 	role.addUser(r)
+	r.updateLevel()
+	role.updateLevel()
```
tangyang9464, line 76: Why is there an extra line? It seems to be repeated.
In method `updateLevel`
```go
r.roles.Range(func(_, value interface{}) bool {
		value.(*Role).updateLevel()
		return true
	})
```

rbac/default-role-manager/role_manager.go
```diff
@@ -329,9 +355,21 @@ func (rm *RoleManagerImpl) Clear() error {
 // AddLink adds the inheritance link between role: name1 and role: name2.
 // aka role: name1 inherits role: name2.
 func (rm *RoleManagerImpl) AddLink(name1 string, name2 string, domains ...string) error {
+	if hasLink, err := rm.HasLink(name2, name1, domains...); err != nil {
+		return err
+	} else if hasLink {
+		return fmt.Errorf("cannot add link from '%s' to '%s': would create a cycle", name1, name2)
+	}
```
tangyang9464, line 362: It is best not to use `HasLink` for loop detection, because it not only includes normal role inheritance, but also pattern matching, which does not match the logic of method `updateLevel`. I think it is better to use `visit[]` to mark in updateLevel

rbac/default-role-manager/role_manager.go
```diff
@@ -329,9 +355,21 @@ func (rm *RoleManagerImpl) Clear() error {
...
+
 	user, _ := rm.getRole(name1)
 	role, _ := rm.getRole(name2)
 	user.addRole(role)
+
+	if rm.maxHierarchyLevel > 0 && user.level > rm.maxHierarchyLevel-1 {
+		user.removeRole(role)
+		return fmt.Errorf("cannot add link from '%s' to '%s': role '%s' is at level %d, exceeding the maximum allowed hierarchy level of %d",
+			name1, name2, name2, role.level, rm.maxHierarchyLevel)
+	}
```
tangyang9464, line 372: Can it be detected before `addRole`?

rbac/default-role-manager/role_manager.go
```diff
@@ -392,7 +430,8 @@ func (rm *RoleManagerImpl) GetRoles(name string, domains ...string) ([]string, e
 	if created {
 		defer rm.removeRole(user.name)
 	}
-	return user.getRoles(), nil
+
+	return rm.filterHelper(user.getRoles()), nil
```
tangyang9464, line 434: Since the `level` has been restricted in `AddLink`, will `getRoles` still exceed the level? If not, is filtering unnecessary?

rbac/default-role-manager/role_manager.go
```diff
@@ -228,7 +252,9 @@ func (rm *RoleManagerImpl) rebuild() {
 	roles := rm.allRoles
 	_ = rm.Clear()
 	rangeLinks(roles, func(name1, name2 string, domain ...string) bool {
-		_ = rm.AddLink(name1, name2, domain...)
+		if err := rm.AddLink(name1, name2, domain...); err != nil {
+			rm.logger.LogError(err)
```
tangyang9464, line 256: This logic has little to do with PR, just keep it as it is

rbac/default-role-manager/role_manager.go
```diff
@@ -402,7 +441,37 @@ func (rm *RoleManagerImpl) GetUsers(name string, domain ...string) ([]string, er
 	if created {
 		defer rm.removeRole(role.name)
 	}
-	return role.getUsers(), nil
+
+	return rm.filterHelper(role.getUsers()), nil
+}
+
+func (rm *RoleManagerImpl) filterHelper(names []string) []string {
+	if rm.maxHierarchyLevel <= 0 {
```
tangyang9464, line 449: Why not just modify role.getUsers()

rbac/default-role-manager/role_manager_test.go
```diff
@@ -364,29 +364,47 @@ func TestTemporaryRoles(t *testing.T) {
 }
 
 func TestMaxHierarchyLevel(t *testing.T) {
-	rm := NewRoleManager(1)
```
tangyang9464, line 367: To maintain compatibility, I think it is best not to restrict AddLink by level, and only filter by level when gettingRoles/Uses and HasLink.

tangyang9464 (changes requested):

tangyang9464:
> TestMaxHierarchyLevel test previously assumed that the setting limits the "difference in role levels." However, based on the documentation and common usage, maxHierarchyLevel should limit the length of the inheritance chain. I've adjusted the test and logic accordingly.

In order to maintain compatibility, I think we only need to restrict `getRoles/Users`, `HasLink`, etc. For users, they can flexibly adjust maxLevel @bugoutianzhen123

rbac/default-role-manager/role_manager.go
bugoutianzhen123: This approach ensures that when two inheritance chains are merged into one, the levels can be correctly updated. I will try to optimize it.

bugoutianzhen123:
I believe it's better to separate concerns: AddLink should handle cycle detection, while updateLevel should be dedicated to maintaining the level structure. Level-based filtering should only be enforced in role resolution methods like getRoles, getUsers, and HasLink

hsluoyz:
@tangyang9464 plz review

rbac/default-role-manager/role_manager_test.go
```diff
@@ -364,29 +364,47 @@ func TestTemporaryRoles(t *testing.T) {
...
+	_ = rm2.AddLink("level1", "level2")
+	_ = rm2.AddLink("level2", "level3")
+
+	testRole(t, rm2, "level0", "level0", true)
+	testRole(t, rm2, "level0", "level1", true)
+	testRole(t, rm2, "level0", "level2", true)
+	testRole(t, rm2, "level0", "level3", false)
+	testRole(t, rm2, "level1", "level2", true)
+	testRole(t, rm2, "level1", "level3", false)
+	testRole(t, rm2, "level2", "level3", false)
```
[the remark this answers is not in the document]
bugoutianzhen123: The old TestMaxHierarchyLevel test previously assumed that the setting limits the "difference in role levels." However, the  new TestMaxHierarchyLevel think maxHierarchyLevel should limit the length of the inheritance chain.

tangyang9464:
@bugoutianzhen123  I think there are some problems with the current implementation

1. There is some deviation in the understanding of the `maxHierarchyLevel`. For the same role, different users have different inheritance depths. For example, 
```
alice -> sub_admin -> admin
bob -> admin
```
`alice -> admin` is 2 depth, but `bob -> admin` is 1 depth. So [TestMaxHierarchyLevel](https://github.com/casbin/casbin/pull/1499/files#diff-07dde4caad09fc8cd341a758cc6ddfbb87ee582f4739392e7eac4a34d130e0a6L367-L389) should not be modified


2. The logic of `HasLink` is quite special. It not only checks the role inheritance relationship, but also checks the pattern matching. This needs to be taken into consideration. It may not be compatible with your `level` design. 
`GetRoles` is also affected by pattern matching

bugoutianzhen123:
The original functions in role_manager.go ensure they are limited by maxHierarchyLevel when called individually, but when upstream functions make multiple calls, they might obtain results that exceed the maxHierarchyLevel limit. This risk primarily exists in functions that need to retrieve implicit roles or users. I abandoned the previous approach using level, and instead added functions in role_manager.go to get implicit roles or users for upstream calls, ensuring that the final results of upstream functions are also limited by maxHierarchyLevel.

hsluoyz:
I don't think this code will be OK if merged given the code quality shown here. Closed for now.

PR #1501, opened 2025-07-21
docs: add ReBAC model examples

D0000M:
add ReBAC model examples


examples/rebac_model.conf
```diff
@@ -0,0 +1,15 @@
+[request_definition]
```
hsluoyz, line 1: Why not add the TestRebac test case?

hsluoyz (changes requested):

Issue #1495, opened 2025-07-02
[Bug] Casbin - fatal error: concurrent map read and map write issue


randomAnalyst:
reopening this issue : https://github.com/casbin/casbin/issues/1468, As it is not solved in latest version.

@hsluoyz we are still getting this concurrent map read and map write issue for AddGroupingPolicy and RemoveGroupingPolicy the latest casbin version v2.108.0.
This is blocking us from concurrency tests.

fatal error: concurrent map read and map write

goroutine 9413 gp=0xc01732b6c0 m=106 mp=0xc000980008 [running]:
runtime.fatal({0x6f52fe6, 0x21})
/home/go/go/src/runtime/panic.go:1123 +0x5c fp=0xc0069080c0 sp=0xc006908090 pc=0x24582bc
internal/runtime/maps.fatal({0x6f52fe6?, 0xc006908150?})
/home/go/go/src/runtime/panic.go:1058 +0x18 fp=0xc0069080e0 sp=0xc0069080c0 pc=0x2492bf8
runtime.mapaccess2_faststr(0xc00e180060?, 0x75f2050?, {0xc0001f5980?, 0x75e5bd8?})
/home/go/go/src/internal/runtime/maps/runtime_faststr_swiss.go:174 +0x117 fp=0xc006908160 sp=0xc0069080e0 pc=0x241f6d7
github.com/casbin/casbin/v2/model.Model.HasPolicy(0xc000962870, {0x75f2050, 0x1}, {0x75f2050, 0x1}, {0xc00e180060, 0x3, 0x3})
/root/go/pkg/mod/github.com/casbin/casbin/v2@v2.103.0/model/policy.go:203 +0xd0 fp=0xc0069081a8 sp=0xc006908160 pc=0x5823550
github.com/casbin/casbin/v2.(*Enforcer).addPolicyWithoutNotify(0xc00117c0c0, {0x75f2050, 0x1}, {0x75f2050, 0x1}, {0xc00e180060, 0x3, 0x3})
/root/go/pkg/mod/github.com/casbin/casbin/v2@v2.103.0/internal_api.go:43 +0x115 fp=0xc006908230 sp=0xc0069081a8 pc=0x5844315
github.com/casbin/casbin/v2.(*Enforcer).addPolicy(0xc00117c0c0, {0x75f2050, 0x1}, {0x75f2050, 0x1}, {0xc00e180060, 0x3, 0x3})
[...]

PR #1511, opened 2025-08-04
feat: change PolicyMap from nomal map to sync.map (#1495)

D0000M:
feat: change PolicyMap from nomal map to sync.map (#1495), may  have some help.

I couldn't reproduce the issue on my side, but I suspect it might be due to the lock granularity in SyncedEnforcer, or maybe some methods are being called inside a locked context and then call other functions that aren't aware of the lock.

My proposed solution is to change Assertion.PolicyMap from a map[string]int to a sync.Map, which seems to fix the problem. But it does have a noticeable impact on performance—overall benchmark performance drops by about 3%, and BenchmarkAddPolicy and BenchmarkRemovePolicy drop by 10–46%.

So it's really a trade-off.

### DIFF
--- a/model/assertion.go
+++ b/model/assertion.go
@@ -17,6 +17,7 @@ package model
 import (
 	"errors"
 	"strings"
+	"sync"
 
 	"github.com/casbin/casbin/v2/log"
 	"github.com/casbin/casbin/v2/rbac"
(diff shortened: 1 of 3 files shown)


LarsArtmann:
Link/Related to: #1513

PR #1513, opened 2025-08-05
feat: Add RWMutex to Assertion to ensure safe concurrent map access(#…

bugoutianzhen123:
…1495)

### DIFF
--- a/model/assertion.go
+++ b/model/assertion.go
@@ -16,8 +16,11 @@ package model
 
 import (
 	"errors"
+	"strconv"
 	"strings"
+	"sync"
 
+	"github.com/casbin/casbin/v2/constant"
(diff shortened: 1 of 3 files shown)


bugoutianzhen123:
I was not able to reproduce the bug reported in issue #[1495].

However, I noticed that `policyMap` in `Assertion` may be accessed concurrently without proper synchronization.  
To address this, I introduced a `policyMu sync.RWMutex` to `Assertion` to protect concurrent reads/writes, which also helps ensure data consistency to some extent.

LarsArtmann:
Do we have any Benchmarks for this?
I am new here, but it looks like this is a hot path.
I am worried that this could tank the performance.

LarsArtmann:
Link/Related to: #1511

Issue #316, opened 2019-10-24
Allow empty string policy field


vzts:
Hello, I was giving [some update to SQLAlchemy adapter](https://github.com/pycasbin/sqlalchemy-adapter/pull/5), but I found out that there's no set standard on handling the empty string policy field.

I read https://github.com/casbin/casbin/issues/171#issuecomment-445434013 and the original author @hsluoyz doesn't seem to have a strong opinion on whether to allow empty string or not.

However, I think we should whether 1) do not allow empty string in the first place when writing rules, or 2) explicitly separate empty string and null in the persistence. This is because if we allow empty string on rules, we cannot differentiate whether it's a valid rule element when we do the loading.

Consider a following rule example:
```
'p', 'x', '', 'y', 'z'
```

Above can be persisted in several ways as below:
```
| v0 | v1   | v2 | v3 | v4   | v5   |
|----|------|----|----|------|------|
| x  |      | y  | z  | null | null |
| x  | null | y  | z  | null | null |
| x  |      | y  | z  |      |      |
```

The first option is okay, we can differentiate between empty and nil, so that when loading we can decide that it's `'x', '', 'y', 'z'`.
The second and third option is not okay, because we cannot differentiate between the two, which lead to a situation where we can't decide it should whether be `'x', '', 'y', 'z'` or `'x', '', 'y', 'z', '', ''`, or even `'x'`.. when loading.

If you look at current implementation of one most famous orm adapter(xorm-adapter in Go), this behavior is actually giving a false result.
[Here](https://github.com/casbin/xorm-adapter/blob/master/adapter.go#L28), it sets the field as nonnull, and give empty string '' by default.
[Here](https://github.com/casbin/xorm-adapter/blob/master/adapter.go#L224), it does not forbid an empty string input.
However, [here](https://github.com/casbin/xorm-adapter/blob/master/adapter.go#L177) when loading, it ignores the value if it's an empty string.
So if we've ever inputted `'x', '', 'y', 'z'`, it will save in db with third option illustrated above, and be loaded as `'x', 'y', 'z'`, which is wrong.

So it might seem trivial, and ever question whoever will input an empty string field, it's important for the adapter library to have a standard behavior so that all adapters can be compatible (e.g. when switching from lang A to lang B(change of adapter) while staying with same DB, it should just work).

[...]

PR #1716, opened 2026-03-24
Fix: Treat "*" as explicit wildcard instead of "".

### LINKED ISSUE #1708: e.GetFilteredPolicy Empty string treated as wildcard
Expected behaviour:
In GetFilteredPolicy, "" empty string should be treated as empty string, and "*" should be treated as wildcard.

Actual behaviour:
"" empty string acts as a wildcard in this line of GetFilteredPolicy:
if fieldValue != "" && rule[fieldIndex+i] != fieldValue {

Suggested fix:
GetStrictFilteredPolicy which treats "" empty string as empty string, and treats "*" as wildcard.

GetStrictFilteredPolicy so as to not introduce a breaking change in GetFilteredPolicy

Why does this matter:
1. It's been discussed that Casbin adapters should support empty strings, therefore there should be a way to Get Filtered Policy for rules with empty string values.
2. Semantic accuracy.

DatScreamer:
This aligns with standard adapter behaviour, allowing "" to act as a literal value.

Fixes #1708 

Note: This is a breaking change for anyone using empty strings as wildcards, but is consistent with the standardized adapter behaviour discussed in #316 . 

As discussed in #316 ,  Adapters should/do support empty string values in rules in database. GetFilteredPolicy/RemoveFilteredPolicy couldn't properly handle those rules with empty string values in database due to empty strings being treated as wildcards.

Tests pass

### DIFF
--- a/management_api_test.go
+++ b/management_api_test.go
@@ -177,8 +177,8 @@ func TestGetPolicyAPI(t *testing.T) {
 		{"data2_admin", "data2", "write"}})
 
 	testGetFilteredPolicy(t, e, 0, [][]string{{"data2_admin", "data2", "read"}, {"data2_admin", "data2", "write"}}, "data2_admin", "data2")
-	// Note: "" (empty string) in fieldValues means matching all values.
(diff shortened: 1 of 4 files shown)


hsluoyz:
not planned as a breaking change

PR #1729, opened 2026-05-01
perf: optimize batch policy removal

huynhanx03:
## Overview

Optimize batch policy removal by rebuilding `Policy` and `PolicyMap` once in `RemovePoliciesWithAffected`. The previous implementation removed each rule one by one and updated policy indexes after every deletion, which becomes expensive for large batch removals.

## Change

- Remove policies in batch by collecting keys, compacting `Policy`, and rebuilding `PolicyMap` once.
- Add `BenchmarkRemovePoliciesLarge` to cover batch removal performance.

## Benchmark

```text
goos:  darwin
goarch: arm64
cpu:   Apple M1

BenchmarkRemovePoliciesLarge
metric      old       new      delta
ns/op       481,529   14,657   -97.0%
B/op        155,938   12,660   -91.9%
allocs/op   5,159     207      -96.0%
```

### DIFF
--- a/management_api_b_test.go
+++ b/management_api_b_test.go
@@ -172,3 +172,36 @@ func BenchmarkRemovePolicyLarge(b *testing.B) {
 		_, _ = e.RemovePolicy(fmt.Sprintf("user%d", rand.Intn(10000)), fmt.Sprintf("data%d", rand.Intn(10000)/10), "read")
 	}
 }
+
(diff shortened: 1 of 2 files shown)


huynhanx03:
Hi @hsluoyz , could you review?

huynhanx03:
Hi @hsluoyz , PTAL and approve if LGTM

PR #1739, opened 2026-07-03
Fix policy deletion with commas in quoted fields (#1733)

### LINKED ISSUE #1733: Quoted policy conditions with commas: load works, exact batch delete does not remove all rules
### Summary

We use Casbin with **6-field `p` policies** where the last field is a **condition expression** (quoted in CSV). Policies load correctly, but on **exact batch delete** (`RemovePolicies`) not all matching rules are removed — we see partial deletes when the condition text contains commas.

### Environment

- `github.com/casbin/casbin/v2` v2.77.x
- Large policy set (~350k `p` rules)
- Incremental deletes via `RemovePolicies` with full 6-field rules (including `condition`)

### What we see

- **Delete event:** 14 policies (user, tenant, resource, access, role, condition)
- **Requested:** batch delete of **14** exact-match rules
- **Result:** only **9** removed from persisted policy data
- The **5 not removed** all have **commas inside the condition string**
- Rules with conditions that have **no commas** (e.g. only `&&` between matchers) delete as expected

### When commas appear in conditions (valid ABAC expressions)

Commas are not only in `in (...)` lists. They can also appear in other matchers, for example:

- **List / enum checks:**  
  `r.attrs.ExceptionAllowed in ('( true','false )')`
- **Comma-separated allowed values in a matcher:**  
  `r.attrs.Center =~ '^center1,center2,center3$'`  
  or similar patterns where the regex/value side lists multiple centers/streams
- **Multiple constraints in one expression:**  
  `(r.attrs.Center =~ '^.*$' && r.attrs.Stream =~ '^.*$' && ...)`

So the condition field is a **free-form expression**, not a simple scalar — CSV must treat the **whole quoted field** as one token, including any commas inside it.

### Example — removed successfully (no commas in condition)

`csv
p, aqDsp4Rqe0PqHDJMZDI4P, aUSsW8GI03dyQ0AIFVn92, RESOURCE_STUDENT_MANAGEMENT, ACCESS_LEVEL_FULL, e0bmnxKv9L5AQGUwBtHFN, "(r.attrs.Center =~ '^.*$' && r.attrs.Stream =~ '^.*$')"`

### Example - Policies we have trouble removing (commas inside quoted condition)
```p, aqDsp4Rqe0PqHDJMZDI4P, aUSsW8GI03dyQ0AIFVn92, RESOURCE_SUPPORT_REFUND_PROCESS, ACCESS_LEVEL_FULL, e0bmnxKv9L5AQGUwBtHFN, "(r.attrs.Center =~ '^.*$' && r.attrs.Stream =~ '^.*$' && r.attrs.CourseType =~ '^.*$' && r.attrs.ExceptionAllowed in ('( true','false )') && r.attrs.DeductiblePercent < 100)"

p, aqDsp4Rqe0PqHDJMZDI4P, aUSsW8GI03dyQ0AIFVn92, RESOURCE_SUPPORT_ID_CARD, ACCESS_LEVEL_FULL, e0bmnxKv9L5AQGUwBtHFN, "(r.attrs.Center =~ '^.*$' && r.attrs.Stream =~ '^.*$' && r.attrs.CourseType =~ '^.*$' && r.attrs.ExceptionAllowed in ('( true','false )') && r.attrs.DiscountPercent < 100)"

p, aqDsp4Rqe0PqHDJMZDI4P, aUSsW8GI03dyQ0AIFVn92, RESOURCE_SUPPORT_STREAM_CHANGE, ACCESS_LEVEL_FULL, e0bmnxKv9L5AQGUwBtHFN, "(r.attrs.ExceptionAllowed in ('( true','false )') && r.attrs.DiscountPercent < 100 && r.attrs.Center =~ '^.*$' && r.attrs.Stream =~ '^.*$' && r.attrs.CourseType =~ '^.*$')"

p, aqDsp4Rqe0PqHDJMZDI4P, aUSsW8GI03dyQ0AIFVn92, RESOURCE_SUPPORT_BATCH_CHANGE, ACCESS_LEVEL_FULL, e0bmnxKv9L5AQGUwBtHFN, "(r.attrs.Center =~ '^.*$' && r.attrs.Stream =~ '^.*$' && r.attrs.CourseType =~ '^.*$' && r.attrs.ExceptionAllowed in ('( true','false )') && r.attrs.DiscountPercent < 100)"

p, aqDsp4Rqe0PqHDJMZDI4P, aUSsW8GI03dyQ0AIFVn92, RESOURCE_SUPPORT_PII_DATA, ACCESS_LEVEL_READ, e0bmnxKv9L5AQGUwBtHFN, "(r.attrs.Center =~ '^.*$' && r.attrs.Stream =~ '^.*$' && r.attrs.CourseType =~ '^.*$' && r.attrs.ExceptionAllowed in ('( true','false )'))"

Jbansal2:
This fixes an issue where policies with commas in quoted fields (e.g., ABAC conditions) were not being deleted correctly in batch operations.

Root Cause:
- Policy hash keys were generated using comma as separator via strings.Join(rule, ',')
- This created hash collisions when policy fields contained commas
- Example: ['alice', 'data,file', 'read'] and ['alice', 'data', 'file', 'read'] produced the same hash key 'alice,data,file,read'

Solution:
- Introduced policyKey() helper function that uses ASCII Unit Separator (\x1f)
- This character is designed for field separation and won't appear in policy data
- Updated all policy operations to use policyKey() instead of strings.Join()

Changes:
- model/policy.go: Added policyKey() and replaced all hash key generation
- Added comprehensive tests demonstrating the fix
- All existing tests pass (216 tests)

Fixes #1733

### DIFF
--- a/model/policy.go
+++ b/model/policy.go
@@ -35,6 +35,15 @@ const (
 
 const DefaultSep = ","
 
+// policyKey generates a unique key for a policy rule that handles fields containing the separator.
+// It uses ASCII unit separator (US, 0x1F) which is explicitly designed for this purpose
+// and is unlikely to appear in policy data.
+const policySeparator = "\x1f"
+
+func policyKey(rule []string) string {
+	return strings.Join(rule, policySeparator)
+}
+
 // BuildIncrementalRoleLinks provides incremental build the role inheritance relations.
 func (model Model) BuildIncrementalRoleLinks(rmMap map[string]rbac.RoleManager, op PolicyOp, sec string, ptype string, rules [][]string) error {
 	if sec == "g" && rmMap[ptype] != nil {
@@ -174,7 +183,7 @@ func (model Model) HasPolicy(sec string, ptype string, rule []string) (bool, err
 	if err != nil {
 		return false, err
 	}
-	_, ok := model[sec][ptype].PolicyMap[strings.Join(rule, DefaultSep)]
+	_, ok := model[sec][ptype].PolicyMap[policyKey(rule)]
 	return ok, nil
 }
 
@@ -200,7 +209,7 @@ func (model Model) AddPolicy(sec string, ptype string, rule []string) error {
 		return err
 	}
 	assertion.Policy = append(assertion.Policy, rule)
-	assertion.PolicyMap[strings.Join(rule, DefaultSep)] = len(model[sec][ptype].Policy) - 1
+	assertion.PolicyMap[policyKey(rule)] = len(model[sec][ptype].Policy) - 1
 
 	hasPriority := false
 	if _, ok := assertion.FieldIndexMap[constant.PriorityIndex]; ok {
@@ -215,10 +224,10 @@ func (model Model) AddPolicy(sec string, ptype string, rule []string) error {
 					break
 				}
 				assertion.Policy[i] = assertion.Policy[i-1]
-				assertion.PolicyMap[strings.Join(assertion.Policy[i-1], DefaultSep)]++
+				assertion.PolicyMap[policyKey(assertion.Policy[i-1])]++
 			}
 			assertion.Policy[i] = rule
-			assertion.PolicyMap[strings.Join(rule, DefaultSep)] = i
+			assertion.PolicyMap[policyKey(rule)] = i
 		}
 	}
 	return nil
@@ -238,7 +247,7 @@ func (model Model) AddPoliciesWithAffected(sec string, ptype string, rules [][]s
 	}
 	var affected [][]string
 	for _, rule := range rules {
-		hashKey := strings.Join(rule, DefaultSep)
+		hashKey := policyKey(rule)
 		_, ok := model[sec][ptype].PolicyMap[hashKey]
 		if ok {
 			continue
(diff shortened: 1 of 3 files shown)


hsluoyz:
Thanks for taking a look at this, but we're going to close this PR.

**It doesn't fix #1733.** A comma collision requires two rules with the same field count
whose comma-joins are identical, e.g. `["alice","b,c","d"]` vs `["alice,b","c","d"]`.
The rules in #1733 all differ in the resource field and cannot collide, and a collision
would cause wrong deletions rather than missed ones. We also can't reproduce #1733 on
master — batch-deleting rules with commas in the condition works correctly. The reporter
is on v2.77.x and that issue still needs a fresh repro.

**It introduces a regression.** `policyKey()` was only applied in `model/policy.go`.
`model/model.go:263` and `model/model.go:343` still build `PolicyMap` with
`strings.Join(policy, ",")`, and both run on every `LoadPolicy`. With
`examples/priority_model_explicit.conf`, `RemovePolicy({"1","alice","data1","write","allow"})`
returns `true` but deletes `[10 data1_deny_group data1 write deny]` instead — silent
authorization data corruption, correct on master.

**The tests don't cover the change.** All four new tests pass against unmodified master,
assert via `t.Logf` instead of failing, and neither file is gofmt-clean.

The underlying key-collision concern is real, though narrow. If you'd like to pursue it,
please open a separate issue and a new PR that routes every `PolicyMap` write through one
helper (including the two `model.go` sites), and includes a test that fails without the fix.

Issue #1004, opened 2022-05-03
[Bug] Performance issue in DomainMatcher with domain matching func


silverspace:
**Describe the bug**
When I modify the `BenchmarkRBACModelWithDomainPatternLarge` [performance test](https://github.com/casbin/casbin/blob/master/model_b_test.go#L269-L277) to add a bunch of unrelated users and then try to fetch an unauthorized resource, I see an exponential number of calls to the domain matching function `util.KeyMatch4`, resulting in exponentially bad performance related to the number of additional users. 

For context, I am trying to use a model.conf similar to this large scale performance test, but I am hitting massive performance issues when many different users with different domains are added. 

**To Reproduce**
I've modified the existing `BenchmarkRBACModelWithDomainPatternLarge` benchmark test to first add 1000 unrelated users with different domains to the unrelated role `staffOrgUser`. The result is that each of these users are being evaluated an exponential number of times when we try to run the enforcer.

```
func BenchmarkRBACModelWithDomainPatternLarge(b *testing.B) {
	e, _ := NewEnforcer("examples/performance/rbac_with_pattern_large_scale_model.conf", "examples/performance/rbac_with_pattern_large_scale_policy.csv")
	e.AddNamedDomainMatchingFunc("g", "keyMatch4", util.KeyMatch4)

	_ = e.BuildRoleLinks()
	for i := 0; i < 1000; i++ {
		orgID := rand.Int()
		user := fmt.Sprintf("staffUser%d", orgID)
		role := fmt.Sprintf("staffOrgUser")
		dom := fmt.Sprintf("/orgs/%d/sites/*", orgID)
		if err := e.GetRoleManager().AddLink(user, role, dom); err != nil {
			b.Fatal(err)
		}
	}

	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		const unauthorizedSite = "/orgs/999/sites/site001"
		_, _ = e.Enforce("staffUser1001", unauthorizedSite, "App001.Module001.Action1001")
	}
}
```

This results in an exponential number of calls to `util.KeyMatch4` (the domain matching function), exponentially related to the number of users added. (e.g. if I increase `1000` to `10000`, the benchmark never ends). 

```
goos: darwin
goarch: amd64
pkg: github.com/casbin/casbin/v2
cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
BenchmarkRBACModelWithDomainPatternLarge-12    	      19	  61330779 ns/op
PASS
```

[...]

Issue #330, opened 2019-11-23
error parsing regexp: missing argument to repetition operator: `*`


goapifirst:
Just started using this.. and man this library is amazing in what it provides.

I have followed the example on setting up a REST auth. I have auth.conf and auth.csv. They load, and from the example csv whenever I access a path that has the * in it in the .csv, I get the panic and the error message in the subject.

I am using the REST conf as linked to from here: https://casbin.org/docs/en/supported-models

My csv looks like:

```
p, admin, /*, *
p, anonymous, /login, *
p, p, /logout, *
p, p, /p/*, *
p, n, /logout, *
p, n, /n, *
```

I am using Chi, and have followed the example of the chi-authz from here: https://github.com/casbin/chi-authz

When I change the path to /p (removing the /* after it), for example, and try to access via Postman /p, /p/, /p/test, etc.. I either get Forbidden or see a panic output with the error message.

I have verified in my code that the role, path and method types passed to the Enforcer match what is found in the csv.

I am not clear what the first p is on each line though.

Is the regex call in the auth.conf file (e.g. this: [matchers]
m = r.sub == p.sub && keyMatch(r.obj, p.obj) && regexMatch(r.act, p.act)  ) wrong??

hsluoyz:
`*` is not valid in regex, use `.*` instead.

kevinduffey:
Interesting.. the example for Rest has that in their csv file. How does their code work (or does it)?

I did find another .conf example for Rest that doesn't have the regexp call in it, and it seems to work. Not sure if it is any better or not, but the matcher is:
m = r.sub == p.sub && keyMatch(r.obj, p.obj) && (r.act == p.act || p.act == "*")
vs 
m = r.sub == p.sub && keyMatch(r.obj, p.obj) && regexMatch(r.act, p.act) )

Is there anything better/faster with the regex one?

kdufeydev-pn:
Interesting.. the example for Rest has that in their csv file. How does their code work (or does it)?

I did find another .conf example for Rest that doesn't have the regexp call in it, and it seems to work. Not sure if it is any better or not, but the matcher is:
m = r.sub == p.sub && keyMatch(r.obj, p.obj) && (r.act == p.act || p.act == "*")
vs
m = r.sub == p.sub && keyMatch(r.obj, p.obj) && regexMatch(r.act, p.act) )

Is there anything better/faster with the regex one?

goapifirst:
[...]

PR #1740, opened 2026-07-17
perf: add fast paths to KeyMatch2/3/4/5 for literals and "*"

akmalsyrf:
## Summary

Adds cheap fast paths to `KeyMatch2`, `KeyMatch3`, `KeyMatch4`, and `KeyMatch5` before the existing `strings.Replace` + `RegexMatch` machinery, via a shared `keyMatchFastPath` helper. Two cases short-circuit:

1. `key2 == "*"` → `true` (bare wildcard; avoids depending on compiling `^*$` — see [#330](https://github.com/apache/casbin/issues/330))
2. `key2` is a **pure literal** → return `key1 == key2`

Anything else (real patterns, or `key2` carrying stray regexp metacharacters) falls through to the unchanged regex path.

**Why:** With `AddNamedDomainMatchingFunc("g", ..., KeyMatch2)` (or KeyMatch3/4/5), `DomainManager` evaluates the matcher O(N×D) times while building role links. Many real deployments use mostly **concrete** domains (`team/123`, `org/456`, …) plus a bare `"*"` admin/wildcard domain. After #1721, `RegexMatch` no longer recompiles on every call, but literal≠literal comparisons still pay for `strings.Replace`, pattern `ReplaceAllString`, map lookup, and `MatchString`. That remains a dominant cost for large domain graphs and shows up as cold-start time when policies are loaded on boot.

Related:

- Addresses a major practical hotspot from [#1004](https://github.com/apache/casbin/issues/1004) (does **not** fully solve #1004’s O(N²) role-manager architecture)
- Complements [#1721](https://github.com/apache/casbin/pull/1721) (cached `RegexMatch`)
- Related in spirit to [#1699](https://github.com/apache/casbin/pull/1699) (fast path around unconditional `strings.Replace`)
- Makes bare `"*"` explicit instead of relying on `^*$` ([#330](https://github.com/apache/casbin/issues/330); on current Go that pattern already match-alls)

## Correctness

`RegexMatch` embeds `key2` **unescaped** into `"^"+key2+"$"`, so the fast path must be careful:

1. **Literal detection.** Only treat `key2` as a literal when `regexp.QuoteMeta(key2) == key2` **and** it contains none of the matcher-specific extras `QuoteMeta` does not escape (`":"` for KeyMatch2). This keeps cases like `KeyMatch2("acmeXcom", "acme.com")` on the regex path.
2. **Equality is gated by literal detection.** A naive `key1 == key2 → true` is *not* equivalent: `"^a+b$"` does not match `"a+b"`. Non-literal `key2` falls through to the regex path.
3. **KeyMatch2 `:` asymmetry.** `":"` is `:param` dialect, not a Go regexp meta. `KeyMatch2("team:1", "team:2")` must stay `true` (`team[^/]+`); KeyMatch3 treats `":"` as a literal so the same pair is `false`. Covered by tests.

| Matcher | `extraPatternChars` | Notes |
|---------|---------------------|--------|
| `KeyMatch2` | `:` | `:param` dialect; `{`/`*` already covered by `QuoteMeta` |
| `KeyMatch3` | _(empty)_ | `{param}` / `*` covered by `QuoteMeta` |
| `KeyMatch4` | _(empty)_ | same; repeated-token logic unchanged |
| `KeyMatch5` | _(empty)_ | query string stripped from `key1` **before** fast path |

## Test plan

- [x] `go test ./util/` — KeyMatch2/3/4/5: bare `"*"`, pure literals, `.` metas, meta-equality (`a+b`/`a?b`), KeyMatch2 vs KeyMatch3 colon asymmetry, KeyMatch5 query strip + fast paths
- [x] `go test ./...` — full suite green
- [x] `gofmt` clean; benchmarks `BenchmarkKeyMatch2_LiteralDomains`, `BenchmarkKeyMatch2_RESTPatterns`

Illustrative local numbers (not a formal perf claim; use CI / longer `-benchtime` if citing):

```
BenchmarkKeyMatch2_LiteralDomains    ~30ms/op    0 B/op    0 allocs/op
```

Pure concrete domains + `"*"` short-circuit with **0 allocs** (`regexp.QuoteMeta` returns the input unchanged when there is nothing to escape).

## Out of scope

- Redesigning `DomainManager` to index pattern vs plain domains (#1004 root cause)
- Changing the accidental "regexp-in-`key2`" behavior on the slow path (left unchanged)
- `KeyMatch`/`GlobMatch`, adapter, or enforcer API changes

### DIFF
--- a/util/builtin_operators.go
+++ b/util/builtin_operators.go
@@ -57,32 +57,20 @@ func mustCompileOrGet(key string) *regexp.Regexp {
 }
 
 // keyMatchFastPath handles cheap cases shared by KeyMatch2/3/4/5 before the
-// string-transform + RegexMatch path.
+// string-transform + RegexMatch path. handled=true means matched is final.
 //
-// When handled is true, matched is the final result and the caller must return it.
-//
-// Cases:
-//  1. key1 == key2 → true
-//  2. key2 == "*" → true (bare wildcard). Explicitly handled so domain matchers
-//     using AddNamedDomainMatchingFunc do not depend on compiling "^*$", which
-//     historically panicked ("missing argument to repetition operator") on some
-//     Go/regexp versions — see #330. On current Go, "^*$" match-alls; either way
-//     the intended semantics for a bare "*" is match-any.
-//  3. key2 is a pure literal → false (equality already checked). RegexMatch
-//     embeds key2 unescaped into "^"+key2+"$", so it is not enough to look only
-//     for matcher-specific pattern characters (* / : / {): values like "acme.com"
-//     contain no KeyMatch dialect markers but `.` is still a regexp metacharacter.
-//     We therefore require regexp.QuoteMeta(key2) == key2, plus none of
-//     extraPatternChars (e.g. ":" for KeyMatch2, which QuoteMeta does not escape).
+//  1. key2 == "*" → true (bare wildcard / domain admin). Avoids depending on
+//     compiling "^*$", which historically panicked on some Go versions (#330).
+//  2. key2 is a pure literal → key1 == key2. Literal means QuoteMeta(key2) == key2
+//     and no matcher-specific chars in extraPatternChars (e.g. ":" for KeyMatch2's
+//     :param dialect). Equality is inside this gate so patterns or stray regexp
+//     metas (".", "+", …) still take the regex path and keep prior behavior.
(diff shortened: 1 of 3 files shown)


hsluoyz:
replaced by: https://github.com/apache/casbin/commit/9e2b76713fc4d19d8d49c41dcba336696f88ce3c

